PR #7, opened 2010-12-09
ask and you shall receive!

NotMyself:
I threw in a quick build script using psake for you.


damianh:
Powershell is xplat now :)

PR #671, opened 2012-07-06
render child requests in razor views

tylerburd:
This is a proof of concept on rendering child requests within views.
This allows for more modular view pieces than you get with Partial
views, as the parent doesn't have to worry about constructing the model
for the child view piece.  To see the proof in action, launch the
Nancy.Demo.Hosting.Aspnet project and browse to the /viewwithchildview
route.


uluhonolulu:
@grumpydev  Was it ever finished? I read it was going to be in 0.18..

phillip-haydon:
@uluhonolulu we don't support rendering routes inside views. It's currently not finished and not on the table. There's very little reason to support it now that most web applications are moving serverside rendering to the client. Besides we already support partials.

Issue #751, opened 2012-09-16
Race Conditions in Razor View Engine


thefringeninja:
I'm getting two exceptions at random intervals when hosting on AppHarbor. When the view engine b0rks like this, you can't recover unless you unload the appdomain. Like most multi-threaded bugs this one is really hard to reproduce in a test.

tl;dr One possible workaround is precompiled views via msbuild task. If anyone else is working on this I would love to contribute to their effort.

System.InvalidOperationException: The output writer for code generation and the writer supplied don't match and cannot be used. This is generally caused by a bad implementation of a CodeGenerator derived class.

Stack trace

at Microsoft.CSharp.CSharpCodeGenerator.System.CodeDom.Compiler.ICodeGenerator.GenerateCodeFromCompileUnit(CodeCompileUnit e, TextWriter w, CodeGeneratorOptions o)
at Microsoft.CSharp.CSharpCodeGenerator.FromDomBatch(CompilerParameters options, CodeCompileUnit[] ea)
at Microsoft.CSharp.CSharpCodeGenerator.System.CodeDom.Compiler.ICodeCompiler.CompileAssemblyFromDomBatch(CompilerParameters options, CodeCompileUnit[] ea)
at System.CodeDom.Compiler.CodeDomProvider.CompileAssemblyFromDom(CompilerParameters options, CodeCompileUnit[] compilationUnits)
at Nancy.ViewEngines.Razor.RazorViewEngine.GenerateRazorViewFactory(CodeDomProvider codeProvider, GeneratorResults razorResult, Assembly referencingAssembly, IEnumerable`1 rendererSpecificAssemblies, Type passedModelType)
at Nancy.ViewEngines.Razor.RazorViewEngine.GetCompiledViewFactory(String extension, TextReader reader, Assembly referencingAssembly, Type passedModelType)
at Nancy.ViewEngines.Razor.RazorViewEngine.<>c__DisplayClass40.<GetOrCompileView>b__3f(ViewLocationResult x)
at Nancy.ViewEngines.DefaultViewCache.<>c__DisplayClass1`1.<GetOrAdd>b__0(ViewLocationResult x)
at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
at Nancy.ViewEngines.DefaultViewCache.GetOrAdd[TCompiledView](ViewLocationResult viewLocationResult, Func`2 valueFactory)
[...]

PR #773, opened 2012-10-03
Race Conditions in Razor View Engine

thefringeninja:
Here's the failing test and possible solution for issue #751 (couldn't figure out how to attach code to it). 

System.IndexOutOfRangeException: Probable I/O race condition detected while copying memory. The I/O package is not thread safe by default. In multithreaded applications, a stream must be accessed in a thread-safe way, such as a thread-safe wrapper returned by TextReader's or TextWriter's Synchronized methods. This also applies to classes like StreamWriter and StreamReader.

### DIFF
--- a/src/Nancy.ViewEngines.Razor/CSharp/CSharpRazorViewRenderer.cs
+++ b/src/Nancy.ViewEngines.Razor/CSharp/CSharpRazorViewRenderer.cs
@@ -32,9 +32,12 @@
         public RazorEngineHost Host { get; private set; }
 
         /// <summary>
-        /// Gets the provider that is used to generate code.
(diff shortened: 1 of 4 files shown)


mylemans:
I just checked the decompiled code of Nancy.ViewEngines.Razor.2.0.0-clinteastwood and I do not see the CreateProvider method but just the Provider property. 

I'm having the exact issue found at https://github.com/NancyFx/Nancy/issues/751

It happens when I try to refresh or open the website in another tab while the website ls launching (before the first request finishes)

Why isn't this fix included there?

KnowMoreIT:
Looks like the same issue I opened an issue about: #2731

PR #988, opened 2013-02-15
New route resolution engine

### LINKED ISSUE #785: Route matching too greedy: parsed segments may contain segment seperators
Suppose we have the route `/containers/{containerId}/contents`.

If we request `/containers/4/something/contents`, this route will be matched and `containerId` will get a value of `4/something`.

This seems counter-intuitive. I would expect `{containerId}` to only match a single path segment and think the currently implemented greedy alternative is a far less occurring situation.

[The documentation](https://github.com/NancyFx/Nancy/wiki/Defining-routes) seems to suggest this is by design. Is this really the case? If it is, I am interested in the reasoning behind this (since it conflicts with all the other routing mechanism I know of).

My proposed fix is not to match any character in a segment, but any non-slash character. Change at [`DefaultPatternMatcher#ParameterizeSegment` line 141](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Routing/DefaultRoutePatternMatcher.cs#L141):
`@"(?<{0}>.+?)"`
becomes
`@"(?<{0}>[^\/]+?)"`

If one then actually wants to match the segment seperator, (s)he can always use the original RegEx: `?<foo>.+?`.

grumpydev:
Switched from the old approach to a route Trie, which is faster, and much more readable.

Also removed the Indiana Jones and the Tuple of Doom that the old route resolver was returning and replaced it with a class.

Routing should (hopefully) be unaffected, but to be clear the new system works on route length and scoring as follows:
- Longest routes take precedence - if a route has 5 matched segments it will win over one with 4, regardless of what kind of nodes they were. In the case of node that have 'default' values, if the default value is used then the node isn't counted towards the length.
- Node scoring - if more than one result has the same length then each node in each route is totted up to a score, and the highest score wins. The scores for each node type are currently:
  - Literal node - 10000
  - Capture node  - 1000
  - Regex node - 1000
  - Greedy capture node - 0

The full set of node types are:
- Literal node - just text
- Capture node - {foo}
- Optional capture node - {foo?}
- Optional capture node with default - {foo?defaultValue}
- RegEx node - (?<foo>\d{2,4})
- Greedy capture node - {foo*}

Note that now captures are _not_ greedy by default, you need to specify the greedy capture if that's the behaviour you want (this fixes #785)

### DIFF
--- a/src/Nancy/Routing/Trie/Nodes/TrieNode.cs
+++ b/src/Nancy/Routing/Trie/Nodes/TrieNode.cs
@@ -196,7 +196,7 @@ namespace Nancy.Routing.Trie.Nodes
         /// this node is the end result
         /// </summary>
         /// <param name="capturedParameters">Currently captured parameters</param>
-        /// <param name="dictionary"></param>
+        /// <param name="localCaptures">Parameters captured by the local matching</param>
         /// <returns>Array of <see cref="MatchResult"/> objects corresponding to each set of <see cref="NodeData"/> stored at this node</returns>
         protected IEnumerable<MatchResult> BuildResults(IDictionary<string, object> capturedParameters, IDictionary<string, object> localCaptures)
         {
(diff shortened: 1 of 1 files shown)


ti24horas:
Hi,
Tried lastest nuget version (1.2.0) with self host and an error occurs.
After trying earlier versions, I found this is the offending pull.

The error occurs as following:

1. Created a Console App(x64) self host application using Nancy.Hosting.Self (ver 0.17.0) with Nancy ver 0.17.0
2. Created a custom module inheriting NancyModule
3. In constructor created the following routes:

<pre lang="c#">
public HomeModule()
        {
            Get["test.html"] = _ => "OK";
            
            Get["/"] = _ => "Hello world";
        }
</pre>

4. When calling http://localhost:8080/test.html, following exception appears:

<pre>
Nancy.RequestExecutionException: Oh noes! ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Nancy.Routing.Trie.Nodes.TrieNode.d__7.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Nancy.Routing.Trie.RouteResolverTrie.GetMatches(String method, String path, NancyContext context)
   at Nancy.Routing.DefaultRouteResolver.Resolve(NancyContext context)
   at Nancy.Routing.DefaultRequestDispatcher.InvokeRouteResolver(NancyContext context, String path, IEnumerable`1 acceptHeaders)
   at Nancy.Routing.DefaultRequestDispatcher.Resolve(NancyContext context)
   at Nancy.Routing.DefaultRequestDispatcher.Dispatch(NancyContext context)
   at Nancy.NancyEngine.InvokeRequestLifeCycle(NancyContext context, IPipelines pipelines)
   --- End of inner exception stack trace ---
   at Nancy.NancyEngine.InvokeOnErrorHook(NancyContext context, ErrorPipeline pipeline, Exception ex)
</pre>

Tested using chrome

versions prior this (0.16.1) works as usual, same code.

ti24horas:
Hi guys,
Just downloaded from sources, referenced Nancy and SelfHost from csproj and it worked.
Even in debug mode. Don't know why this happened.

PR #1071, opened 2013-04-18
Body populated with error when XML/JSON serialization takes place

jchannon:
I had a cyclical POCO reference that was marked with XmlIgnore attributes and when a request came in and I used application/xml this returned my data however when using application/json I got a blank response.

This puts a try/catch around the call to Serialize and writes the exception message to the stream.

### DIFF
--- a/src/Nancy.Testing.Tests/BrowserFixture.cs
+++ b/src/Nancy.Testing.Tests/BrowserFixture.cs
@@ -353,6 +353,19 @@ namespace Nancy.Testing.Tests
             result.Body.AsString().ShouldNotBeEmpty();
         }
 
+        [Theory]
+        [InlineData("application/json")]
+        [InlineData("application/xml")]
+        public void Should_return_no_error_message_on_cyclical_exception_when_disabled_error_trace(string accept)
+        {
+            //Given/When
+            StaticConfiguration.DisableErrorTraces = true;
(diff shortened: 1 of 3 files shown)


thecodejunkie:
@jchannon should we revive this or close it?

jchannon:
Revive? I must have worked around it but still seems a valid issue?

thecodejunkie:
Alright

> Nope, was after some help

What kind of help were you looking for? Should we target this for the 1.3 or 1.4 timeframe?

phillip-haydon:
@jchannon why do you break everything!

jchannon:
We can try 1.3. I think I was after some help as in someone else clone and
try on their machine to see if they see the same

@philliphaydon just pushing the limits :p

thecodejunkie:
@jchannon what can I do to help bring this ship into port?

jchannon:
Clone it and see if you see the debug vs release issue I guess

ClemPi:
Hi.

I'm not sure silently catching the exception is a good practice. It will lead to a request returning 200 OK with empty content when the serialization fail.

jchannon:
This is a PR that wasn't merged

ClemPi:
Yes, but #2086 was merged.

jchannon:
#2086 looks the one you want. Take a look at that and then append the tests to check the response status code and see what they say. If not correct feel free to send a PR :)

Issue #1082, opened 2013-05-02
Adding Nancy.Viewengines.Razor NuGet package breaks xbuild on mono on Ubuntu


colinangusmackay:
If a project has the "Nancy.Viewengines.Razor" NuGet package added then the build of the project  fails when using xbuild on mono on Ubuntu.

This is because the package adds the following to the csproj file:
```xml
<PostBuildEvent>
if $(ConfigurationName) == Debug (
xcopy /s /y /R "$(SolutionDir)packages\Nancy.Viewengines.Razor.0.17.1\BuildProviders\Nancy.ViewEngines.Razor.BuildProviders.dll" "$(ProjectDir)bin\"
xcopy /s /y /R "$(SolutionDir)packages\Nancy.Viewengines.Razor.0.17.1\lib\Net40\Nancy.ViewEngines.Razor.dll" "$(ProjectDir)bin\"
)</PostBuildEvent>
```

My current workaround is to remove the content of the `<PostBuildEvent>` and reference any type in the `Nancy.ViewEngines.Razor.dll` explicitly somewhere in the project. e.g. Overriding the default bootstrapper and then adding this:

    protected override System.Collections.Generic.IEnumerable<System.Type> ViewEngines
    {
      get 
      { 
        yield return typeof (RazorViewEngine);
      }
    }

So far, this seems to work without issue - although the `Nancy.ViewEngine.Razor.BuildProviders.dll` is not copied in this instance.

zippy1981:
Let me ask the bigger question. Lots of OSS C# has xcopy commands in the msbuild. It would be relatively trivial to write an xcopy program in C or python, or shell that would sit in the path and make this magicallywork. The hard part is getting mono to ship this binary alongsize xbuild

stkrzysiak:
@zippy1981 I just learned about xcopy, but for those familiar with both xcopy and rsync, couldn't we just write a wrapper around rsync that converts xcopy->rsync flags? At the very least we could make an alias for xcopy?

LinuxDoku:
Why don't simply use the copy task of msbuild which is supported by xbuild on mono out of the box? https://msdn.microsoft.com/library/3e54c37h.aspx (also featured in the msbuild task porting guide: http://www.mono-project.com/archived/porting_msbuild_projects_to_xbuild/)

There was a little bug in monodevelop but with the latest release it does not break the .csproj anymore and can be widely adopted with full tooling support.

phillip-haydon:
*shrug* 

[...]

PR #1170, opened 2013-08-01
Fixing Nancy.Viewengines.Razor NuGet package support for Mono on Linux

salaerts:
I ran into issue #1082 and I couldn't get the PostBuildEvent and CustomCommands to play nicely and consistently across all platforms.
Replacing the PostBuildEvent with an AfterBuild target seems to work everywhere (tested on VS 2012, Xamarin on Windows 8 and xbuild on Debian), but of course there are minor issues.

The solution consists of a targets file, which upon installation will automatically be added to the project file by NuGet.
To have the BuildProvider immediately available, there is a small addition to the installation script. Install.ps1 gets the active configuration, checks if it is a debug config and copies the files to the output dir.

Unfortunately this doesn't work when building with package restore. When the build starts the targets file is not available for import and so even after the package is installed the AfterBuild target is not executed.
NuGet also doesn't clean up the project file when uninstalling, which is covered by the Exists condition but still isn't very neat.

Maybe some additional tweaks are possible to get the PostBuild event to work only within visual studio, perhaps using the VisualStudioVersion property as a condition.

Any thoughts?


thecodejunkie:
Nuget 3.x will deprecate install.ps1/uninstall.ps1 so we will revisit this to support the new infrastructure http://blog.nuget.org/20150729/Introducing-nuget-uwp.html

PR #1263, opened 2013-10-14
Test to show binding to a List<T> property on a model fails

jchannon:
The binder should be able to bind to List <T> property on a model but currently doesn't.


khellang:
@jchannon Are you expecting this to be pulled? I think we need an actual fix as well :wink:

khellang:
Ping @jchannon! Are you planning on submitting a fix for this failing test as well? :smile:

jchannon:
I think this is one for the model binder king @grumpydev

jchannon:
Deleted and issue raised #2169

Issue #1232, opened 2013-09-20
OnError pipeline hooks receive System.AggregateException


rasmus-beck:
This is a bug introduced in v. 0.20.0

If you hook up an error handler to the OnError pipeline to catch and handle exceptions thrown in Nancy, the exceptions passed on to the error handler are wrapped in two System.AggregateExceptions.

The expected behaviour is that these should have been unwrapped before handing off to the error handler hooked up on the pipeline. 

I have created a small project that provokes this behaviour.

https://github.com/rasmus-beck/NancyAggregateExceptionBug

If time permits this weekend, then I will try to create a failing test in the Nancy project. But I wanted to report the error right away.

PR #1269, opened 2013-10-16
Added method to flatten exceptions before passing them on to the OnError pipeline hook

rasmus-beck:
This is a proposed fix for issue #1232.

As described in issue #1232 the error pipeline hook does not unwrap the AggregateExceptions thrown from Tasks in the request handler.

One comment has to be made on this pull request. More than one exception can be nested inside AggregateExceptions, but the OnError pipeline hook only expects a single exception, so unless only one exception is actually nested inside the AggregateException we cannot fully unwrap it. In the case where more than one exception exists we have to return all of them in order not to swallow up any exception.

This is mentioned in a comment in the code as well, that should perhaps be deleted if the pull request is accepted.

Jon Skeet describes this issue in a short comment on Stack Overflow:
http://stackoverflow.com/questions/13570855/aggregateexception-what-does-aggregateexception-flatten-innerexception-repre

### DIFF
--- a/src/Nancy/NancyEngine.cs
+++ b/src/Nancy/NancyEngine.cs
@@ -287,9 +287,9 @@
                 {
                     try
                     {
-						var flattenedException = FlattenException(t.Exception);
+                        var flattenedException = FlattenException(t.Exception);
 
-						InvokeOnErrorHook(context, pipelines.OnError, flattenedException);
+                        InvokeOnErrorHook(context, pipelines.OnError, flattenedException);
 
                         tcs.SetResult(context);
                     }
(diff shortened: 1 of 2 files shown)


src/Nancy/NancyEngine.cs
```diff
@@ -338,5 +340,31 @@ private static void InvokeOnErrorHook(NancyContext context, ErrorPipeline pipeli
...
+                {
+                    return flattenedAggregateException;
+                }
+
+                return flattenedAggregateException.InnerException;
+            }
+
+            if (exception != null && exception.InnerException != null)
+            {
+                return FlattenException(exception.InnerException);
```
khellang: @rasmus-beck Can you remember why you're doing this? This will essentially throw away valuable information, no?

PR #1300, opened 2013-10-31
When using Content Negotiation, and the CoerceStupidBrowsers convention,...

btompkins:
... content is negotiated as HTML even when the browser requests json.  This fixes by adding an additional check for json.


thecodejunkie:
Will revisit this in another form.. a lot has changed since. Thanks!

Issue #1220, opened 2013-09-11
Should be able to specify a full url when using Nancy.Testing


grumpydev:
Should be able to specify protocol, host name, base path and port, in addition to the path that's already supported.

PR #1312, opened 2013-11-06
Made Url on Browser defineable

jchannon:
As defined in #1220


cloudhunter89:
@jchannon Should this pr allow using relative paths and query strings in the testing browser? (something like my sample in #1326)

jchannon:
I can't remember, it would appear so but who knows 😄

PR #1316, opened 2013-11-08
Razor view compilation errors now return a 500 response code

davidalpert:
I was recently surprised to learn that requesting a Razor view in Nancy would return a status 200 OK even when the view failed to compile.

An xUnit test to document my expectation is included in the pull request and can be reviewed below.

The core change I made to the Nancy RazorViewEngine was to move rendering of the body's HTML out of the Response.Content Action<Stream> delegate and complete it before returning the Response object so that any rendering issues could be discovered and the Response's StatusCode property set properly before returning the Response to the infrastructure.

All existing RazorViewEngine tests continue to pass for me after this change.

### DIFF
--- a/src/Nancy.ViewEngines.Razor/NancyRazorBuildViewBodyResult.cs
+++ b/src/Nancy.ViewEngines.Razor/NancyRazorBuildViewBodyResult.cs
@@ -1,15 +1,37 @@
-﻿using System.Linq;
-using System.Collections.Generic;
-
-namespace Nancy.ViewEngines.Razor
+﻿namespace Nancy.ViewEngines.Razor
 {
+    using System.Linq;
+
     /// <summary>
-    /// Encapsulates the result of generating the HTML body of a view, allowing the <see cref="RazorViewEngine"/>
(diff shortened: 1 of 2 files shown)


davidalpert:
just saw this; is action required on my part or has the issue been handled by another pull request?

Jevonius:
Hi @davidalpert , I've just been bitten by this issue again (failing View render, but the tests still pass as status code is still OK). I've just updated our code to 1.4.1 and the problem remains, so I don't think this issue has been fixed elsewhere.

It's been quite a while since you raised this issue originally, but if you're still active it would be great to get this implemented so that I don't have to have something like the following in my tests:
``` C#
Assert.That( result.Body.AsString().Contains( "Razor Compilation Error" ), Is.False, "Razor Compilation Error occurred." );
```
I believe the tests in my commit referenced above (c4d6b9d5d35be9cd5a68590d737d28d8ccc62529) are still valid and could be dropped into place to demonstrate the issue/fix.

If you're not still active, I'll give it a go myself since this has bitten me again!

Issue #1096, opened 2013-05-22
ModelBinding bad data to a System.DateTime does not throw a ModelBindingException


FlukeFan:
In test:
    Should_throw_ModelBindingException_if_convertion_of_a_property_fails()

... adding the following:

context.Request.Form["DateProperty"] = "baddate";

... does not cause the test to fail.

PR #1378, opened 2013-12-20
Fix for issue #1096 (DateTime converter on model binding)

phillip-haydon:
This PR replaces the botched #1099 PR.

Fix is for  #1096

The unit test in question is fine, the unit test didn't include the type converter, however the Fallback converter was being called before the Date so the Date Converter didn't have a chance to say he could convert it.

I added an Order property and order the collection before looking for the first matching one.

I figured this solution was fine since there's only 4-5 converters so the overhead of sorting would be almost nothing, if not nothing already.

Fixed up a few places where type converters were being used.


thecodejunkie:
@phillip-haydon what's your say on this? Close or rebase and revisit?

khellang:
I'm closing this as it seems @phillip-haydon doesn't have time to fix his mess :wink:

PR #1470, opened 2014-03-02
Fixed module registrations being ignored when that module type is resolved

despertargz:
Whenever Nancy would resolve module types it wouldn't check if the type was already registered in the IOC container or not and would just re-register it. This caused any module type registrations done by the user to be ignored.

### DIFF
--- a/src/Nancy.Testing.Tests/ModuleRegistrationTests.cs
+++ b/src/Nancy.Testing.Tests/ModuleRegistrationTests.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace Nancy.Testing.Tests
+{
+    public class ModuleRegistrationTests 
+    {
+        [Fact]
+        public void test_module_resolves_from_previous_container_registration()
+        {
+            Browser browser = new Browser(new ModuleRegistrationBootStrap());
+            string responseText = browser.Get("/test").Body.AsString();
+            Assert.Equal("return value", responseText);
+        }
+    }
+
+    public class ModuleRegistrationBootStrap : DefaultNancyBootstrapper
+    {
+        protected override void ConfigureApplicationContainer(TinyIoc.TinyIoCContainer container)
+        {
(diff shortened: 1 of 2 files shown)


thecodejunkie:
Ping @despertargz could you provide some more background on this? Either going to pull it into the pending 1.3 release or close it. Thanks =)

despertargz:
Sure. The bootstrapper provides an IoC container which allows the user to register types. This gives the user complete flexibility over how and what instance gets returned when Nancy needs to resolve the module type. 

However, at https://github.com/NancyFx/Nancy/pull/1470/files#diff-54eef3b8fa44e8d7ab316b8da9943c07L127 , Nancy __overwrites__ any registrations the user may have done with its own generic registration. 

I think this takes control away from the user being able to specify the instance they want returned when a type is resolved. The IOC container also gives the user an illusion that they have this control.

thecodejunkie:
Ok, now we are getting somewhere. If I understand you correctly, you are manually registering `INancyModule` **instances** in the container, which is then overridden by the default registration of modules. Correct? If so then my question to you is:

Why are you manually registering modules in the container, rather than relying on Nancy doing it for you?

So I am much more interested in the _why_ than the _how_ because we've stepped outside of the realm of what Nancy has been designed for. I need to understand the use case / scenario that you are attempting to resolve, so that I can form an opinion on whether or not this is something we should be supporting in the core.

Thanks

thecodejunkie:
Also, as a side note, I want to point out that such a change is not as trivial as it might seem. We would have to update all `Nancy.Bootstrapper.*` (autofac, unity, structuremap and so on) to provide the same functionality so we ensure that we provide a consistent behavior independent of which bootstrapper that is being used. If that was the case then we would probably have to re-design the bootstrapper API a bit and perhaps provide an `abstract` method (or similar) to help surface the requirement of having the module registration support not overriding user-registered modules.

thecodejunkie:
@despertargz any more feedback on this? Thanks

khellang:
Closing this as it requires a bit more though (and response to feedback). Feel free to reopen (or maybe file an issue, so we can discuss it further) :smile:

PR #1473, opened 2014-03-03
Update data annotations validation to support the DisplayName attribute

### LINKED ISSUE #1469: DataAnnotations validators don't use the display name in error messages
I'm working on a `Nancy.Validation.DataAnnotations.Extensions` project to add support for the .net 4.5 validation attribtutes and I found that even when I add `[DisplayName("Some Field Name")]` to my property the error message is still using the property name instead of `Some Field Name`.

I had to override the `Validate` function as follows and this gets me the display name if it's available.

```c#
public override IEnumerable<ModelValidationError> Validate(object instance, ValidationAttribute attribute, PropertyDescriptor descriptor)
{
    var context = new ValidationContext(instance, null, null)
    {
        DisplayName = descriptor == null ? null : descriptor.DisplayName,
        MemberName = descriptor == null ? null : descriptor.Name
    };

    if (descriptor != null)
    {
        instance = descriptor.GetValue(instance);
    }

    var result = attribute.GetValidationResult(instance, context);

    if (result != null)
    {
        yield return new ModelValidationError(result.MemberNames, attribute.FormatErrorMessage(context.DisplayName));
    }
}
```

I'm not sure if this is a general enough fix to add to the [DataAnnotationsValidatorAdapter](https://github.com/NancyFx/Nancy/blob/319933dc4cbb640486e05ff77f4ceb87a86f80cc/src/Nancy.Validation.DataAnnotations/DataAnnotationsValidatorAdapter.cs). The changes were adding
```c#
DisplayName = descriptor == null ? null : descriptor.DisplayName,
```
and changing
```c#
string.Join(" ", result.MemberNames.Select(attribute.FormatErrorMessage))
```
 to
```c#
attribute.FormatErrorMessage(context.DisplayName)
```

xt0rted:
This fixes https://github.com/NancyFx/Nancy/issues/1469

It also contains a change to allow modifying the `ModelValidationError` returned from the `Validate` method without the need to override it.

This is useful for class level validators which won't return member names, or even built-in attributes such as the `CompareAttribute` which also doesn't return member names.

### DIFF
--- a/src/Nancy.Validation.DataAnnotatioins.Tests/DataAnnotationValidatorFixture.cs
+++ b/src/Nancy.Validation.DataAnnotatioins.Tests/DataAnnotationValidatorFixture.cs
@@ -1,6 +1,7 @@
 ﻿namespace Nancy.Validation.DataAnnotations.Tests
 {
     using System;
+    using System.ComponentModel;
     using System.ComponentModel.DataAnnotations;
     using System.Linq;
 
@@ -161,9 +162,25 @@
             subject.Description.Rules.SelectMany(r => r.Value).ShouldHave(r => r.RuleType == "Oops" && r.MemberNames.Contains(string.Empty));
         }
 
+        [Fact]
+        public void Should_use_display_name_attribute()
+        {
+            // Given
+            var subject = this.factory.Create(typeof(TestModel));
+            var instance = new TestModel { FirstName = "a long name", Age = "1" };
+
+            // When
+            var result = subject.Validate(instance, new NancyContext());
+
+            // Then
+            result.IsValid.ShouldBeFalse();
+            result.Errors["FirstName"][0].ErrorMessage.ShouldContain("First Name");
+        }
+
         [OopsValidation]
         private class TestModel : IValidatableObject
         {
+            [DisplayName("First Name")]
             [Required]
             [StringLength(5)]
             public string FirstName { get; set; }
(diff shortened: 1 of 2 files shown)


thoemmi:
Will this PR be merged eventually? Or this issue fixed? I really would like to use `DisplayAttribute` or `DisplayNameAttribute` :confused:

khellang:
@xt0rted This needs a rebase. What are you thinking here? Should we just leave things as is? We're moving to newer versions of Mono for the 2.0 stuff. We'd like to just get this in ASAP to continue that work.

xt0rted:
@khellang as of mono 3.10 their bug is fixed. If it's ok that versions prior to 3.10 will give different results then I can remove that last commit and make the unit tests conditional as @thecodejunkie had suggested.

khellang:
@xt0rted Did you just rebase this? :stuck_out_tongue:

xt0rted:
@khellang I sure did

khellang:
I :heart: your sneakyness :wink:

khellang:
For some reason GitHub still think this is "out-of-date with the base branch" :confused:

khellang:
Thanks! :+1:

xt0rted:
@khellang when I rebased the first time the note at the bottom said the branch could be automatically merged. Once the most recent PR was merged this note changed saying it couldn't be merged. Did someone play around with the new [protected branches](https://github.com/blog/2051-protected-branches-and-required-status-checks) feature by any chance? TSLint had an issue with this and had to turn it off https://github.com/palantir/tslint/issues/747.

khellang:
For some reason it says it's out of date again. Seems like a GitHub bug :confused:

khellang:
OK, I killed the required status checks for now. Don't know what's going on...

// @NancyFx/most-valued-minions

khellang:
Awesome :grin:

xt0rted:
:tada:

PR #1489, opened 2014-03-16
Retain casing for dictionary types in tests

### LINKED ISSUE #1484: DeserializeJson doesn't retain casing for dictionary types
I'm not sure if this is intentional or not but the [DeserializeJson](https://github.com/NancyFx/Nancy/blob/09f3556de53c9d976344308d2d9a56aacb57faab/src/Nancy.Testing/BrowserResponseBodyWrapperExtensions.cs#L56) function doesn't retain casing for `Dictionary<string, object>`.

This modified test shows a repro of the issue.

``` c#
[Fact]
public void Should_be_able_to_send_json_in_body()
{
    // Given
    var model = new { someString = "Some String", someInt = 29, someBoolean = true };

    // When
    var result = browser.Post("/", with =>
                                    {
                                        with.JsonBody(model);
                                    });

    // Then
    var actualModel = result.Body.DeserializeJson<Dictionary<string, object>>();

    actualModel.ShouldNotBeNull();
    actualModel["someString"].ShouldEqual(model.someString);
    actualModel["someInt"].ShouldEqual(model.someInt);
    actualModel["someBoolean"].ShouldEqual(model.someBoolean);
}
```

Using upper case keys isn't too major of an issue since it's just for tests, but it would be nice for consistency.

xt0rted:
Since this is a breaking change I wasn't sure if I should include a setting similar to `JsonSettings.RetainCasing` that was specific to `Nancy.Testing` and the `DeserializeJson` method.

This fixes https://github.com/NancyFx/Nancy/issues/1484


MrTomWhite:
@xorted and @grumpydev: do we have any intent to merge this pull request?

xt0rted:
Just rebased, everything looks good so far.

xt0rted:
I'm not sure why that token auth test failed but since it's been removed in v2 I'm not going to look into this any further unless we need it for a pre-v2 release.

khellang:
@xt0rted We're not going to break builds of master because we have a v2 branch going where we're removing token auth...

khellang:
@xt0rted Looks like it was SNAFU with the CI build. Restarted it and all looks OK :smile:

horsdal:
I'm sorry, but I still don't see how this is better than the current behavior.
I can see that this is one possible way to have deserialization for tests work, but I don't see how it is necessarily better than the current behavior.
In fact since the `RetainCasing` value becomes hard coded, I think we just leave people with less control and a new default that is just as debatable as the current default.

khellang:
I think I agree with @horsdal on this one. The tests should be in line with the actual settings and not hard coded to retain casing.

In JSON, a .NET dictionary is the same as an object, so the keys should automatically be converted to camelCase. When deserializing this "object" to a .NET dictionary again, there's no way to know if its keys actually were converted to camelCase, or if they were camelCase in the first place. You've actually lost data in the transformation.

See this round-trip example:

```csharp
private class Person
{
    public string Name { get; set; } = string.Empty;

    public IDictionary<string, string> Values { get; set; } = new Dictionary<string, string>();

    public override string ToString()
    {
        var values = Values.Select(x => string.Format($"{x.Key}:{x.Value}"));

        return $"Name={Name}, Values=[{string.Join(",", values)}]";
    }
}

public static void Main()
{
    var settings = new JsonSerializerSettings
    {
        ContractResolver = new CamelCasePropertyNamesContractResolver()
    };

    var values = new Dictionary<string, string>
    {
        { "Age", "27" },
        { "Location", "Stavanger" }
    };

    var person = new Person
    {
        Name = "Kristian Hellang",
        Values = values
    };

    // {"age":"27","location":"Stavanger"}
    var dictionaryTest = JsonConvert.SerializeObject(values, settings);

    // {"name":"Kristian Hellang","values":{"age":"27","location":"Stavanger"}}
    var objectTest = JsonConvert.SerializeObject(person, settings);

    // {[age, 27]} {[location, Stavanger]}
    var reverseDictionaryTest = JsonConvert.DeserializeObject<IDictionary<string, string>>(dictionaryTest);

    // {Name=Kristian Hellang, Values=[age:27,location:Stavanger]}
    var reverseObjectTest = JsonConvert.DeserializeObject<Person>(objectTest);
}
```

If you want to retain casing, you could easily do as others has done:

```csharp
with.ApplicationStartup((x, y) => JsonSettings.RetainCasing = true);
```

Makes sense?

PR #1525, opened 2014-05-05
Support early Flush() of AspNet responses

csainty:
Created a wrapper stream for the AspNet response.
Calling `Flush()` on the `Stream` defers to `HttpResponse.Flush()`.

it should be noted that for this to work you also need to set `StaticConfiguration.DisableErrorTraces = true;` as it buffers responses.

Thoughts?

### DIFF
--- a/src/Nancy.Hosting.Aspnet/Nancy.Hosting.Aspnet.csproj
+++ b/src/Nancy.Hosting.Aspnet/Nancy.Hosting.Aspnet.csproj
@@ -116,6 +116,7 @@
     <Compile Include="NancyFxSection.cs" />
     <Compile Include="NancyHandler.cs" />
     <Compile Include="NancyHttpRequestHandler.cs" />
+    <Compile Include="NancyResponseStream.cs" />
     <Compile Include="TinyIoCAspNetExtensions.cs" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


thecodejunkie:
@csainty @grumpydev @NancyFx/most-valued-minions could we see any issues with pulling this in?

khellang:
First of all, it needs a rebase :wink: I'm not sure I understand how this would help, since Nancy uses `Action<Stream>` instead of just `Stream`. The delegate won't be invoked until the absolute end of the pipeline. This means that all (?) work has already been performed and the response would be flushed immediately anyway, right?

khellang:
Any update on this?

csainty:
Haven't put any time towards it sorry.
Out of the MS world at the moment and don't have Windows or VS readily available.

When I was playing around with it, it certainly worked. I'd set up a `Thread.Sleep()` in one of the ViewModel property getters - it would flush the head of the template, wait at the property then send the rest.
So Nancy wasn't getting in the way.

https://github.com/csainty/Veil/blob/master/Src/Nancy.ViewEngines.Veil/VeilViewEngine.cs#L49

khellang:
@csainty Yes, *within* the `Contents` delegate, you'd be able to flush "early". But that means that Nancy has already done a whole lot of work (including executing the route handler). Are you thinking that this should be used primarily by view engines? If that's the case, the I think this is fine. Otherwise, I'm not sure I see much value.

csainty:
My only consideration was ViewEngines, ie Veil. :stuck_out_tongue:

khellang:
If you'd like to rebase this. I'll pull it ASAP, then :smile:

khellang:
Replaced by #2083

PR #1549, opened 2014-05-29
Turn off output buffering, allowing Nancy to control it

washcroft:
Under ASP.NET hosting, output buffering is enabled by default and must be disabled on its Response object if it isn't to be used (regardless of any IIS settings).

Unless this is disabled at this point (since there is no way to control it at runtime), output will always be buffered and ultimately exhaust available memory when dealing with large outputs (i.e. streamed file downloads using GenericFileResponse).


thecodejunkie:
@washcroft We agree that it would be nice to be able to control this behaviour. However we do not agree with the current solution. A better approach would be to add it to the existing [NancyFx](https://github.com/NancyFx/Nancy/blob/master/src/Nancy.Hosting.Aspnet/NancyFxSection.cs) configuration section and let it be controlled from your `*.config` file. The default (and fallback) should be the current behaviour of the buffer, i.e you explicitly have to opt-in to disable it.

You up to the task of changing this? Would love to pull this into 1.3 :yum:

thecodejunkie:
Replaced by #2084

PR #1550, opened 2014-05-29
Cookie and Query string fallback update to PR #1468

jeffdoolittle:
There are situations where resources cannot be retrieved with a request that allows sending headers (direct resource access for images, files, etc.). This modification will allow a fallback to a cookie, or a querystring parameter for token authentication in scenarios where headers are not an option.


jeffdoolittle:
@sloncho let me know if you still want to get this done. I like what @khellang recommends for the configuration of the TokenSource and that would be a small modification to the code from this PR.

khellang:
@sloncho @jeffdoolittle Did you guys fight this out? :stuck_out_tongue_closed_eyes:

jeffdoolittle:
@khellang I didn't hear back. Our team is talking about a couple of significant changes to Token auth:

* use Json Web Tokens (JWT) instead of the custom algorithm currently in place
* provide some expiration/blacklisting options (expire by user, persist tokens and provide a management/admin api/dashboard for revoking tokens, etc.)

I'd love to see if others are interested in contributing and discussing these possible feature changes.

--Jeff

jchannon:
I'm happy for anyone to use this/strip bits out for JWT https://github.com/jchannon/Owin.StatelessAuth

I also have a similar thing for querystring and not being able to use headers here but its probably more basic than your proposal https://github.com/VQComms/Owin.QueryStringToHeader

Both are at the OWIN level though

damianh:
I wonder if there are some security subtleties to falling back to cookie in case of not being able header. If @leastprivilege could like to confirm, afaik mixing cookie and header auth is a big no-no, particularly with respect to APIs. 

I could be talking through my hat too of course.

jeffdoolittle:
The reason @jessetechie and I implemented this with a cookie fallback is because there are times where you want to access a secured resource from an html element (such as ). There's no way to include headers here. Anyone using the fallback on a production system would suffer if this functionality was removed. 

Are there other ways to accomplish this goal that maybe we're unaware of?

leastprivilege:
You should have a strict separation between implicit authentication mechanisms (typically browser based - e.g. basic or windows auth, cookies) and explicit authentication mechanisms (e.g. putting a credential/token explicitly on a header or query string param). Since you don't have control over over implicit mechanisms, they are often prone to CSRF attacks.

Implicit is well understood for web UI - but has issues with APIs. That's why I would recommend allowing explicit for APIs only (unless you are fully aware of the implications and implement anti CSRF on **every** endpoint).

An automatic fallback from explicit to implicit is considered harmful.

jeffdoolittle:
So where does this leave us? My sense is that there are legitimate situations where you need to access  resources requiring authorization from an html element. In such a situation, headers are not going to work. I can think of lots of potential work arounds but none of them match the simplicity of the fallback mechanism. If the fallback is considered harmful, it would be helpful to know more about why and what the threat and the risk of the threat entails.

leastprivilege:
You are absolutely right - these situations totally exist.

But as I said - there are explicit and implicit authentication mechanisms. They are fundamentally different.

It is OK to have authentication middleware for implicit (e.g. cookies) and middleware for explicit (e.g. bearer tokens on headers or query strings - or from wherever). I would just not recommend to do an automatic fallback across categories - namely from explicit to implicit.

damianh:
My vote is to close this as 'wontfix'.

horsdal:
A fallback from header to query string would OK, wouldn't it? -Both are in the explicit category, right?

jeffdoolittle:
At this point I'm with @damianh 

If someone desires a fallback, maybe they should just implement it themselves instead of this being added directly to the Nancy project.

@horsdal your recommendation still doesn't address the scenario I've described above regarding hrefs or src attributes in html elements.

jessetechie:
Yeah, this doesn't need to be in the Nancy project after all.  While waiting for this PR we've come up with our own solution.  We wrote a TokenHeaderFallback class:

```
    public static class TokenHeaderFallback
    {
        public static Func<NancyContext, Response> Hook()
        {
            return context =>
                {
                    var token = context.Request.Headers.Authorization
                        ?? ExtractTokenFromCookie(context.Request)
                        ?? ExtractTokenFromQuery(context.Request);

                    if (token != null)
                    {
                        context.Request.Headers.Authorization = "Token " + token;
                    }

                    return null;
                };
        }

        private static string ExtractTokenFromCookie(Request request)
        {
            ...
        }

        private static string ExtractTokenFromQuery(Request request)
        {
            ...
        }
    }
```

Then in Bootstrapper.cs, we add the hook to the beginning of the BeforeRequest pipeline:

```
    public class Bootstrapper : DefaultNancyBootstrapper
    {
        protected override void RequestStartup(TinyIoCContainer container, IPipelines pipelines, NancyContext context)
        {
            base.RequestStartup(container, pipelines, context);

            TokenAuthentication.Enable(pipelines, new TokenAuthenticationConfiguration(container.Resolve<ITokenizer>()));
            pipelines.BeforeRequest.AddItemToStartOfPipeline(TokenHeaderFallback.Hook());
        }
    }
```

So this PR can be closed.  Thanks!

damianh:
Thanks for sharing your solution Jesse ;)

horsdal:
Cool, @jessetechie. If you want to share, maybe that solution could be packaged up as an add on nuget. Would that make sense?

horsdal:
@jeffdoolittle I see your point.

damianh:
Think a gist + blog post would suffice.

jeffdoolittle:
and that's how a PR dies! thanks @jessetechie for the idea which can now live on as a gist perhaps? and thanks to everyone else for the feedback.

jessetechie:
Here's the gist:
https://gist.github.com/jessetechie/5275c0f6c9f8264e9e76

PR #1560, opened 2014-06-09
Negotiation throws in status code handler

### LINKED ISSUE #1855: Should this test for custom error handler be failing?
i've recently been upgrading my Nancy 0.23.1 references up to 1.1 and all my tests pass save for one. 

There seems to be a [related breaking change](https://github.com/NancyFx/Nancy/pull/1653) moving Negotiation to the `INancyModule` that i suspect i need to adjust my test for, but i'm not sure how my test should look in light of this change. And i'm pretty sure my error handler doesn't work the way it's supposed to now.

Would anyone mind taking a look at it and seeing what's wrong? Can you suggest a fix?

:trollface: In vb because i know you love it :trollface: 

:cry: also that's what i often have to code in here at work :cry:

My test:

```vb
    <Fact> _
    Public Sub Internal_server_error_returns_500_page()
        Dim errorView As String = "Error/500"
        Dim errorModule As New ConfigurableNancyModule(Sub(use)
                                                           use.Get("/", Function()
                                                                            Return HttpStatusCode.InternalServerError
                                                                        End Function)
                                                       End Sub)

        Dim testViewFactory = Substitute.For(Of IViewFactory)()
        testViewFactory.RenderView(errorView, Arg.Any(Of Object), Arg.Any(Of ViewLocationContext)) _
            .Returns(errorView)

        Dim browser As New Browser(Sub(use)
                                       use.ApplicationStartup(Sub(container, pipelines)
                                                                  pipelines.OnError.AddItemToEndOfPipeline( _
                                                                      Function(ctx, exception)
                                                                          ctx.Items.Add("OnErrorException", exception)
                                                                          Return ctx.Response
                                                                      End Function)
                                                              End Sub)
                                       use.ViewFactory(testViewFactory)
                                       use.Module(errorModule)
                                       use.StatusCodeHandler(Of InternalServerErrorHandler)()
                                   End Sub)

        Dim response = browser.Get("/", Sub(use)
                                            use.HttpRequest()
                                        End Sub)

        Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode)
        Assert.True(response.Body.AsString.Contains(errorView))
        Ensure.HTML(response)
    End Sub
```

My custom error handler:

```vb
    ''' <summary>
    ''' If there was an <c>HttpStatusCode.InternalServerError</c>, then send the user a friendly error message
    ''' </summary>
    ''' <param name="statusCode">The HTTP status code of the current response</param>
    ''' <param name="context">The current nancy context</param>
    Public Sub Handle(statusCode As HttpStatusCode, context As NancyContext) Implements IStatusCodeHandler.Handle
        Try
            Dim response = New Negotiator(context)

            ' Log a 500 error to the trace output which will show up in the Nancy debugging tools.
            Dim exceptionDetails As String = context.GetExceptionDetails
            context.WriteTraceLog(Function(s) s.AppendLine(exceptionDetails))

            ' Build up the ErrorViewModel
            Dim errorVM As New ErrorPageViewModel
            If context.Items.ContainsKey("OnErrorException") Then
                Dim ex As Exception = context.Items("OnErrorException")
                errorVM.ErrorMessage = ex.Message
                errorVM.FullException = ex.ToString
                errorVM.StackTrace = ex.StackTrace
            Else
                errorVM.ErrorMessage = exceptionDetails
            End If

#If DEBUG Then
            errorVM.Debug = True
#End If

            ' Negotiate the response
            response = response.WithModel(errorVM).WithStatusCode(statusCode).WithView("Error/500")

            ' Set the final response to the negotiated response we just made
            Dim errorResponse As Response = _responseNegotiator.NegotiateResponse(response, context)

            context.Response = errorResponse

        Catch ex As Exception
            ' Log a 500 error to the trace output
            context.WriteTraceLog(Function(s) s.AppendLine(String.Format("Exception hit while processing an issue: {0} -- {1}", ex.Message, ex.StackTrace)))

        End Try
    End Sub
```

---

**PS**: also, i'm guessing i should post support-y questions here?

khellang:
The negotiation throws when used in a status code handler and no module is found (404). This was discovered in http://stackoverflow.com/questions/24112091/automatic-content-negotation-when-assigning-to-response.

The `NegotiationContext` is `null` because it's set in the [`DefaultNancyModuleBuilder`](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Routing/DefaultNancyModuleBuilder.cs#L60), which never gets invoked.

This PR sets the `NegotiationContext` of the `NancyContext` in the constructor and makes the property setter `private`. This way there will always be a negotiation context available.

@Grumpydev and @TheCodeJunkie, what do you think?

Fixes #1855. Closes #1663.

### DIFF
--- a/src/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
+++ b/src/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
@@ -5,6 +5,8 @@ namespace Nancy.Tests.Functional.Tests
     using System.IO;
     using System.Linq;
     using Cookies;
+
+    using Nancy.ErrorHandling;
     using Nancy.IO;
     using Nancy.Responses.Negotiation;
     using Nancy.Testing;
@@ -654,6 +656,19 @@ namespace Nancy.Tests.Functional.Tests
             Assert.Equal(HttpStatusCode.SeeOther, result.StatusCode);
         }
 
+        [Fact]
+        public void Can_negotiate_in_status_code_handler()
+        {
+            // Given
+            var browser = new Browser(with => with.StatusCodeHandler<NotFoundStatusCodeHandler>());
+
+            // When
+            var result = browser.Get("/not-found");
+
+            // Then
+            Assert.Equal(HttpStatusCode.NotFound, result.StatusCode);
+        }
+
         private static Func<dynamic, NancyModule, dynamic> CreateNegotiatedResponse(Action<Negotiator> action = null)
         {
             return (parameters, module) =>
@@ -776,6 +791,27 @@ namespace Nancy.Tests.Functional.Tests
 
             private class Foo
             {
+            }
+        }
+
+        private class NotFoundStatusCodeHandler : IStatusCodeHandler
+        {
+            private readonly IResponseNegotiator responseNegotiator;
+
+            public NotFoundStatusCodeHandler(IResponseNegotiator responseNegotiator)
+            {
+                this.responseNegotiator = responseNegotiator;
+            }
+
+            public bool HandlesStatusCode(HttpStatusCode statusCode, NancyContext context)
+            {
+                return statusCode == HttpStatusCode.NotFound;
+            }
+
+            public void Handle(HttpStatusCode statusCode, NancyContext context)
+            {
+                var error = new { StatusCode = statusCode, Message = "Not Found." };
+                context.Response = this.responseNegotiator.NegotiateResponse(error, context);
             }
         }
     }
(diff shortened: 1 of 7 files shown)


horsdal:
If it's breaking it should go v. 2 if we're doing sem ver.

JeanTerblanche:
Using this PR, and an example as per @jchannon ([gist](https://gist.github.com/jchannon/6fe235343fe2e5c1d845), I get a null ref on the MaterialisingResponse ctor when I use Hosting.AspNet

khellang:
@JeanTerblanche Does that mean this PR broke anything?

khellang:
@JeanTerblanche BTW, is it the same as #1822?

JeanTerblanche:
@khellang Yes, sorry #1822 does seem more relevant. I have added some more details there.

khellang:
@horsdal I've reverted the private setter, so this shouldn't be a breaking change anymore :smile:

Issue #1565, opened 2014-06-12
Int Route Constraint Still Matching for Long


RPM1984:
I have a route like this:
`Get["/details/{value:int}"]`

But if i pass a value like this:
`/details/1111111111111111111`

It still matches, and errors with the following message:
Value was either too large or too small for an Int32

Which is valid. BUT, it shouldn't have matched in the first place, according to the [documentation here](https://github.com/NancyFx/Nancy/wiki/Defining-routes#route-segment-constraints)

PR #1569, opened 2014-06-14
Added new long route constraint and updated the int route constraint to int

phillip-haydon:
Fix for #1565 because @mat-mcloughlin was too slow.

Note:

If you have two constraints on the same URL for `int` and `long` then first in wins since the scoring will be the same. But there's no way for us to truly know which route to pick.

If the user puts in 1000 but wants it to hit the long route then it may pick the int route.

But we can't prevent this, and the user is stupid if they're making routes like this anyway.

### DIFF
--- a/src/Nancy.Tests/Unit/Routing/ConstraintNodeRouteResolverFixture.cs
+++ b/src/Nancy.Tests/Unit/Routing/ConstraintNodeRouteResolverFixture.cs
@@ -24,6 +24,23 @@
         }
 
         [Fact]
+        public void When_int_is_larger_than_max_int_should_has_no_match()
+        {
+            var result = this.browser.Get("/intConstraint/" + long.MaxValue);
+
+            result.StatusCode.ShouldEqual(HttpStatusCode.NotFound);
+            result.Body.AsString().ShouldEqual("");
+        }
+
+        [Fact]
+        public void Should_resolve_long_constraint()
+        {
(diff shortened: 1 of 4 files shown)


khellang:
Added the new constraint to the wiki :smile:

khellang:
How is this a breaking change?

phillip-haydon:
https://github.com/NancyFx/Nancy/pull/1569/files#diff-beb661f87d0a8be138928b34c1530c1fR8

Because of this, the 'int' constraint changed from `long` to `int` and a new 'long' constraint was added.

If anyone was using `int` expecting larger than Int32 values it will break for them.

khellang:
:ok_hand:

drankard:
What is the reason for this "fix" ?

A lot of refactoring, checking, casting and/or double routing is needed now.

I don't agree in the statement that "the user is stupid"
Most NUMBER routes are resource identifiers living in a database, that span from small int to long, hence called NUMBER, why not add a boxing datatype or let it be up to the developer to do the casting. if really needed ?

Am i missing something ?

phillip-haydon:
@drankard we could have added 'number' but then we have `int` and `number`, removing `int` we now have a major breaking change, and a higher barrier to entry when MVC supports `int` and `long`. jchannon also agreed with int/long approach further down. And while we could let the developer do the casting the request would be more expensive for them to execute the module definition rather than just deny the request up front.

The check for us is not expensive. Its potentially expensive for the module.

PR #1617, opened 2014-07-17
Allow configuring reading authentication token from request query.

sloncho:
Sometimes it's not possible (or difficult) for an the consumer to set the auth header with a token. Example: JS app, which creates "img" tags with links to put in the DOM.

As we still want to be able to secure the images, it's much better if the JS adds the token as part of the query to the image URL.

This change adds a configuration option to change where from the authentication token is read - the Authorization header, or Authorization query parameter.

The default (header) behavior is preserved, so this should not introduce breaking change to existing apps.


thecodejunkie:
@sloncho Thanks for this pull-request :heart: for the time being we decided that the functionality in #1550 was simple enough to move forward for the time being. :heart:

PR #1627, opened 2014-07-22
Move modules registration after all other registrations

sloncho:
Move module registrations after all other registrations.

That way the container will have everything in place to be used if one overrides RegisterModules.


khellang:
I can't see why this would break anyone? If anyone has any objections to this, say "AYE" :hand:. Otherwise I'm pulling this once @sloncho has rebased it :smile:

grumpydev:
Can't see why it would break anything.. but I would also bet that it does ;)

phillip-haydon:
Looks fine to me.

khellang:
Replaced by #2082

khellang:
Thanks, @sloncho! :+1:

PR #1644, opened 2014-08-04
Adding Metadata conventions

jchannon:
This is an attempt to add the ability to override Metadata conventions as previously it was hard wired to DefaultMetadataModuleConventions.

You can now implement `IMetadataConvention` and implement it to how you want the system to find metadata modules.

This is the first attempt before we decide along with @liddellj how best to do this. 

Tests will fail but just wanted you to take a look at the propsed API


thecodejunkie:
@jchannon @liddellj if I understand this correctly, we should ditch this pull-request and perhaps revisit it using a different approach (the one that @liddellj  suggested) ?

liddellj:
Yes - I think that's appropriate. Might be worth opening a separate PR referencing this one so the history is retained?

Unfortunately I'm not developing on .NET anymore so it's very unlikely I'll get a chance to do any of this. I'll mention it to my co-workers to see if I can find a new home for it.

liddellj:
Seems I already created a separate issue for this: https://github.com/NancyFx/Nancy/issues/1729

thecodejunkie:
@liddellj Cool. Let me be blunt. Any plans on acting on that or should we close it until the impulse comes back? Currently culling the issues/pull-requests lists very hard and only want to keep stuff that is actually going to happen in a very near future :yum:

liddellj:
Kill it.

PR #1663, opened 2014-08-26
Populated NegotiationContext when NancyContext is created

jchannon:
If you have a `StatusCodeHandler` that responds to exceptions and returns a negotiated response the `NegotiationContext` is null when an exception occurs before a module is built.  `NegotiationContext` is needed in the  `WithModel` method. The `DefaultNancyModuleBuilder` is where it currently only gets set. I have set the `NegotiationContext` now in `DefaultNancyContextFactory` when the `NancyContext` is created.


jugglingnutcase:
OK, now i'm hitting up against this :cry:. i can give up Negiotiation in my custom error handler, but i'd like not to.

Any way i can help with this?

PR #1668, opened 2014-08-31
Break when status code has been handled

### LINKED ISSUE #1667: Custom IStatusCodeHandler's response gets overriden by DefaultStatusCodeHandler's Response
(as noted in the Jabber chatroom but I had to run :-/ ):

I've noticed a somewhat odd behaviour regarding NancyFx's usage of custom IStatusCodeHandler implementations: whenever a json content-type / accept-header Request is made and my Nancy module  returns a (Json)Response with a 404 / NotFound header, the client does in fact get the text/html standard 404 page of NancyFx (the one inside the Resource).

When debugging and checking what's happening inside my own IStatusCodeHandler (that does handle 404s) I still do see my (json) response being.. well.. there & the context's .Response, but when the request is fully processed/finished, the client does get the default text/html one as mentioned above.

Obviously something is going on after my IStatusCodeHandler is invoked and when digging a bit deeper, I saw that the iteration over the NancyEngine's statusCodehandlers is not stopped/break;'ed once one of the handlers does handle the error and I 'think' that's why the default one gets invoked after mine is.. and overwrites my IStatusCodeHandler's Response (see https://github.com/NancyFx/Nancy/blob/6ceb54daec2dc230ab6fe55b367d3837e262c1db/src/Nancy/NancyEngine.cs#L195).

Or maybe I am just not understanding it correctly - is there any way to prevent Nancy to return its text/html 404 (and 500) error pages and thereby overriding 'my' responses?

khellang:
Fixes #1667. I don't think it makes sense to run through all status code handlers even after one has been called. This will try to find a custom handler and fall back on the `DefaultStatusCodeHandler` if no custom handler is found.

### DIFF
--- a/src/Nancy/NancyEngine.cs
+++ b/src/Nancy/NancyEngine.cs
@@ -197,13 +197,15 @@
                 return;
             }
 
-            foreach (var statusCodeHandler in this.statusCodeHandlers)
+            var handler = this.statusCodeHandlers
+                .FirstOrDefault(x => x.HandlesStatusCode(context.Response.StatusCode, context));
+
+            if (handler == null)
             {
-                if (statusCodeHandler.HandlesStatusCode(context.Response.StatusCode, context))
-                {
-                    statusCodeHandler.Handle(context.Response.StatusCode, context);
-                }
+                return;
             }
+
+            handler.Handle(context.Response.StatusCode, context);
         }
 
         private Task<NancyContext> InvokeRequestLifeCycle(NancyContext context, CancellationToken cancellationToken, IPipelines pipelines)
--- a/src/Nancy/NancyEngine.cs
+++ b/src/Nancy/NancyEngine.cs
@@ -197,9 +197,17 @@
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


khellang:
This could potentially be a breaking change if people rely on multiple status code handlers  to handle a specific status code.

src/Nancy/NancyEngine.cs
```diff
@@ -187,13 +187,23 @@ private void CheckStatusCodeHandler(NancyContext context)
...
+                .ToList();
+
+            var defaultHandler = handlers
+                .FirstOrDefault(x => x is DefaultStatusCodeHandler);
+
+            var customHandler = handlers
+                .FirstOrDefault(x => !(x is DefaultStatusCodeHandler));
+
+            var handler = customHandler ?? defaultHandler;
+            if (handler == null)
```
jchannon: this will never be null?

khellang: If the `DefaultStatusCodeHandler` has been removed, it could be `null`. Better safe than sorry :wink:

PR #1673, opened 2014-09-05
Make sure question mark is included in query string

### LINKED ISSUE #1672: Return URL with query string misses question mark
When I open a URL like `/Page/Secured?foo=bar` which requires authentication and I get redirected to the login page, the `returnUrl` parameter lacks the question mark, so I'm redirected to `/Login?ReturnUrl=/Page/Securedfoo%3dbar` instead of `/Login?ReturnUrl=/Page/Secured%3ffoo%3dbar`. Of course, this leads to an incorrect redirection after having logged in. It looks like the error is in `FormsAuthentication.GetRedirectToLoginHook` where the URL is constructed.

khellang:
Fixes #1672. This was edited in the GitHub GUI so bear with me :wink:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ You should **never** work on a clone of master, and you should **never** send a 
 
 While you're working away in your branch it's quite possible that your upstream master (most likely the canonical NancyFx version) may be updated. If this happens you should:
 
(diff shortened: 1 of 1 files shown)


der-andy:
This pull request does _not_ fix #1672. The question mark is still missing in 1.0 stable.

Issue #1676, opened 2014-09-06
Path containing escaped / (slash) cannot be handled by DefaultRouteResolver


rasmus-beck:
Consider the path /first-part/second%2fpart, which has two segments (first-part and second/part).

In combination with the route:

Get["/first-part/{part}"] = parameters => return 200;

A get request to the path mention returns 404, because the default route resolver has a bug, if any of the path segments contains an escaped slash.

The first thing the route resolver does, is to decode the entire path, before splitting it up into segments, this means that a path segment, that wasn't meant to be there, is introduced.

https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Routing/DefaultRouteResolver.cs#L49

I'll try to create a failing unit test for this at least. But too tired right now.

Issue #1280, opened 2013-10-20
Incoming URL is decoded twice (at least with Nancy.Hosting.Self)


ChrisBrandhorst:
When using `Nancy.Hosting.Self` (at least in this case), an incoming URL is decoded using `HttpUtility.UrlDecode` twice.

First, in `NancyHost.cs#ConvertRequestToNancyRequest`. The resulting `Request` is then passed to the routeresolver, where in `DefaultRouteResolver#ResolveResult`, it is decoded again.

This is a problem when you have URLs like
```
/containers/123/artist/AC%252FDC
```
combined with the following route:
```
/containers/{id}/{filter*}
```
In this case, the `filter` part is a string I would like to split on the slash again (using the slash like this looks nicer in the UI). However, because of the double decoding, the resulting value for `filter` is `AC/DC`, where I would like to have `AC%2FDC`

kenotron:
This is a pretty serious flaw. Why don't we create a UrlPathDecode that is mostly what UrlDecode does in HttpUtility and use THAT inside the parts where you are decoding a path? So, the only thing you need is to modify the DefaultRouteResolver.cs to use that. For everyone else out there, you can get by with this blog's recommendations:

http://programmaticallyspeaking.com/the-mysteriously-escaped-request-path-in-aspnet.html

simon-brooke:
I've hit this bug this morning, and, assuming that double-decoding was going on, tried both:

http://localhost:3579/Json/PointsMatching/%2B

and

http://localhost:3579/Json/PointsMatching/%252B

Both of these were received serverside as spaces, so it's not quite as simple as double-decoding, since double-decoding should decode

'%252B' to '%2B' and subsequently to '+'

I had planned heroically to contribute a patch, but having read through the comments above I see that it's substantially more subtle than I had supposed!

taharrison:
Hey guys.

I just tried to patch the behaviour of url-decoding in DefaultRouteResolver to get around a bug with the treatment of encoded `/`s.  But I've closed my pull request, because it relied on the DefaultRouteResolver being responsible for URL Decoding, which from the discussion here it seems it shouldn't be.

I carefully documented the issues I saw though, so am reproducing the summary and test cases here so that they can help inform future work on this issue:

> Currently the DefaultRouteResolver url-decodes capture groups before identifying where the capture groups begin and end (i.e. splitting on "/")
>
> This leads to odd behavior around encoded "/"s.

> Examples:
> 
> With routes
> 
[...]

PR #1678, opened 2014-09-07
Proposed fix for #1280

rasmus-beck:
This PR proposes a fix for #1280, and #1676.

It is not ready to be merged, as I need a few more tests. 

Input for tests are very welcome.


thecodejunkie:
The OWIN v1 specification does mention encoding requirements for various environment keys
http://owin.org/spec/spec/owin-1.0.0.html#PercentEncoding

rasmus-beck:
Hmmm... if I read this right, then we a decoded path when using OWIN. Well that kind of sucks doesn't it. We want the encoded path, so we can tell what forward slashes are path separators and which are not. Am I reading this right?

grumpydev:
It does appear to be the opposite of what we want :/

rasmus-beck:
What's the verdict then? What are we going to do?

damianh:
Capturing a segment that has a  '/' encoded in it as a route param is never going to work. Not just on Nancy, but on _all_ webservers and web frameworks.

> Most web server implementations will perform percent-decoding on paths in order to perform request routing (as they rightly may, see: RFC 2616 section 5.1.2, also 3.2.3), and OWIN follows this precedent. 

http://owin.org/spec/spec/owin-1.0.0.html#PercentEncoding

> If the Request-URI is encoded using the "% HEX HEX" encoding [42], the origin server MUST decode the Request-URI in order to properly interpret the request.

http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.1.2

> For example, the following three URIs are equivalent:

      http://abc.com:80/~smith/home.html
      http://ABC.com/%7Esmith/home.html
      http://ABC.com:/%7esmith/home.html

http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.2.3

This means Nancy will (should) never see a uri in the form http://host/foo%2Fbar/baz

If you want a `/` in a route param you will have to do a workarounds http://stackoverflow.com/a/26800648/1010

damianh:
Given the above, #1676 this isn't fixable.

grumpydev:
Yeah, having dug into the OWIN issue a bit more it looks like it is broken, but so is every other platform out there, which unfortunately means this isn't going to be fixable in a generic way (especially considering we are heading towards OWIN only hosting eventually) and should be "fixed" with education (don't put encoded slashes in paths).

rasmus-beck:
I guess we might as well close this PR, and all related issues. I will see if I can write a paragraph on this for the wiki.

damianh:
Yeah agree. Frustrating after all that effort. Of which, many thanks Rasmus.

rasmus-beck:
Yes frustrating indeed, but that's how it goes sometimes. You have to get to that point where things just won't fit together to find out that it isn't possible.

jchannon:
So this means we still have to have code likes this in projects?
```
var email = (string)parameters.email;
email = email.Replace(" ", "+");
```

rasmus-beck:
Depends on what host you use. Do you use Asp.Net host, then you can still use the web.config work around, but you really shouldn't go down this route. I would suggest you do your own escaping in the urls you provide. So when you return a `/` escape it with something like `__s__`, or whatever you prefer.

The web.config work around is this (ref: http://danielwertheim.se/2014/10/11/uri-behaves-differently-in-net4-0-vs-net4-5/):
```xml
<configuration>
  <uri>
    <schemeSettings>
      <add name="http" genericUriParserOptions="DontUnescapePathDotsAndSlashes"/>
    </schemeSettings>
  </uri>
</configuration>
```

jchannon:
Well this is incoming, so when you have a `/person/fred+1@home.com` the Nancy route parameter value will be decoded twice so have to put the + back in

rasmus-beck:
Sorry... didn't actually look at your example @jchannon. 

I don't think that the fix proposed in this PR, or the web.config one, would do anything for you in that case. A plus is not a path character in windows. Dots and slashes are, and that's why they are treated differently, to prevent directory traversal / injection attacks.

If you try to double escape the query string you are just going to get a 404.11 status code back.

I can't remember where I got this list of characters from, but these seem to cause trouble for IIS `<>*%:&/\_`.

In your example the plus isn't even escaped once. You can't really expect that to work any other way than what you have explained. In a URL `+` means space, so you told it to put a space in there, not a plus sign. Whomever sent that request should at least try to escape once.

jchannon:
Good point :+1:

rbirkby:
To help others like me asking the same question, this PR does not cater for GMail Address Alias email addresses contained in path segments. For example:

    /a/b/c/test+account@gmail.com/d/e/f

Even though + is not a special character in URL paths, this PR [uses](https://github.com/NancyFx/Nancy/pull/1678/files#diff-1f270b3a7e4b130880a6d3535ebb232bR101) HttpUtility.UrlDecode() to decode each path segment. And UrlDecode will decode the + into a space character.

rasmus-beck:
@rbirkby `+` is a special character in a URL. It means space. Look at this comment:
https://github.com/NancyFx/Nancy/pull/1678#issuecomment-71070613

rbirkby:
@rasmus-beck The + is only special in a querystring. It is a valid character in the URL path (have a read of the spec).

Also:

```csharp
using System;
using System.Net;
					
public class Program
{
	public static void Main()
	{
		CheckValidity("http://a.b.com/address+alias@gmail.com/path");
		CheckValidity("http://a.b.com/address alias@gmail.com/path");
	}
	
	public static void CheckValidity(string url)
	{
		var uri = new Uri(url);
		Console.WriteLine("{0} ({1}): {2}", url, string.Join(", ", uri.Segments), Uri.IsWellFormedUriString(url, UriKind.Absolute));
	}
}
```

    http://a.b.com/address+alias@gmail.com/path (/, address+alias@gmail.com/, path): True
    http://a.b.com/address alias@gmail.com/path (/, address%20alias@gmail.com/, path): False

See how a space is encoded as %20, not as +, and how a + is not special in any way.

rasmus-beck:
@rbirkby Ahhh good point. I really should learn to read. Anywho, closed the PR since it fails to follow multiple standards.

taharrison:
@damianh, 
I think you've misinterpreted how %2f should behave.

from http://www.w3.org/Addressing/URL/uri-spec.html

> Example 2
> 
> The URIs
> 		http://www.w3.org/albert/bertram/marie-claude
>
> and
> 		http://www.w3.org/albert/bertram%2Fmarie-claude
>
> are NOT identical, as in the second case the encoded slash **does not have** hierarchical significance.

I believe that where you say: 
> the origin server MUST decode the Request-URI in order to properly interpret the request.

This isn't blind URL decoding.  It must first respect the url segments split by the unencoded "/"s, and then decode the segments.

Also, from http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.2.3 which you also quote:

> Characters **other than those** in the "reserved" and "unsafe" sets (see RFC 2396 [42]) are equivalent to their ""%" HEX HEX" encoding.

(added emphasis is mine.)  "/" is a character in reserved the reserved set.  It is not equivalent to "%2f"


shoutout to @rasmus-beck @khellang also.

khellang:
I think @damianh's point was that *Nancy* will never see an encoded URI because the host will always decode it before it surfaces the request...

khellang:
I just tested the following using `Microsoft.Owin.Host.HttpListener` and `Microsoft.Owin.Hosting`:

```csharp
public static class Program
{
    private const string Url = @"http://localhost:1337";

    public static void Main(string[] args)
    {
        using (WebApp.Start(Url, Configure))
        {
            Console.WriteLine("Listening at '{0}'...", Url);
            Console.ReadLine();
        }
    }

    private static void Configure(IAppBuilder app)
    {
        app.Run(ctx => ctx.Response.WriteAsync(ctx.Request.Uri.OriginalString));
    }
}
```

And the results:

 - `http://localhost:1337/albert/bertram%2Fmarie-claude` -> `http://localhost:1337/albert/bertram/marie-claude`
 - `http://localhost:1337/albert/bertram/marie-claude` -> `http://localhost:1337/albert/bertram/marie-claude`

As you can see, the `HttpListener` decodes the URI before passing it to the application, thus we will never know what the original URI was.

grumpydev:
Yep, that's the problem, the attempted fix was to make sure nancy was responsible for decoding so we could create the segments correctly, but unfortunately it was out of our hands in several instances so there's not a whole lot we can do about it.

khellang:
However, there *are* methods to get a hold of the raw URI, but it's far from host agnostic (which is kinda the point of OWIN).

If you change the `Configure` method above to

```csharp
private static void Configure(IAppBuilder app)
{
    app.Run(ctx =>
    {
        object context;
        if (ctx.Environment.TryGetValue("System.Net.HttpListenerContext", out context))
        {
            var httpListenerContext = context as HttpListenerContext;
            if (httpListenerContext != null)
            {
                return ctx.Response.WriteAsync(httpListenerContext.Request.RawUrl);
            }
        }

        return ctx.Response.WriteAsync(ctx.Request.Uri.OriginalString);
    });
}
```

The results are:

 - `http://localhost:1337/albert/bertram%2fmarie-claude` -> `/albert/bertram%2fmarie-claude`
 - `http://localhost:1337/albert/bertram/marie-claude` -> `/albert/bertram/marie-claude`

I'm pretty sure this is possible using System.Web and WCF (:mask:) as well. Maybe it's an idea to propose a `owin.RequestRawUrl` key for OWIN?

damianh:
> Maybe it's an idea to propose a owin.RequestRawUrl key for OWIN?

@khellang something about that make me feel icky... think it'll just lead to confusion. May not be supportable in all hosts either i.e. SystemWeb.

khellang:
@damianh System.Web has the same: http://referencesource.microsoft.com/#System.Web/Abstractions/HttpRequestBase.cs,c676f095dc09e1f5,references

damianh:
ok, make it so then :p

PR #1692, opened 2014-09-18
Changed JsonSettings to use Encoding instead of string

khellang:
Instead of using a simple `string` for JSON charset, this adds a new property, `DefaultEncoding`, which is set to UTF8 by default, and we'll use the [`Encoding.WebName`](http://msdn.microsoft.com/en-us/library/system.text.encoding.webname(v=vs.110).aspx) property instead.


khellang:
Meh,

PR #1695, opened 2014-09-20
Improved handling of exceptions in Hosting.Aspnet (Avoids YSOD)

Jevonius:
While working through some Razor view issues, I noticed that `Nancy.Hosting.Aspnet.NancyHandler.SetNancyResponseToHttpResponse` doesn't handle an exception being thrown in `response.Contents.Invoke`. You therefore get thrown back to YSOD rather than the (nicer) 500 page that Nancy would usually provide.

Although the MaterialisingResponse (#1504) that @grumpydev implemented sort-of offers a workaround for this with the warning of "Copies the existing response into memory, so use with caution.", because of the output buffering mentioned in #1549, you actually get a double-whammy of memory usage if configured to use it (plus you show users your error stack).

This is a first attempt to fix the issue as I couldn't find anything to suggest this (YSOD) is desired behaviour. I've tried to do it in a way that doesn't mean any breaking changes, though I have re-implemented #1668 as I noticed the issue separately (sorry, couldn't work out how to pull it into my branch), which may or may not be breaking. Ideally the two new public methods would be added to `INancyEngine` (and have some tests written!! :worried: ) to avoid the `as NancyEngine` cast and additional `if` block.`

To try and make sure tests are consistent, I'd be interested in applying the same kind of try-catch block to `Nancy.Testing.BrowserResponseBodyWrapper` (and indeed the other Hosting implementations should probably be adjusted). Not much point doing any of that if this PR isn't actually wanted though! :smile:

### DIFF
--- a/src/Nancy.Hosting.Aspnet/NancyHandler.cs
+++ b/src/Nancy.Hosting.Aspnet/NancyHandler.cs
@@ -50,7 +50,7 @@ namespace Nancy.Hosting.Aspnet
             return tcs.Task;
         }
 
-        public static void EndProcessRequest(Task<Tuple<NancyContext, HttpContextBase>> task)
+        public static void EndProcessRequest(Task<Tuple<NancyContext, HttpContextBase>> task, INancyEngine engine = null)
         {
             if (task.IsFaulted)
             {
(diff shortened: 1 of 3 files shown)


khellang:
@Jevonius This (unfortunately) now has a conflic. Probably because #1668 was pulled. Would you like to rebase this, or should we close it and try again? :smile:

khellang:
Hey @Jevonius! We're desperately trying clear the pull request queue in order to release Nancy 1.4 and start working on version 2.0 :sparkles:.

This pull request still has some issues that needs to be addressed in order for us to pull it in:
 - It re-implements the logic that landed with #1668
 - We're not exactly sure if this is the way we want to handle exceptions outside the `NancyEngine` (and its pipeline). Currently, the error handling is left up to the different hosts. The self-host has a [`UnhandledExceptionCallback`](https://github.com/NancyFx/Nancy/blob/3eadbd642a08e15569aef34248ec4ccff5c1f0b6/src/Nancy.Hosting.Self/HostConfiguration.cs#L38), IIS/System.Web has YSOD and OWIN has whatever the host and/or middleware specifies. We would, at some point, like to unify this, but we have a lot of changes planned for Nancy v2 and we're not sure how to handle it yet.

Because of this, I'm closing this pull request for now. Feel free to open it up or submit a new one when you have time.
Maybe we should open an issue and discuss how we want to approach a unified error handling approach for the body delegate?

PS! #2084 has also landed now, so you'll be able to turn off response buffering if you wish.

Thanks! :smile:

Jevonius:
Hi @khellang, no problem with closing this issue - it's not been causing problems for us, and the potential change in approach sounds like it might impact it anyway. If I have some time at some point, I'll review the current state of the code and see if it's worth trying again!

PR #1698, opened 2014-09-22
Mono post options method without body

jjchiw:
Following the issue #1598. I've found that in nancy hosted via aspnet in mono all the request made without a body for any Http Method besides `GET` throws and exception

```
[SocketException (0x274c): A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond]
   System.Net.Sockets.Socket.Receive(Byte[] buffer, Int32 offset, Int32 size, SocketFlags socketFlags) +114
   System.Net.Sockets.NetworkStream.Read(Byte[] buffer, Int32 offset, Int32 size) +698

[IOException: Unable to read data from the transport connection: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.]
   System.Net.Sockets.NetworkStream.Read(Byte[] buffer, Int32 offset, Int32 size) +932
   Mono.WebServer.Mono.WebServer2124797.XSPWorker.Read(Byte[] buffer, Int32 position, Int32 size) +70
   Mono.WebServer.BaseRequestBroker.Read(Int32 requestId, Int32 size, Byte[]& buffer) +345
   Mono.WebServer.XSPWorkerRequest.ReadInput(Byte[] buffer, Int32 offset, Int32 size) +240
   Mono.WebServer.XSPWorkerRequest.ReadEntityBody(Byte[] buffer, Int32 size) +131
   System.Web.HttpRequest.GetEntireRawContent() +1125
   System.Web.HttpRequest.get_InputStream() +151
   System.Web.HttpRequestWrapper.get_InputStream() +41
   Nancy.Hosting.Aspnet.NancyHandler.CreateNancyRequest(HttpContextBase context) +1164
   Nancy.Hosting.Aspnet.NancyHandler.ProcessRequest(HttpContextBase context, AsyncCallback cb, Object state) +131
   Nancy.Hosting.Aspnet.NancyHttpRequestHandler.BeginProcessRequest(HttpContext context, AsyncCallback cb, Object state) +130
   System.Web.CallHandlerExecutionStep.System.Web.HttpApplication.IExecutionStep.Execute() +833
   System.Web.HttpApplication.ExecuteStep(IExecutionStep step, Boolean& completedSynchronously) +188
```

It seems that  `context.Request.InputStream` if there is nothing to read in mono, there is a timeout.

So I just changed that if the expectedRequestLenght is different than 0, then `RequestStream.FromStream`

I had this problem using angular in a CORS environment, because of this

>It all comes down to how browsers manage CORS. When making a cross-domain request in JavaScript that is not "simple" (i.e. a GET request - which explains why the query() function worked), the browser will automatically make a HTTP OPTIONS request to the specified URL/URI, called a "pre-flight" request or "promise". As long as the remote source returns a HTTP status code of 200 and relevant details about what it will accept in the response headers, then the browser will go ahead with the original JavaScript call.

source: http://stackoverflow.com/questions/21221688/angularjs-resource-makes-http-options-request-instead-of-http-post-for-save-me


khellang:
First I was worried that this could break someone (since request body could now be null), but noticed that we check for that in the `Request` ctor, so we should be all good! :sparkles:

PR #1699, opened 2014-09-24
Additional ScanModes in AssemblyTypeScanner

jeff-pang:
Added ScanModes.OnlyNancyNamespace and ExcludeNancyNamespace. This will
check for types with namespaces starting with "Nancy"

Usage:
You can make use of these scanmodes via AppDomainAssemblyTypeScanner, e.g
```
var providerTypes = AppDomainAssemblyTypeScanner
.TypesOf(ScanMode.ExcludeNancyNamespace)
.ToArray();
```
It is possible to override default behaviour by overriding default nancy bootstrapper

Note:
The old pull request #1492 is closed and replaced with this one due to wrongly branched pull request


khellang:
I'm not sure if we need tests for this. We currently don't have any tests for ADATS, and it's super hard to write good tests for AppDomain scanning.

PR #1700, opened 2014-09-24
Exception when resource name duplicates

jeff-pang:
Throws a more meaningful exception when resource name duplicates


khellang:
This is a nice improvement! :heart:

PR #1701, opened 2014-09-24
An option on XmlSettings to not have newlines and whitespace in xml responses

AnthonySteele:
Added IncludeWhiteSpace flag to XmlSettings
Use this in a XmlWriterSettings when writing xml.
In one common case of ours this reduces the response size from 16kb to 13kb.


src/Nancy/Responses/DefaultXmlSerializer.cs
```diff
@@ -39,13 +40,36 @@ public void Serialize<TModel>(string contentType, TModel model, Stream outputStr
         {
             var serializer = new XmlSerializer(typeof(TModel));
 
-            if (XmlSettings.EncodingEnabled)
+            if (XmlSettings.IncludeWhiteSpace)
```
thecodejunkie: Can't the entire `if`-block be removed completely? The only thing that is changed is the use of the `XmlWriterSettings` where only `Indent` is set based on the value of `XmlSettings.IncludeWhatSpace` ?

thecodejunkie:
Added a review comment + could you add Given/When/Then sections to your tests? Thanks :heart:

AnthonySteele:
I must review this some time - I never got all the things working to my satisfaction at the same time.
For the time being, it's not urgent for me.

thecodejunkie:
@AnthonySteele Ok, thanks. Closing this for the time being. I think it would be a more suitable thing for the new configuration API in v2 anyway

AnthonySteele:
Thanks. Sorry that we couldn't get this to work in the current version, hope that we can in V2.

The requirement is basically for "self-describing but compact xml" as most useful in production scenarios.
This is usually done IIRC by adding a custom xml serialiser to the project. I was trying to get something with less custom code!

PR #1708, opened 2014-10-02
Improved fallback converter by allowing binds to object properties

nathanmarlor:
Original issue - could not bind value to object type. Would always result in 'null'.

Removed confusing test - will now throw on any property that cannot be converted. Also modified original test to ensure an object property can be bound.


khellang:
This needs a rebase if you want us to pull it :smile:

horsdal:
This is a breaking change, since exceptions are throw in situations, that would previous result in default vale of null. As I breaking change, it can't go in 1.x.

PR #1734, opened 2014-10-31
Don't crash on an invalid cookie

omeaart:
If, for some reason, someone has a broken cookie, all the requests end in an exception.

It seems logical to just ignore the session if that happens.


khellang:
I like it! :+1:

PR #1735, opened 2014-10-31
Ignore invalid cookie attributes

omeaart:
If a cookie contains an invalid attribute the current code throws an System.IndexOutOfRangeException.
Why are attributes other then HttpOnly and Secure not accepted?

I see this error frequently in my logging, so I propose to just ignore all attributes.


khellang:
Let's get this in! Thanks :heart:

PR #1741, opened 2014-11-07
Allow GenericFileResponse to serve files without extension

ClemPi:
I don't see any reason to filter out files without extensions, serving application/octet-stream content should be fine.


khellang:
:heart:

PR #1757, opened 2014-12-02
DynamicDictionaryValue adjust to TryParse

khalidsalomao:
Adjust TryParse method to allow more convertions. Convertions like Int64 to Int32 were ignored in the previous version, since there was no ending else statement.
Now the last rule is to try a type convertion. 

Note that if this convertion fails any exception will be treated and the default value returned.


khalidsalomao:
I created a new pull request for this adjustments with another feature (nullable convertion) and some tests. See pull request https://github.com/NancyFx/Nancy/pull/1804

PR #1764, opened 2014-12-05
Remove unneeded/bad IModelValidator registrations

sloncho:
Short: As these classes are never resolved from the container, but are created with "new" in the factories, there is no need for the registrations to exist at all.

Long:

These 2 registrations are incomplete, as their ctors expect Type as parameter. This causes IoC container diagnostics to fail (example, Windsor reports missing dependencies).

I have a test which validates if after bootstraper.Initialize() the container is in a consistent state, and that's how I found the problem.


khellang:
Makes sense. Doesn't look like they're used anywhere :+1:

PR #1766, opened 2014-12-11
Forms Authentication Validation - Replace IsValid with EnsureConfigurationIsValid()

xt0rted:
There's two breaking changes here. The first is the removal of the `IsValid` property and the second is changing the exception type in the `Enable` method.


thecodejunkie:
@xt0rted so the only reason this is a breaking change is because of the `IsValid -> EnsureConfigurationIsValid()` change? That is, if we ignore the fact that someone might be catching the `ArgumentException` for `.Enable`, but you now throw `InvalidOperationException`. I think the exception is less trivial change. It's still breaking, but I would be more inclined to pull this into 1.3 if it was the only breaking change. As it stands now it would probably end up in the 2.x time frame.

To be honest I would much rather see is keep `IsValid` and be able to pull this into 1.3

src/Nancy.Authentication.Forms.Tests/FormsAuthenticationConfigurationFixture.cs
```diff
@@ -1,9 +1,12 @@
 namespace Nancy.Authentication.Forms.Tests
 {
+    using System;
+
     using FakeItEasy;
 
     using Nancy.Cryptography;
     using Nancy.Tests;
+
```
khellang: :fire:

src/Nancy.Authentication.Forms.Tests/FormsAuthenticationFixture.cs
```diff
@@ -3,14 +3,16 @@ namespace Nancy.Authentication.Forms.Tests
     using System;
     using System.Linq;
     using System.Threading;
-    using Bootstrapper;
-    using Cryptography;
+
```
khellang: :fire:

src/Nancy.Authentication.Forms.Tests/FormsAuthenticationFixture.cs
```diff
@@ -3,14 +3,16 @@ namespace Nancy.Authentication.Forms.Tests
...
     using System.Linq;
     using System.Threading;
-    using Bootstrapper;
-    using Cryptography;
+
     using FakeItEasy;
-    using Fakes;
-    using Helpers;
-    using Nancy.Security;
+
```
khellang: :fire:

src/Nancy.Authentication.Forms.Tests/FormsAuthenticationFixture.cs
```diff
@@ -3,14 +3,16 @@ namespace Nancy.Authentication.Forms.Tests
...
+    using Nancy.Authentication.Forms.Tests.Fakes;
+    using Nancy.Bootstrapper;
+    using Nancy.Cryptography;
+    using Nancy.Helpers;
     using Nancy.Tests;
     using Nancy.Tests.Fakes;
+
```
khellang: :fire:

src/Nancy.Authentication.Forms/FormsAuthentication.cs
```diff
@@ -1,11 +1,12 @@
 namespace Nancy.Authentication.Forms
 {
     using System;
-    using Bootstrapper;
-    using Cookies;
-    using Cryptography;
-    using Helpers;
+
```
khellang: :fire:

src/Nancy.Authentication.Forms/FormsAuthenticationConfiguration.cs
```diff
@@ -1,6 +1,8 @@
 namespace Nancy.Authentication.Forms
 {
-    using Cryptography;
+    using System;
+
```
khellang: :fire:

src/Nancy.Authentication.Forms/FormsAuthenticationConfiguration.cs
```diff
@@ -68,39 +70,35 @@ public FormsAuthenticationConfiguration(CryptographyConfiguration cryptographyCo
...
-                    return false;
-                }
-
-                return true;
+                throw new InvalidOperationException("CryptographyConfiguration HmacProvider cannot be null.");
             }
         }
     }
-}
+}
```
khellang: Line break, please? https://robots.thoughtbot.com/no-newline-at-end-of-file :innocent:

khellang:
If you fix the nits and rebase this, I'll be here to pull it :wink:

xt0rted:
@khellang  rebased and cleaned up

khellang:
Above and beyond! :sparkles: Thanks, @xt0rted :heart:

PR #1787, opened 2015-01-23
Add nancy dependency to validation nuspecs

xt0rted:
These were missing the core Nancy package as a dependency for some reason


khellang:
The external dependencies should really have a minimum version specified. The Nancy dependency doesn't matter cause its version is set in the build script. Are you up for it?

khellang:
I guess it's best to use the version the projects are built against :smile:

xt0rted:
@khellang do you want me to update the fluent validation nuspec with a version? At the moment we're referencing [v3.4.0](https://github.com/NancyFx/Nancy/blob/600463b0ee9f8d3a465522599bbd9f2b8b527e83/src/Nancy.Validation.FluentValidation/packages.config#L3) while the latest on nuget is [v5.5.0](http://www.nuget.org/packages/FluentValidation/5.5.0)

For now I can't use CI builds since my projects are all 1.0 and the feed is still `0.24-Pre`

PR #1788, opened 2015-01-26
Fixed a bug that prevented transmitting chunked content with Nancy SelfHost Library

mtharrington:
With the latest 1.0 release, attempting to download static content caused an unhandled exception due to the Content-Length being incorrectly added as a header.

Updated the conversion between a Nancy response and an HttpListenerResponse to ignore certain headers that are defined as properties.

### DIFF
--- a/src/Nancy.Hosting.Self/IgnoredHeaders.cs
+++ b/src/Nancy.Hosting.Self/IgnoredHeaders.cs
@@ -1,0 +1,35 @@
+﻿namespace Nancy.Hosting.Self
+{
+
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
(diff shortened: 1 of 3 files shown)


thecodejunkie:
Bonus points for getting the code conventions right :P

netclectic:
Any idea when this will make it into the NuGet package?

xt0rted:
As long as you use `0.24-Pre1469` or newer from the [myget feed](https://www.myget.org/F/nancyfx/) you should get this update. @thecodejunkie any chance of fixing the CI builds so they use v1.0?

khellang:
@xt0rted The CI build version number has been changed to 1.1 and it's building now. There should be new packages (with correct version number) on the MyGet feed Very Soon™ :wink: (See [status here](http://nancy-ci.cloudapp.net/))

PR #1797, opened 2015-02-02
"Address already in use" error in linux

Kibele:
To avoid this error, port numbers or something else should be different...

### DIFF
--- a/src/Nancy.Demo.Hosting.Self/Views/staticview.html
+++ b/src/Nancy.Demo.Hosting.Self/Views/staticview.html
@@ -17,8 +17,8 @@
         </ul>
         <a href="http://localhost:8888/nancy/">http://localhost:8888/nancy/</a><br/>
         <a href="http://localhost:8888/nancy/testing">http://localhost:8888/nancy/testing</a><br/>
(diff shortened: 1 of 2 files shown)


src/Nancy.Demo.Hosting.Self/Views/staticview.html
```diff
@@ -17,9 +17,9 @@
         </ul>
         <a href="http://localhost:8888/nancy/">http://localhost:8888/nancy/</a><br/>
         <a href="http://localhost:8888/nancy/testing">http://localhost:8888/nancy/testing</a><br/>
-        <a href="http://127.0.0.1:8888/nancy/">http://127.0.0.1:8888/nancy/</a><br/>
-        <a href="http://127.0.0.1:8888/nancy/testing">http://127.0.0.1:8888/nancy/testing</a><br/>
+        <a href="http://127.0.0.1:8898/nancy/">http://127.0.0.1:8888/nancy/</a><br/>
```
khellang: Maybe change the link text as well?

thecodejunkie: Yep, the link text needs changing before we pull it :P

PR #1804, opened 2015-02-05
Adjust DynamicDictionaryValue TryParse method to allow more convertions

khalidsalomao:
Adjust DynamicDictionaryValue TryParse method to allow more convertions.
. Convertions like Int64 to Int32 were ignored in the previous version,
since there was no ending else statement.
Now the last rule is to try a type convertion.
. Convertions to a nullable type were also ignored (like Int64 to
Int64?), now there is a test to check if a nullable type convertion is
possible.

Note that if the convertion fails any exception will be treated and the
default value returned.

Some test cases were created for these situations.

### DIFF
--- a/src/Nancy.Tests/Unit/DynamicDictionaryValueFixture.cs
+++ b/src/Nancy.Tests/Unit/DynamicDictionaryValueFixture.cs
@@ -898,6 +898,66 @@
         }
 
         [Fact]
+        public void Should_implicitly_convert_from_int_based_on_given_parameter_type_of_short ()
+        {
+            //Given
+            const int expected = 42;
+            const short notExpected = 100;
+            dynamic value = new DynamicDictionaryValue (expected);
+
+            //When
+            int actual = (int)value.TryParse<short> (notExpected);
+
+            //Then
+            Assert.Equal (expected, actual);
+        }
+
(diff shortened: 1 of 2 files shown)


src/Nancy.Tests/Unit/DynamicDictionaryValueFixture.cs
```diff
@@ -898,6 +898,66 @@ public void Should_implicitly_convert_from_string_based_on_given_generic_type_of
         }
 
         [Fact]
+        public void Should_implicitly_convert_from_int_based_on_given_parameter_type_of_short ()
```
horsdal: Nitpickng: There is usually no spaces before the parenthesis in Nancy code.

horsdal:
:+1:

khalidsalomao:
Hi Horsdal, 

One more thing about DynamicDictionaryValue: the methods "TryParse" and "TryConvert" share some similarities.... Should I try to merge the common funcionalities of both methods? Or do you guys prefer to leave as it is?

best!

horsdal:
Hmm, not sure they are similar enough for unification.

khalidsalomao:
OK!

khellang:
Love it! :heart:

PR #1805, opened 2015-02-05
DynamicDictionary review

khalidsalomao:
Bug fix:
Two bug fixes for ContainsKey and TryGetValue. They should use GetNeutralKey to normalize the key before a dictionary search.

Performance tuning:
GetNeutralKey : Since GetNeutralKey is executed for every operation in DynamicDictionary this small performance gain doen't hurt.
To use character Replace is a tiny faster (2x)... and since the use is internal to DynamicDictionary, this change shouldn't break anything...

internal dictionary constructor: The dictionary performance with StringComparer.Ordinal is faster than InvariantCulture (6x-7x).
Since this is a hash generation, shouldn't have any impact.


khalidsalomao:
I am afraid the change on the internal dictionary to use the Ordinal StringComparer could be a little controvertial...

I still believe that the Ordinal methods should be used, since it is way faster. But there is a possible behavior change that should be considered here.

Behavior change example:
// http://stackoverflow.com/questions/492799/difference-between-invariantculture-and-ordinal-string-comparison
var s1 = "Strasse";
var s2 = "Straße";
s1.Equals(s2, StringComparison.Ordinal);          //false
s1.Equals(s2, StringComparison.InvariantCulture);  //true

khalidsalomao:
Close this push request to split it in two different push request since this commit broke some unit tests... sorry!

PR #1806, opened 2015-02-06
DynamicDictionary BugFix

khalidsalomao:
BugFix:  ContainsKey and TryGetValue key normalizations.
Both methods must also use "GetNeutralKey" method to prepare the
dictionary key.


khellang:
Makes perfect sense! :+1:

PR #1809, opened 2015-02-10
Compare full paths in generic file response sanity checks

ClemPi:


khellang:
I guess this makes sense. :+1:

Issue #1679, opened 2014-09-08
SSL Offloading not supported


mvandiest:
Nancy.Security.ModuleSecurity:

Currently this check for a secure request is limited to inspecting the requesting url:

`if (!request.Url.IsSecure)`

Using SSL Offloading this check will usually be false as the communiction between the Offloader and the server is done over http internally. 

Typically in offloading scenarios the offloader will add a custom header, e.g. AWS uses X-Forwarded-Proto: http://amzn.to/1ipyAil

I'm under the gun on a project so I don't have time to make the mod and pull request, but it would be nice if this could be modified to look for a configurable custom header to inspect in addition to the Url check.

khellang:
Closed with #1815

khellang:
In your bootstrapper, you can now do `SSLProxy.RewriteSchemeUsingForwardedHeaders(pipelines)` to rewrite the scheme to `https` if there's a `X-Forwarded-Proto` or a `Forwarded` header present.

mvandiest:
Awesome. Thanks!

khellang:
@mvandiest Hopefully we'll be able to push v1.1 sometime tomorrow. If you need the bits before then, pull them [from MyGet](https://www.myget.org/gallery/nancyfx) :smile:

PR #1815, opened 2015-02-15
added Before hook to make Url https if forwarded header present

jchannon:
Added optional before hook to make url scheme https if `forwarded` header present

Usage : `SSLProxy. MakeNancyUrlSecure(pipelines)`

Addresses #1679

### DIFF
--- a/src/Nancy/Security/SSLProxy.cs
+++ b/src/Nancy/Security/SSLProxy.cs
@@ -11,12 +11,12 @@
     public static class SSLProxy
     {
         /// <summary>
-        /// Checks for Forwarded or X-Forwarded-Proto header and if so makes curent url schemme https
+        /// Checks for Forwarded or X-Forwarded-Proto header and if so makes current url scheme https
         /// </summary>
-        /// <param name="pipelines"></param>
-        public static void MakeNancyUrlSecure(IPipelines pipelines)
+        /// <param name="pipelines">Application pipelines</param>
+        public static void RewriteSchemeUsingForwardedHeaders(IPipelines pipelines)
         {
-            pipelines.BeforeRequest += (ctx) =>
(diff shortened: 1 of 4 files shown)


src/Nancy/Security/SSLProxy.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    using System;
+    using System.Linq;
+
+    using Nancy.Bootstrapper;
+
+    /// <summary>
+    /// Allows a BeforeRequest hook to change Url to HTTPS if X-Forwarded-Proto header present
+    /// </summary>
+    public class SSLProxy
```
khellang: `static`?

jchannon: same as CSRF

khellang: You mean `public static class Csrf`? Doesn't mean it shouldn't be static. Why would you create an instance of `SSLProxy`?

jchannon: Well I was just trying to find something similar.  I thought about
InternalConfiguration but that didn't seem the right place

khellang: I'm not saying you should move it, just make the class static :wink:

thecodejunkie: Yeah, making the class static makes sense :)

khellang:
LGTM :shipit:

khellang:
I'm not sure I can come up with a better name, but `MakeNancyUrlSecure` sounds a bit weird considering what the method is doing IMO.

jchannon:
Yeah I had Enable before but that wasn't right either :(

khellang:
:boom: I added the "Update Documentation" label, @jchannon: want to update the wiki?

PR #1820, opened 2015-02-19
Replacing current "user" objects with ClaimsPrincipal/ClaimsIdentity

### LINKED ISSUE #1800: Claims support
# Replacing current "user" objects with ClaimsPrincipal/ClaimsIdentity

## Mandatory:

- [x] Remove IUserIdentity
- [x] Replace UserIdentityExtensions with a version that has the same (as much as possible - will need to be updated for the new, correct, structure of a claim) API but uses ClaimsPrincipal
- [x] Update SecurityHooks to use the new extensions and claim structure
- [x] Update ModuleSecurity to use the new extensions and claim structure
- [x] Update NancyContext to use ClaimsIdentiy/Principal
- [x] Allow the principal to "flow" from hosting
- [x] Update OWIN to automatically flow the identity
- [x] Update tests

## Preferred:

- [x] Update Forms Auth to use the new extensions and claim structure
- [x] Update Stateless Auth to use the new extensions and claim structure
- [x] Remove Token Auth :fire:
- [x] Update Basic Auth to use the new extensions and claim structure

## 4.0 Support (up for debate):

- [ ] Create a new project Nancy.Security containing a "copy" of ClaimsPrincipal/ClaimsIdentity (nabbed from Mono?), possibly renamed to NancyClaimsPrincipal/NancyClaimsIdentity
- [ ] Reference that new project in Nancy, and the auth projects, and use that new type everywhere, rather than the .net 4.5 one (see Mandatory above)
- [ ] For .net 4.5 add a conditional compilation that adds implicit casting between NancyClaimsPrincipal/NancyClaimsIdentity and the .net 4.5 versions
- [ ] Update core build script to build both versions 
- [ ] Include the new assemblies in the core Nancy nuget for the relevant platforms

The working branch for this issue can be found at https://github.com/nancyfx/nancy/tree/v2

damianh:
PR to fix #1800

## Mandatory:

- [x] Remove IUserIdentity
- [x] Replace UserIdentityExtensions with a version that has the same (as much as possible - will need to be updated for the new, correct, structure of a claim) API but uses ClaimsPrincipal
- [x] Update SecurityHooks to use the new extensions and claim structure
- [x] Update ModuleSecurity to use the new extensions and claim structure
- [x] Update NancyContext to use ClaimsIdentiy/Principal
- [x] Allow the principal to "flow" from hosting
- [x] Update OWIN to automatically flow the identity
- [x] Update tests

## Preferred:

- [x] Update Forms Auth to use the new extensions and claim structure
- [x] Update Stateless Auth to use the new extensions and claim structure
- [x] ~~Update Token Auth to use the new extensions and claim structure~~ Remove Token Auth :fire: 
- [x] Update Basic Auth to use the new extensions and claim structure

## Additional
 - [x] Don't use bootstrapper static location in `NancyOptions`. It WILL  cause conflicts and composition problems in an OWIN pipeline.


damianh:
@NancyFx/owners @NancyFx/most-valued-minions  Right lads, bit of a big fish landing here. Going to need your careful attention. I'd like to get feedback before going any further.

At this point everything other that `Nancy.Authentication.Token` and `Nancy.Authentication.Token.Tests` compiles. Tests pass in all of the remaining projects, though I did delete a couple of tests that may be contentions. I'll highlight those.

To review, best start by looking at changes to `ClaimsPrincipalExtensions`, `ModuleSecurity` and `SecurityHook`.

damianh:
What should we do with `Nancy.Authentication.Token` ? There is already a standard for this, OAuth, that uses JWT

jchannon:
https://github.com/jchannon/Owin.StatelessAuth

src/Nancy.Tests/Unit/Security/ModuleSecurityFixture.cs
```diff
@@ -286,84 +207,15 @@ public void Should_return_null_with_RequiresAnyClaim_and_any_claim_met()
         }
 
         [Fact]
-        public void Should_return_forbidden_response_with_RequiresValidatedClaims_enabled_but_claims_missing()
```
damianh: Any tests around `RequiresValidatedClaims` were removed. HasClaim(s) performs that function.

src/Nancy.Tests.Functional/Modules/PerRouteAuthModule.cs
```diff
@@ -19,21 +17,14 @@ public PerRouteAuthModule()
 
             Get["/requiresclaims"] = _ =>
             {
-                this.RequiresClaims(new[] { "test", "test2" });
+                this.RequiresClaims(c => c.Type == "test", c => c.Type == "test2");
```
thecodejunkie: Perhaps we could have an overload of `RequireClaims` that takes a `params string[] claim`. The overload would iterate over the params and compare them against the `Type` of each claim. It would give us a nicer syntax. Thoughts?

thecodejunkie:
I think the changes so far looks alright :smile: :+1:

src/Nancy.Tests.Functional/Modules/PerRouteAuthModule.cs
damianh: It would have to be called 'RequiredClaimTypes'.

Regardless, its not particularly useful as you generally need to inspect
_at least_ 2 attributes of a claim - the Type and the Value. See HasClaim
https://msdn.microsoft.com/en-us/library/hh194432%28v=vs.110%29.aspx and
the predicate overload
https://msdn.microsoft.com/en-us/library/hh159713(v=vs.110).aspx

thecodejunkie:
Why is `Nancy.Authentication.Token` an issue here? I get it that there are other, existing, things for this but could we leave it as is for now so we don't get stuck on that? I know it makes use of `IUserIdentity`, are you seeing issues with making it compatible with the new stuff?

grumpydev:
Might be nice to loop in @leastprivilege and @brockallen - be nice to get some identityserver integration/samples/lovin' alongside this.

leastprivilege:
We could add a Nancy sample API here:

https://github.com/IdentityServer/Thinktecture.IdentityServer3.Samples/tree/master/source/Clients

(analogous to https://github.com/IdentityServer/Thinktecture.IdentityServer3.Samples/tree/master/source/Clients/SampleAspNetWebApi)

leastprivilege:
And with "we" I really mean "you" ;) I am currently busy with working on the PR on my family repo ;)

khellang:
We should loop in @jeffdoolittle as well, he was talking about JWT etc. over at #1550.

damianh:
We can't carry `Nancy.Authentication.Token` forward with just swapping `ClaimsPrincipal` for `IUserIdentity`

It currently does two things:
 1. [Serializes IUserIdentity into a token (hashing it etc)](https://github.com/NancyFx/Nancy/blob/master/src/Nancy.Authentication.Token/Tokenizer.cs#L85-L87).
 2. [Deserializes it in the request pipeline](https://github.com/NancyFx/Nancy/blob/master/src/Nancy.Authentication.Token/Tokenizer.cs#L113).

Serializing the `ClaimsPrincipal` in a similar way is going to _a lot_ heavier in terms of token size. In anycase, it's also questionable if that is even a reasonable idea considering JWTs  (which are not necessarily trivial - https://tools.ietf.org/html/draft-ietf-oauth-json-web-token and https://auth0.com/blog/2014/01/27/ten-things-you-should-know-about-tokens-and-cookies/ ). JWTs, because they're standard, means the authorization header is verifiable in other middleware / frameworks etc. Even if serialization of `ClaimsPrincipal` worked, do we really want a Nancy specific token auth scheme?

I suggest we delete `Nancy.Authentication.Token` and open a separate issue for JWT support (if we want that _in_ nancy).

damianh:
I've also looked into Owin.StatelessAuth  ( @jchannon correct me if I'm wrong ), it appears it just checks for the existence of the header and offloads the token verification to the user and it doesn't assist in generating the tokens.

jchannon:
Correct but if you lol at the example it's easy to generate a token. We
could also have s DefaultTokenValidator?. We've also added a pass through
to continue into the pipeline if a property in the options is true

grumpydev:
Serializing the claims etc. is exactly what most things do, and yeah, they are massive as a result (and often chunked) - do we need anything regarding jwts? I would expect to leverage identity server for that.
________________________________

damianh:
Can you chunk an Authorization header? I know that is done with cookies, never heard of that with a header.

A practice I've seen is to put minimal amount of claims in a Auth header and have a server side claims store to "enrich" the ClaimsPrincipal in your pipeline. Though this isn't a standard thing by any means.

damianh:
http://beingabstract.com/2014/10/serializing-the-claimsidentity/

damianh:
If we hypothetically created a DTO to serialize a ClaimsPrincipal, all identities contained therein, is there any chance of bugs / missing stuff when deserializing?

grumpydev:
Sorry, assumed you were talking cookies.. You could chunk a header, but don't think I have ever seen it done. 

-----Original Message-----

damianh:
@grumpydev 
> do we need anything regarding jwts? I would expect to leverage identity server for that.

Agree. Which is why `Nancy.Authentication.Token` has suspect utility.

jchannon:
Hope not cause that's what I do

damianh:
@jchannon How many identities and claims? What size is the payload?

jchannon:
https://github.com/jchannon/Owin.StatelessAuth/blob/master/src/Owin.StatelessAuthExample/HomeModule.cs#L25

danbarua:
I dont shove the entire ClaimsIdentity in the JWT due to bloat, just a few key value bits with short keys which is used to reconstitute a ClaimsMonster on the server.

leastprivilege:
Have a look at cookie/bearer token authN middleware in Katana. Lot of work all in all. Why do you want that inside Nancy? that looks like middleware to me.

damianh:
Cheers @leastprivilege

khellang:
If we yank out `Nancy.Authentication.Token`, let's not forget to update the NuGet package to cap it before 2.0. I think `(,2)` is the dependency version syntax...

damianh:
@khellang all nuget packages should be capped to the next major version at the very least.

khellang:
@damianh Not the ones we plan to carry forward?

damianh:
@khellang I'm talking about in general practice.

damianh:
image

@grumpydev any ideas? Thought mono 3.12.1.0 supported ClaimsPrincipal...

thecodejunkie:
Yeah we need to put a cap on them because we can't have v1 nugets trying to install v2 versions of their dependencies and the other way around

thecodejunkie:
@damianh it should. Miguel added it months ago and @jchannon even sent a pull-request on it

thecodejunkie:
Here it is https://github.com/mono/mono/blob/master/mcs/class/corlib/System.Security.Claims/ClaimsPrincipal.cs

jchannon:
Yup its supported in mono

damianh:
Er so why is it failing on Travis? :S

jchannon:
It may depend on how you are using them, I had to change the way I
initialized them for it to work on Mono, can't remember if my PR to mono
solved that or not

grumpydev:
No idea why it's failing.. it should be using the latest mono version :/

khellang:
:boom: :grin::+1:

grumpydev:
Yay :)

damianh:
The TC failing test.... (but didn't fail on Travis lol)

image

is because of https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Owin/NancyOptions.cs#L19 .

Static service location? We _really_ don't want this from an OWIN perspective. Going to make a change to require a bootstrapper to be supplied to `NancyOptions`.

phillip-haydon:
Omg stop failing!

damianh:
Ready to be reviewed, with particular attention to that last commit pls. 

Obviously needs rebasing, but will do that seperatedly after you've had a chance to review.

khellang:
This will break A LOT of `app.UseNancy()` cases :disappointed:

damianh:
Yes it will. Allowing this is pit-of-fail. User does app.UserNancy() twice in a pipeline and stuff breaks. Remember, middleware can contain middleware, and it may come from other places, so it may not be obvious to them in their wireups.

Nancy must no longer assume it's the _only_ framework in an application.

Fwiw, there is no such thing as `app.UseWebApi()`. It explicitly requires a configuration parameter too.

grumpydev:
I disagree, we still need our happy path - the 99% use case is Nancy on its own, or potentially nancy with some auth middleware - anyone using complex middleware setups should already be aware of how to set it up.

damianh:
I knew you would :)

grumpydev:
And WebApi as an example on how to do things? You'll be suggesting sharepoint next :)

damianh:
One data point an projection does not make :)

damianh:
Will happy-path users still be coming via OWIN(\whatever) pipelines though? hmm

grumpydev:
We could take a similar approach to the one I suggested for the razor config wrecking problem - have two "levels" of nuget, our normal ones with happy paths, then a .core that is just bare bones. 

In razor's case .core would be everything but the config transforms, razor main package would have a dependency on that, and add in the transforms. 

For the owin stuff perhaps it could be that the happy path stuff isn't in core, but is included as a dependency in Nancy.hosting.aspnet (just thinking out loud to give an idea)

damianh:
Could Nancy.Hosting.AspNet could do the bootstrapper static location?

grumpydev:
That's what it does currently, but not sure how it will work if we shift it to owin.. Maybe the "hosting" packages that use katana will bring in a startup.cs

grumpydev:
Regardless of the outcome of the wireup discussion, I don't think it has any relevance to the scope of this issue/PR, so if you could roll it back that'd be awesome.

damianh:
Yah fair enough. What about that failing test because it's finding multiple
bootstrappers in the test assembly? Test for the exception?

grumpydev:
Sorry @damianh I missed your question.. yeah, those tests are a bit broken I think.. something like this makes more sense:

```csharp
namespace Nancy.Owin.Tests
{
    using Nancy.Bootstrapper;
    using Nancy.Tests;

    using Xunit;

    public class NancyOptionsFixture
    {
        private readonly NancyOptions nancyOptions;

        public NancyOptionsFixture()
        {
            this.nancyOptions = new NancyOptions();
        }

        [Fact]
        public void Bootstrapper_should_use_locator_if_not_specified()
        {
            // Given
            var bootstrapper = new DefaultNancyBootstrapper();
            NancyBootstrapperLocator.Bootstrapper = bootstrapper;

            //When
            //Then
            this.nancyOptions.Bootstrapper.ShouldNotBeNull();
            this.nancyOptions.Bootstrapper.ShouldBeSameAs(bootstrapper);
        }

        [Fact]
        public void Bootstrapper_should_use_chosen_bootstrapper_if_specified()
        {
            // Given
            var bootstrapper = new DefaultNancyBootstrapper();
            var specificBootstrapper = new DefaultNancyBootstrapper();
            NancyBootstrapperLocator.Bootstrapper = bootstrapper;

            //When
            this.nancyOptions.Bootstrapper = specificBootstrapper;

            //Then
            this.nancyOptions.Bootstrapper.ShouldNotBeNull();
            this.nancyOptions.Bootstrapper.ShouldBeSameAs(specificBootstrapper);
        }

        [Fact]
        public void PerformPassThrough_should_not_be_null()
        {
            this.nancyOptions.PerformPassThrough.ShouldNotBeNull();
        }

        [Fact]
        public void PerformPassThrough_delegate_should_return_false()
        {
            this.nancyOptions.PerformPassThrough(new NancyContext()).ShouldBeFalse();
        }
    }
}
```

damianh:
@grumpydev only problem with these tests is they can't be run in parallel (static mutable state being the enemy of parallel testing). May or not be a concern for @NancyFx/owners .

damianh:
Test fixed. I guess a rebase is required. @NancyFx/owners Could one of you rebase the 2.0 branch (if deemed necessary)?

grumpydev:
I thought about that @damianh, but I couldn't think of anywhere else that used that property - I suppose it could be refactored so NancyOptions has:

public Func<INancyBootstrapper> BootstrapperLocator { get; set; }

ctor() {
    this.BootstrapperLocator = () => NancyBootstrapperLocator.Bootstrapper;
}

Then that can be swapped out to whatever you like.. quite possibly overkill though :)

damianh:
Yeah, think it's a bit of overkill right now. Lets move on and address it separately as you've suggested.

xt0rted:
Is there a CI feed for the v2 branch so I can try this out?

grumpydev:
@xt0rted : not yet, we will put a beta out after 1.2 in the next week.

khellang:
@xt0rted You can grab the bits [from here](http://nancy-ci.cloudapp.net/viewLog.html?buildId=2180&tab=artifacts&buildTypeId=Nancy_v2) :smile:

PR #1854, opened 2015-03-05
Another round of typo clean up & Converted tabs to spaces.

06b:
Converted tabs to spaces & Fixed some more typos.

I'm sure I might of missed a few but oh well.


khellang:
:heart: it! Thanks :smile:

albertjan:
@06b are you just pasting the code in word and running a spell checker or are you actually good at spelling :smille:

06b:
@albertjan HA! I actually think I have a VM which for some sick reason has word as it's default html editor... I need to update that. I wish I was this good at spelling, some of it typos I caught at first was when I was looking in the code and trying to get a better understanding of how Nancy works (Still over my head - you guys are crazy smart :+1: ) but recently like within the last two weeks I found some plugin which has a spell checker either on my work or home machine.

So that's why part of my commits were just like one offs and others were a whole batch of updates. It depended on which machine I was working on at the time, also some of them aren't typos just the wrong word spelled right so some of it was me reading.

If you want I can share the plugin, funny enough is I think it has helped me a bit with my own spelling because I've started catching like statements in my own code comments or even like messages I type in websites which wasn't provided by a copywriter where it's either broken english or just typos.

There are a few parts in Nancy where I noticed it would say like ontop instead of on top, that didn't bother me as much but I mean I can always find them and fix that as well.

I figured this would be the best I could do to help out, haha didn't mean to bomb you guys with a ton of pull requests all at once.

PR #1857, opened 2015-03-06
Add Head route hook with static config to enable old behavior

### LINKED ISSUE #1771: Explicit support for HEAD requests
Even though #2 'reads' like it was added to Nancy a while back, there's, at least to my understanding, no explicit support for HEAD requests and what I am doing right now is checking/comparing the NancyContext for the request-type inside my Get["/"] etc handlers... which seems cumbersome.

While GET and HEAD are very closely related, I do have different implementations for the two: e.g. I only want to return the cache / ETAG value in a HEAD request and nothing else.. if I kept my GET implementation dumb it would always hit the underlying data "loading" mechanisms, be it cache or db.

So if I could define explicit Head ["/"] handlers it would remove a hell lot of if-conditions inside my Get handlers.

Is there a particular reason why this isn't available or is it.. and I am just not seeing it?

Cheers and thanks,
-J

adamhathcock:
Hopefully the code fits guidelines.  It seems to cover and fix #1771 

I couldn't find a reason why the head verb isn't handled explicitly.

Can invert the static config.

### DIFF
--- a/src/Nancy.Tests.Functional/Tests/BasicRouteInvocationsFixture.cs
+++ b/src/Nancy.Tests.Functional/Tests/BasicRouteInvocationsFixture.cs
@@ -146,6 +146,11 @@
                     return "Default get root";
                 };
 
+                Head["/"] = parameters =>
+                {
+                    return "Default head root";
+                };
+
                 Post["/"] = parameters =>
                 {
                     return "Default post root";
--- a/src/Nancy.Tests.Functional/Tests/BasicRouteInvocationsFixture.cs
+++ b/src/Nancy.Tests.Functional/Tests/BasicRouteInvocationsFixture.cs
@@ -146,11 +146,6 @@
                     return "Default get root";
                 };
 
-                Head["/"] = parameters =>
-                {
-                    return "Default head root";
-                };
-
                 Post["/"] = parameters =>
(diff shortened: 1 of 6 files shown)


khellang:
This works? What about [this](https://github.com/NancyFx/Nancy/blob/974fc83e3817c99e26cad677283b555ed8432cdd/src/Nancy/Routing/DefaultRequestDispatcher.cs#L67-L70)? IMO we should keep the default behavior and you can set the new property if you want to opt-out.

adamhathcock:
I believe that code is ensuring a head request has no body in the response.  I don't think that needs to be changed.

I'll invert the static config bool on Monday.

adamhathcock:
Inverted static bool.  Fixed up tests and made HeadResponse propagate the ReasonPhrase

khellang:
Thanks @adamhathcock! :dancers:

khellang:
Added a line about this in the wiki:

> By default, HEAD requests are automatically handled for all routes that are declared for GET requests. If you want to manually handle HEAD requests, you can set StaticConfiguration.EnableHeadRouting to true and define the routes in your modules.

Looks OK?

adamhathcock:
:+1:

Issue #1629, opened 2014-07-23
Seperate libraries not loading defined nancy dependencies


jchannon:
We all know of the issue where you have a `MyApp.Tests` lib that uses `Nancy.Testing` to test your app and the requirement to install `Nancy.Validation.FluentValidation` in your test lib for the test to pass validation. I think this is so that the dll is in the appdomain.

I've just lost 3 hours on the same issue but this time with `Nancy.Serialization.ServiceStack`

In my test lib I'm using HttpClient and the OWIN pipeline to post the below model as JSON. In my module it calls BindAndValidate on this model:

```
    public class ApiLoggingModel
    {
        public LoggingEnum Level { get; set; }
        public string Message { get; set; }
    }
```

All is well however its using ` Nancy.ModelBinding.DefaultBodyDeserializers.JsonBodyDeserializer` not the `Nancy.Serialization.ServiceStack` one I have registered in my app.  The only reason I noticed was because ServiceStack was throwing an exception I assume on enums when the app was running. My tests were saying all was fine but in reality it wasn't the case.  If I added `Nancy.Serialization.ServiceStack` to my tests lib it would show the same behaviour.

This leads me onto the same scenario with OWIN hosting. I have a IIS Host project and a Self host project. Both reference a class lib with all my modules, validation etc in it.  I have to install the `Nancy.Validation.FluentValidation` package in my host project for the validation stuff to work, the same goes for the serialization stuff.

I'm not sure how we get around this issue but its a huge PITA and I'm not sure if its a nancy issue we need to solve or a .net thing

thecodejunkie:
@jchannon isn't this just the classic ".net trying to be smart with assembly references problem?

PR #1860, opened 2015-03-09
Added IncludeInNancyAssemblyScanningAttribute

### LINKED ISSUE #1842: Using an assembly reference to improve accuracy of assembly loading
During assembly scanning, we try to work a bit of magic by scanning for unloaded assemblies that reference as Nancy assembly. However, as we know, .NET is sometimes trying to be smart and clear referenced assemblies, where a type is not directly referenced in code. This results in Nancy not loading the assembly thus some types will not be loaded and be discovered by Nancy at runtime. There is little we can do about this behavior (unless, explicitly force people to add a type reference somewhere in their code).

One possible workaround could be to look for an attribute (guessing that they'll not be cleaned up)

```c#
[AttributeUsage(AttributeTargets.Assembly)]
public class NancyReferenceAttribute : Attribute
{
}
```

It could then be slapped on an assembly

`[assembly: NancyReference]`

We could then update ADATS to look for the attribute first and if it's not present, then we will try and look for a Nancy reference. Come to think of it, the attribute usage should be considered a directer type usage so it's mimicking referencing a Nancy type, in a less convoluted approach. We shouldn't need to update ADATS, the attribute would just provide (in my opinion) a cleaner approach to the solution we tell people to use.

Thoughts?

khellang:
This adds the `IncludeInNancyAssemblyScanningAttribute`. Let's do a final round on the name. Does anyone care much? :smile:

I also tried to make ADATS look for the attribute before checking an assembly's references, but failed to get it to work in reflection-only context :disappointed: 

Closes #1842

### DIFF
--- a/src/Nancy/IncludeInNancyAssemblyScanningAttribute.cs
+++ b/src/Nancy/IncludeInNancyAssemblyScanningAttribute.cs
@@ -1,0 +1,17 @@
+﻿namespace Nancy
+{
+    using System;
+
+    /// <summary>
+    /// Add this attribute to an assembly to make sure
+    /// it is included in Nancy's assembly scanning.
+    /// </summary>
+    /// <example>
+    /// Apply the attribute, typically in AssemblyInfo.(cs|fs|vb), as follows:
+    /// <code>[assembly: IncludeInNancyAssemblyScanning]</code>
+    /// </example>
+    [AttributeUsage(AttributeTargets.Assembly)]
+    public sealed class IncludeInNancyAssemblyScanningAttribute : Attribute
+    {
+    }
+}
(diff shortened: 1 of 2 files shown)


thecodejunkie:
I think the name is alright. What say the @NancyFx/most-valued-minions @NancyFx/nancy-core-contributors ?

grumpydev:
:shipit:

jchannon:
Does it fix #1629?

horsdal:
Only if the other libraries add the new attribute.

And :+1: for the name (no surpise :smile: )

thecodejunkie:
Yeah it **should** fix it. The theory is that if you slap this on your assembly, then .net can't try and be smart and remove the Nancy reference if you don't directly use a type.

jchannon:
It's not always Nancy needed it's things like the validation lib

khellang:
Updated the wiki with a note on this attribute: https://github.com/NancyFx/Nancy/wiki/Bootstrapper#finding-the-right-bootstrapper

PR #1862, opened 2015-03-11
Do not register dependency twice by the same interface in cofigurable bootstrapper

### LINKED ISSUE #1802: Nancy.Testing With.Dependency<T>(T instance) is disposing of the instance provided
Unit testing code with `Nancy.Testing`.
When I provide an instance of an object using `With.Testing<T>(T instance)` I believe the object is disposed of.

I [made this full repo](https://gist.github.com/PureKrome/ba3a519d807e5103243c) to proove this.

Sample code which hightlights the problem :+1: 

```C#
var browser = new Nancy.Testing.Browser(with =>
            {
                with.Module<SampleModule>();
                with.Dependency<IDocumentStore>(documentStore); // THROWS EXCEPTION
                //with.Dependency(documentStore); // WORKS
            });
```

horsdal:
If two instance dependencies are registered in TinyIoC by the same interface the first one is disposed if it implements `IDisposable`. Therefore registering an instance that implements `IDisposable` twice by the same interface means it will be disposed. The configurable bootstrapper did that.

Fixes #1802.

### DIFF
--- a/src/Nancy.Testing.Tests/ConfigurableBootstrapperDependenciesTests.cs
+++ b/src/Nancy.Testing.Tests/ConfigurableBootstrapperDependenciesTests.cs
@@ -91,6 +91,32 @@
             Assert.Contains("Implemented ITestDependency", response.Body.AsString());
         }
 
+        interface IIinterface { }
+        class DisposableDependency : IIinterface, IDisposable
+        {
+            public bool Disposed { get; private set; }
+      
+            public void Dispose()
+            {
+                this.Disposed = true;
+            }
+        }
+
+        [Fact]
+        public void should_be_able_to_configure_disposbale_dependency_implementing_interface_by_interface()
+        {
+            // Given
+            var disposableDependency = new DisposableDependency();
+
+            // When
+            new Browser(with =>
+              with.Dependency<IIinterface>(disposableDependency)
+            );
+
+            // Then
+            Assert.False(disposableDependency.Disposed);
+        }
+
 
         [Fact]
         public void should_be_able_to_configure_dependencies_by_instances()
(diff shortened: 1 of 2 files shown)


src/Nancy.Testing.Tests/ConfigurableBootstrapperDependenciesTests.cs
```diff
@@ -91,6 +91,32 @@ public void should_be_able_to_configure_dependency_type()
             Assert.Contains("Implemented ITestDependency", response.Body.AsString());
         }
 
+        interface IIinterface { }
+        class DisposableDependency : IIinterface, IDisposable
+        {
+            public bool disposed = false;
```
khellang: Nit: Make this a property with proper naming?

horsdal: think that's overkill here, tbh.

khellang: Hey, correct naming is never overkill :danc

horsdal: But a property is. Not the name of course.

src/Nancy.Testing.Tests/ConfigurableBootstrapperDependenciesTests.cs
```diff
@@ -91,6 +91,32 @@ public void should_be_able_to_configure_dependency_type()
             Assert.Contains("Implemented ITestDependency", response.Body.AsString());
         }
 
+        interface IIinterface { }
+        class DisposableDependency : IIinterface, IDisposable
+        {
+            public bool Disposed { get; private set; }
```
horsdal: There you go, @khellang :cocktail:

Issue #1866, opened 2015-03-13
the questionmark is lost when logging on and redirecting


Tsingbo-Kooboo:
Hi Nancy,

I found a small but import bug.We can see some code in https://github.com/NancyFx/Nancy/blob/master/src/Nancy.Authentication.Forms/FormsAuthentication.cs from line 226.

        private static Action<NancyContext> GetRedirectToLoginHook(FormsAuthenticationConfiguration configuration)
        {
            return context =>
                {
                    if (context.Response.StatusCode == HttpStatusCode.Unauthorized)
                    {
                        string redirectQuerystringKey = GetRedirectQuerystringKey(configuration);

                        context.Response = context.GetRedirect(
                            string.Format("{0}?{1}={2}", 
                            configuration.RedirectUrl,
                            redirectQuerystringKey,
                            context.ToFullPath("~" + context.Request.Path + HttpUtility.UrlEncode(context.Request.Url.Query))));
                    }
                };
        }

So when I try to access for example 

http://www.sample.com/Admin/Sites/Site/?SiteId=c2e464be-5a44-1f49-4633-851f2575161d 

without logon,it will be redirected to 

http://www.sample.com/Admin/Account/Login?returnUrl=/Admin/Sites/Site/SiteId%3dc2e464be-5a44-1f49-4633-851f2575161d 

and the questionmark "?" is lost,so it can not work correctly. Now I change the code like the following:

        private static Action<NancyContext> GetRedirectToLoginHook(FormsAuthenticationConfiguration configuration)
        {
            return context =>
            {
                if (context.Response.StatusCode == HttpStatusCode.Unauthorized)
                {
                    string redirectQuerystringKey = GetRedirectQuerystringKey(configuration);
                    string query = context.Request.Url.Query;
                    string queryString = String.IsNullOrEmpty(query) ? String.Empty : HttpUtility.UrlEncode("?" + query);
                    context.Response = context.GetRedirect(
                        string.Format("{0}?{1}={2}",
                        configuration.RedirectUrl,
                        redirectQuerystringKey,
                        context.ToFullPath("~" + context.Request.Path + queryString)));
                }
            };
        }

It works fine for me now,but maybe it is not the best way to fix this bug,could you please check it?

Regards,

Tsingbo

PR #1867, opened 2015-03-13
bug fixing:the questionmark is lost when logging on and redirecting

Tsingbo-Kooboo:
Hi Nancy,
This is the fix like what https://github.com/NancyFx/Nancy/issues/1866 said,discard this if you have better way to fix.Thx!


khellang:
@Tsingbo-Kooboo There's 4 tests failing now. Want to look at them? http://nancy-ci.cloudapp.net/viewLog.html?buildId=2119&buildTypeId=Nancy_Pulls

khellang:
Several others have mentioned the same on JabbR etc. without filing a bug for it. We'd love to get this bug fixed once and for all, but the tests seem to suggest that the questionmark should be present in the query string. Don't know what's going on.

khellang:
Ah, I see what's going on. The tests passes a query string containing `?` into the request, while the hosts doesn't. So the tests ends up with double `%3f`. What you could do is just remove the `?` from the test query strings.

khellang:
OK, looks like `NancyHandler` (the `System.Web` host) [passes the raw `Uri.Query` to the request](https://github.com/NancyFx/Nancy/blob/master/src/Nancy.Hosting.Aspnet/NancyHandler.cs#L87), which also should contain the `?`. Maybe it's just a matter of normalizing how the query is stored in `Url`? E.g. make the Query property of url backed by a field and doing

```csharp
public string Query
{
    get { return this.query; }
    set { this.query = value.TrimStart('?'); }
}
```

@NancyFx/owners @NancyFx/most-valued-minions What do you think?

phillip-haydon:
I concur.

horsdal:
:+1: for normalizing

khellang:
@Tsingbo-Kooboo Thanks for providing a potential fix. It turned out to be a deeper problem (how the hosts handles the leading `?` in query strings). I sent a PR (#1870) which should solve this problem once and for all. Would you mind trying it out from the [CI feed](https://www.myget.org/gallery/nancyfx) and see if it solves your problem?

PR #1873, opened 2015-03-20
Get provided CSRF-tokens either from formdata or request header

loftum:
It would be nice to post CSRF-tokens either as form data or as a request header.

On validation Csrf.cs could look in both form data and request header for the provided token, before validating against the cookie token.

This would really make our day, especially when posting json-data from JavaScript, as we would not have to deal with the CSRF-token in the posted data at all.


albertjan:
Interesting! How do you send the token to the client in the html? You send it back by adding a header to an xhr request?

Could you add a test?

loftum:
# Last thing first: Yes, I could add a test. But since I didn't find any
other unit tests on Csrf.ValidateCsrfToken(), I assumed you didn't find
that necessary.
Would you like me to put the test in the CsrfFixture class?


# First thing last:
Yes, we add the header to an xhr with JavaScript.
We are currently using angular, and we use a directive to put things in the
header, more or less like this:

<form method=....>
  <input type="hidden" name="NCSRF" value="CSRF token value"
add-this-to-header />
</form>

directive('addThisToHeader', ['$http', function ($http) {
        return function (scope, element, attrs) {
            $http.defaults.headers.common[attrs.name] = attrs.value;
        };

Running example here:
https://jsfiddle.net/jqhw31q0/11/
(Press the submit button to see the header value)


We strive to make the site work even with JavaScript disabled.
By using a hidden input and copy the value to a header with JS, the token
will be submitted (as form data) when JS is disabled.

If Nancy supports reading a header value, we can skip transforming
json-data (including CSRF-token) into form data before posting.


-hans olav

loftum:
2 tests added.
I also moved csrf token generating to an internal static Csrf.GenerateTokenString(), That way we can generate tokens from the tests more easily than running the pipelines and fetching token from Context.Items.

khellang:
Hmm. I think something went wrong with your rebase or something. The commits are screwed up :confused:

loftum:
I think you are right. I screwed up.
I created a new feature branch and a new pull request #1878.

horsdal:
Closing since this is replaced by #1878

PR #1874, opened 2015-03-21
Super Simple Each Perf

darthwalsh:
Each substitutions in the SuperSimpleViewEngine slowed down with N^2 performance with large number of elements (e.g. 1000).
Switching to StringBuilder gives linear performance.


khellang:
Thanks! :heart:

Issue #1831, opened 2015-02-23
IDiagnosticProvider only picks up most derived type members


OZCMRDM:
InteractiveDiagnostics.GetDiagnosticMethods doesn't return methods on base types. Probably needs to use BindingFlags.FlattenHierarchy.

PR #1875, opened 2015-03-22
Changed DeclaredOnly to FlattenHierarchy

ranjitCatchinUp:
Fix for #1831.

### DIFF
--- a/src/Nancy/Diagnostics/InteractiveDiagnostics.cs
+++ b/src/Nancy/Diagnostics/InteractiveDiagnostics.cs
@@ -9,7 +9,7 @@
     {
         private readonly IDiagnosticsProvider[] providers;
 
-        private const BindingFlags Flags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly;
+        private const BindingFlags Flags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.FlattenHierarchy;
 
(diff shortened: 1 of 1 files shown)


khellang:
Would you be able to provide a test for this as well?

ranjitCatchinUp:
This fix required changing a private method. Since private methods are not usually unit tested I am not sure if I should do that. But if you still want it I would be happy to do it.

khellang:
IMO it doesn't really matter what changed, some behavior did, otherwise this change is useless. I'd like a *behavioral* test that shoed that the correct methods are picked up, not an assert that the correct `BindingFlags` are used :smile:

khellang:
Oops. Looks like 74238a5e6b2f4a72a7b0eb8c46000f8ae335e165 was part of the #1876 PR as well, so GitHub thinks this was merged as part of that PR. Feel free to send this PR again.

ranjitCatchinUp:
My bad. Will raise a new PR. A side question - How are you running xUnit tests locally? i had to install the xUnit test runner in order for the tests to be discoverable by VS Test Runner. But that changed all the packages.config files. If I commit/push this, number of changed files is huge. The other option is to ignore the config files but I am not sure if that is good. How are people handling this?

albertjan:
@ranjitCatchinUp I use [testdriven.net](http://testdriven.net/)

khellang:
I think most of us use the R# runner. I actually think we should install the xUnit runner on the test projects, to make it easier for people to run the tests with vanilla Visual Studio. Are you up for the task? I'll create an issue.

ranjitCatchinUp:
My pleasure. Saw #1879. Expect the PR soon.

Issue #1070, opened 2013-04-18
Allow Additions to MIME Type Extension Mappings


mike-ward:
Two requests:

* Add ".gadget", "application/x-windows-gadget" to the Mime Types extension mappings

* Make the MimeTypes Extension Mappings public or provide methods to add to it.

khellang:
Closed with #1876

PR #1876, opened 2015-03-22
Issue1070

ranjitCatchinUp:
Fix for #1070.


albertjan:
Could you possibly add a test that demonstrates what this fixes? So it doesn't get broken in future version.

ranjitCatchinUp:
@albertjan  added a test like you said. Didn't add test for the gadget mime type since that seems trivial.

PR #1877, opened 2015-03-24
Added constraint support for CaptureNodeWithMultiParameters TrieNode

### LINKED ISSUE #1863: Having multiple route parameters with constraints is not supported
Currently when you have modules which multiple parameters with constraints the constraints get never invoked. The following test including the module reproduces it.

	public class RouteConstraintsTests
    {
        private readonly INancyBootstrapper bootstrapper;

        private readonly Browser browser;

		public static bool Invoked { get; set; }

        public RouteConstraintsTests()
        {
			Invoked = false;

            this.bootstrapper = new ConfigurableBootstrapper(
                configuration =>
                {
                    configuration.ApplicationStartup((c,p) => { });
                    configuration.Modules(new Type[] { typeof(RouteConstraintsModule)});
                    configuration.RouteSegmentConstraints(new[] {typeof (VersionRouteConstraint)});
                });

            this.browser = new Browser(this.bootstrapper);
        }

        [Fact]
        public void Should_be_supported()
        {
            var response = browser.Get(
                @"/4.2...5.3",
                with =>
                {
                    with.HttpRequest();
                });

            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
			Assert.True(Invoked);
        }

        public class VersionRouteConstraint : RouteSegmentConstraintBase<Version> {
            public override string Name
            {
                get { return "version"; }
            }

            protected override bool TryMatch(string constraint, string segment, out Version matchedValue)
            {
				Invoked = true;
                return Version.TryParse(segment, out matchedValue);
            }
        }
    }

 	public class RouteConstraintsModule : NancyModule
    {
        public RouteConstraintsModule()
        {
            this.Get["/{left:version}...{right:version}"] = _ =>
            {
                return 200;
            };
        }
    }

The problem is that the [TrieNodeFactory](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Routing/Trie/TrieNodeFactory.cs#L55) matches with `CaptureNodeWithMultipleParameters` which then itself does not serve as a parent node for possible multiple `CaptureNodeWithConstraint`. So essentially the `CaptureNodeWithMultipleParameters` should create `TrieNode` with the `TrieNodeFactory` for the matched parameter segments.

Sorry was too tired today to wrap my head around it more in the short time window I had. Hopefully this is enough for you guys to fix it.

hlesesne:
Closes #1863.

This is my first PR for Nancy, I have a feeling that my test may be a little lacking. I used the class directly from the issue with very little reformatting. Please let me know if I should adjust.

### DIFF
--- a/src/Nancy.Tests.Functional/Tests/RouteConstraintTests.cs
+++ b/src/Nancy.Tests.Functional/Tests/RouteConstraintTests.cs
@@ -25,7 +25,7 @@
                 {
                     configuration.ApplicationStartup((c, p) => { });
                     configuration.Modules(new Type[] { typeof(RouteConstraintsModule) });
-                    configuration.RouteSegmentConstraints(new[] { typeof(VersionRouteConstraint) });
+                    configuration.RouteSegmentConstraints(new[] { typeof(UltimateRouteSegmentConstraint), typeof(VersionRouteSegmentConstraint) });
                 });
 
             this.browser = new Browser(this.bootstrapper);
@@ -35,7 +35,7 @@
         public void multiple_parameters_per_segment_should_support_constraints()
         {
             // Given
-            const string url = @"/4.3...5.3";
+            const string url = @"/42...42";
 
             // When
             var response = this.browser.Get(
@@ -50,18 +50,42 @@
             Assert.True(Invoked);
         }
 
+        [Fact]
+        public void versionsegmentrouteconstraint_should_match_on_version_numbers()
+        {
+            // Given
+            const string url = @"/4.1.2...4.1.5";
 
-        public class VersionRouteConstraint : RouteSegmentConstraintBase<Version>
+            // When
+            var response = this.browser.Get(
+                url,
+                with =>
+                {
+                    with.HttpRequest();
+                    with.Accept("application/json");
+                });
+
+            // Then
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal(response.Body.AsString(), "{\"left\":\"4.1.2\",\"right\":\"4.1.5\"}");
+        }
+
+
+        public class UltimateRouteSegmentConstraint : RouteSegmentConstraintBase<int>
         {
             public override string Name
             {
-                get { return "version"; }
+                get { return "correctanswer"; }
             }
 
-            protected override bool TryMatch(string constraint, string segment, out Version matchedValue)
+            protected override bool TryMatch(string constraint, string segment, out int matchedValue)
             {
                 Invoked = true;
-                return Version.TryParse(segment, out matchedValue);
+                if (int.TryParse(segment, out matchedValue))
+                {
+                    return matchedValue == 42;
+                }
+                return false;
             }
         }
     }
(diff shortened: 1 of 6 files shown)


hlesesne:
Thanks @albertjan and @khellang - I think I've cleaned these up as suggested. Please let me know if there are any other modifications needed.

khellang:
LGTM :+1: I like it! I think this should go in 1.2 :smile: We need to remember to update the wiki to add the new `version` constraint.

jchannon:
:+1: 

@hlesesne is this for versioning an api?

hlesesne:
@jchannon That would be a good use case. I also think update checks or support links would be other good use cases. I can't take credit for it though it originated in @danielmarbach's issue #1863.

jchannon:
Ah ok, wonder what his user case was

Great work though!

khellang:
@hlesesne Could you add a couple of negative tests as well? E.g. the route doesn't get hit when the URL doesn't match the constraints. After that, I think this is good to :ship:

src/Nancy.Tests.Functional/Tests/RouteConstraintTests.cs
```diff
@@ -0,0 +1,111 @@
...
+                {
+                    with.HttpRequest();
+                    with.Accept("application/json");
+                });
+
+            // Then
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal(response.Body.AsString(), "{\"left\":\"4.1.2\",\"right\":\"4.1.5\"}");
+        }
+
```
khellang: While you're at it. :fire: this whitespace. Sorry for the nitpicking :wink:

src/Nancy.Tests.Functional/Tests/RouteConstraintTests.cs
```diff
@@ -0,0 +1,111 @@
...
+            {
+                return HttpStatusCode.OK;
+            };
+
+            // For testing VersionSegmentRouteConstraint
+            this.Get["/{left:version}...{right:version}"] = _ =>
+            {
+                return new {_.left, _.right};
+            };
+
```
khellang: This as well :smile:

danielmarbach:
@jchannon We are building a cool tool which uses nancy behind the scenes... I'll update you soon when it is public (well it is already being built in public but not yet announced *hehehe*)

jchannon:
*searches Github...*

khellang:
https://github.com/ParticularLabs/APIComparer ? :wink

danielmarbach:
@khellang :dancers:

khellang:
@danielmarbach I think that comment just triggered a tweet :confused: https://twitter.com/justgitthings/status/580413625820676096

hlesesne:
Ewww... What'd I do. I'll check my code and tests.

hlesesne:
I'm comparing a json result for Version and it is returning out of my expected order. I'll simplify the and push.

danielmarbach:
How so?

khellang:
@danielmarbach No idea :stuck_out_tongue_closed_eyes: 

@hlesesne Yeah, that happens sometimes. The Travis build is pretty unstable.

hlesesne:
Fixed.

khellang:
@hlesesne Thank you so much for this! Great work :+1:

khellang:
Updated the wiki: https://github.com/NancyFx/Nancy/wiki/Defining-routes#route-segment-constraints

jchannon:
:shipit:

danielmarbach:
I bow as well! Where can I send the Swiss beer too? ;)

danielmarbach:
@khellang yeah no need to say thank you to me finding the bug :p

khellang:
Many thanks to you as well, @danielmarbach :wink:

hlesesne:
No problem at all. Glad I could contribute.

PR #1878, opened 2015-03-24
Get CSRF-tokens either from request header or form data (2)

loftum:
New attempt for pull request #1873.
I screwed up somehow, so I created a new feature branch and a new pull request.

It would be nice to post CSRF-tokens either as form data or as a request header.

On validation Csrf.cs could look in both form data and request header for the provided token, before validating against the cookie token.

This would really make our day, especially when posting json-data from JavaScript, as we would not have to deal with the CSRF-token in the posted data at all.


loftum:
Sure. Done.

khellang:
:+1:

PR #1882, opened 2015-03-25
Added xUnit Runner reference

ranjitCatchinUp:
Fix for #1879.

PR #1886, opened 2015-03-27
Changed DeclaredOnly to FlattenHierarchy

ranjitCatchinUp:
Fix for #1831 with test. I am not sure why my csproj files have been modified with the ItemGroup tag. I only changed two files InteractiveDiagnosticsFixture.cs and InteractiveDiagnostics.cs.

### DIFF
--- a/src/Nancy.Authentication.Basic.Tests/Nancy.Authentication.Basic.Tests.csproj
+++ b/src/Nancy.Authentication.Basic.Tests/Nancy.Authentication.Basic.Tests.csproj
@@ -151,6 +151,9 @@
       <Install>true</Install>
     </BootstrapperPackage>
   </ItemGroup>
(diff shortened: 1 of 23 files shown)


xt0rted:
VS2013 adds this to class library projects which have a 3rd party test adapter. It's so VS knows to load the test explorer when the solution is loaded and is due to the change in #1882 
```xml
<ItemGroup> 
    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" /> 
</ItemGroup> 
```

khellang:
Yeah. I was wondering why VS didn't add this as part of #1882. Anyway, it's fiiine. I'll look over this PR when I get to work. Should be able to get it in quick and ship as part of 1.2 :smile:

PR #1895, opened 2015-04-14
add missing null parameter checks

scottcarr:
Hello, I'm working on a static analysis tool that looks for missing null checks.  It suggested you add these.  The fields are readonly so assigning null to them doesn't seem useful.

Also, I don't see anywhere that NancyEngine.diagnosticsConfiguration is used (at least according to "Find All References").


khellang:
Thanks! :heart:

Issue #1141, opened 2013-07-05
NancyFX throws Argument exception when handling encoded ">" on Get request


ezechibritton:
NancyFX throws the following exception when handling a route with an encoded ">"

```
System.ArgumentException: Illegal characters in path.
   at System.IO.Path.GetFileName(String path)
at Nancy.Conventions.StaticContentConventionBuilder.<>c__DisplayClass2.<AddDirectory>b__0(NancyContext ctx, String ro
ot)
   at Nancy.DefaultStaticContentProvider.GetContent(NancyContext context)
   at Nancy.NancyEngine.HandleRequest(Request request, Func`2 preRequest)
   at Nancy.NancyEngine.HandleRequest(Request request)
   at Nancy.Hosting.Self.NancyHost.Process(HttpListenerContext ctx)
```

I have created a unit test which can be found at https://gist.github.com/ezechibritton/5933111 that replicates this issue, please note that the unit test produces a slightly different stack trace.

```
System.ArgumentException : Illegal characters in path.
    at System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
    at System.IO.Path.GetFileName(String path)
    at Nancy.Conventions.StaticContentConventionBuilder.<>c__DisplayClass2.<AddDirectory>b__0(NancyContext ctx, String root)
    at Nancy.DefaultStaticContentProvider.GetContent(NancyContext context)
    at Nancy.NancyEngine.HandleRequest(Request request, Func`2 preRequest)
    at Nancy.NancyEngine.HandleRequest(Request request)
    at Nancy.Testing.Browser.HandleRequest(String method, String path, Action`1 browserContext)
    at Nancy.Testing.Browser.Get(String path, Action`1 browserContext)
    EncodingSpecs.cs(16,0): at Encoding.Test.EncodingSpecs.Getting()
```

We are using Nancy.0.18.0-Pre720 and .Net 4.5

PR #1898, opened 2015-04-16
fix exception & hang when request path contains invalid characters

etishor:
Currently Nancy will throw an exception and hang if the request path contains any character that is not a valid path character ( > < | * etc.) . The issue is present with any host that decodes the characters before invoking Nancy (tested with Owin & SelfHost, but probably all hosts are affected). 

This is similar to #1141 and the fix is similar to the fix in https://github.com/thecodejunkie/Nancy/commit/180cc73100fbf9099b4b5bc5f52da6c58810ef0c.

At the moment it is not possible to write an unit test for this as the Browser in Nancy.Testing will throw an exception if the path contains an invalid character. If the character is encoded the Browser will pass it to NancyEngine without decoding it. This is in contrast with the Hosts behavior, which is to decode the encoded characters before passing the request path to NancyEngine. 

To test, all you need to do is create a new NancyModule with a simple Get handler:

```csharp
Get["/test/{id}"] = p => Response.AsText((string)p.id);
```

Performing a GET request to /test/%3E results in an exception. ( ">" encoded is %3E )

The issue is in DefaultRequestDispatcher.Resolve(NancyContext context) where a call to Path.GetExtension(context.Request.Path) is made. 

The fix is to avoid calling Path.GetExtension if the request path contains any character from Path.GetInvalidPathChars().

Not sure of the severity of it, but I think this can be regarded as a security issue as an unhandled exception can be triggered and in some cases the request hangs consuming resources.

### DIFF
--- a/src/Nancy/Routing/DefaultRequestDispatcher.cs
+++ b/src/Nancy/Routing/DefaultRequestDispatcher.cs
@@ -154,8 +154,8 @@ namespace Nancy.Routing
 
         private ResolveResult Resolve(NancyContext context)
         {
-            var extension =
-                Path.GetExtension(context.Request.Path);
+            var extension = context.Request.Path.IndexOfAny(Path.GetInvalidPathChars()) >= 0 ? null
+               : Path.GetExtension(context.Request.Path);
 
             var originalAcceptHeaders = context.Request.Headers.Accept;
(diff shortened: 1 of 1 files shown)


khellang:
Thanks! :+1: This will go out very soon with v1.2 :smile:

Issue #1724, opened 2014-10-19
Unhandled exception when use Hosting.Self with port forwarding


ErwinCat:
Nancy.Hosting.Self version 0.23.2.0. Run it on local network PC. 
I use router with port forwarding from provider to local network(see image).
Request from local PC work fine, but when request comes from an external network, unhandled exception occurred: "System.InvalidOperationException: Unable to locate base URI for request: 95.78.85.102". I think Exception thrown at "NancyHost" class "ConvertRequestToNancyRequest" method:
```C#
private Request ConvertRequestToNancyRequest(HttpListenerRequest request)
        {
            var baseUri = this.baseUriList.FirstOrDefault(uri => uri.IsCaseInsensitiveBaseOf(request.Url));
```
Property "request.Url" contain external ip address from provider, and property "this.baseUriList" contain local ip address.
Is it bug or feature?
P.S. Sorry for my english
![1](https://cloud.githubusercontent.com/assets/1770739/4692210/ce2c0b74-5752-11e4-97dc-c51140afa71c.png)

abdullin:
This issue prevents use of NancyFx self-hosting on Windows Azure

grumpydev:
Odd, I have self hosted Nancy services in Azure, although they are running in VMs, how are you running yours?

grumpydev:
Does it dump the error to the error hook or just eat it completely?

jchannon:
You could also make use of Nancy Self Host exception handler `UnhandledExceptionCallback` https://github.com/NancyFx/Nancy/wiki/Self-Hosting-Nancy#host-configuration

abdullin:
1. Using the latest NancyFx
2. Elevated privileges
3. HTTPS endpoint

Launching things this way:

```
public override void Run()
		{
			var processes = ApiEnvironment.WireHandlers( _inputsRegistrar, _client );
			var nancyBootstrapper = ApiEnvironment.WireApi( _inputsRegistrar );


			var internalUri = new Uri(GetEndpointAsUri("Http"));
			using (var nancyHost = new NancyHost(nancyBootstrapper, internalUri))
			{
				nancyHost.Start();
				_log.Information( "Worker role us running" );
				var token = cancellationTokenSource.Token;
				var task = host.Start( token );

				token.WaitHandle.WaitOne();
				if( !task.Wait( 5000 ) )
					_log.Warning( "Host didn't stop in time." );
			}
		}

		private static string GetEndpointAsUri( string name )
		{
[...]

PR #1900, opened 2015-04-17
Added optional fallback base uri matching for port forwarded self host situations (such as azure)

grumpydev:
Workaround for #1724


grumpydev:
Merging my own PR.. :cow: :boy: :shipit:

PR #1903, opened 2015-04-20
Fix for #1881

### LINKED ISSUE #1881: HEAD request failed with return View (version > 0.21)
HEAD request works fine with return "string" or something
but 500 error code when I return View["name"] from module.

 Get["/"] = parameters =>
           {
                //return "meme"; //it works fine
                return View["index"]; // 500 error
            };

horsdal:
When making a `HEAD` request for a route returning a `MaterialisingResponse` a `NullReferenceExpception` was thrown in the content delegate.

`HeadResponse` now defers the access to the inner responses content till the content delegate is executed.

Fixes #1881

### DIFF
--- a/src/Nancy.Tests/Unit/HeadResponseFixture.cs
+++ b/src/Nancy.Tests/Unit/HeadResponseFixture.cs
@@ -1,6 +1,10 @@
 namespace Nancy.Tests.Unit
 {
     using System.Collections.Generic;
+    using System.IO;
+
+    using FakeItEasy;
+
     using Tests.Extensions;
     using Xunit;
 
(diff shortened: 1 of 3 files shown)


src/Nancy.Tests/Unit/HeadResponseFixture.cs
```diff
@@ -1,6 +1,10 @@
 namespace Nancy.Tests.Unit
 {
     using System.Collections.Generic;
+    using System.IO;
+
```
khellang: Unrelated: can we kill this using group whitespace? Preferably add the setting to the .DotSettings-file?

PR #1910, opened 2015-04-25
Tweaked overloads of Dependency and Dependencies

### LINKED ISSUE #1909: Browser with.Dependencies(..) broken?
_Notice: I'm not sure if this issue is related to #1862 or this is a TinyIOC issue and should be raised there._

When testing a route using `Nancy.Testing` and using `with.Dependencies(someCollectionOfDependencies);` it fails to correctly set them up.

As a work around, I need to iterate through each one and `with.Dependency(someDependency);` and _that_ works.

for example ..

*Fails*
```
var browser = new Browser(with =>
{
    with.Module<MyModule>();
    with.Dependencies(dependencies); // a collection of `x` dep's.
});
```
*Succeeds*
```
var browser = new Browser(with =>
{
    with.Module<MyModule>();
    foreach(var dependency in dependencies)
    {
        with.Dependency(dependency);
    }
});
```

btw, the collection `dependencies` are identical.

Lastly, this is with version `1.2` of Nancy, etc.

khellang:
Fixes #1909

The old `Dependencies` method took a `params object[] dependencies` argument, looped through them and called `this.Dependency<T>(dependency)` where `T` was inferred to `object` :confused: I can't see how this has ever worked.

This PR removes the overloads accepting `object` and `object[]` and force you to use an overload with an explicit (or inferred) generic type parameter.

### DIFF
--- a/src/Nancy.Testing.Tests/ConfigurableBootstrapperDependenciesTests.cs
+++ b/src/Nancy.Testing.Tests/ConfigurableBootstrapperDependenciesTests.cs
@@ -95,7 +95,7 @@
         class DisposableDependency : IIinterface, IDisposable
         {
             public bool Disposed { get; private set; }
-      
+
             public void Dispose()
             {
                 this.Disposed = true;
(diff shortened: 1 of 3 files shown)


khellang:
This could actually be a minor breaking change, since people could've used the old method to register all kinds of different types, and the compiler won't be able to infer the type anymore. The fix is as simple as helping the compiler out by specifying an explicit type, like I had to do [here](https://github.com/khellang/Nancy/commit/23183a21d2df73aaf72b57edfbd199f54728341e#diff-1ce2c3a049e7cd9dc2fada91c2cf11d9R128).

khellang:
Rebased :wink:

Issue #1905, opened 2015-04-23
DefaultStatusCodeHandler should negotiate content and provide a JSON response


jonathanfoster:
It would be nice if the `DefaultStatusCodeHandler` negotiated content and provided a JSON response instead of just the custom 400 and 500 error pages.

khellang:
Agreed :+1: Are you up for the task?

jonathanfoster:
I'll pull this issue.

jonathanfoster:
Pull request 1911 (https://github.com/NancyFx/Nancy/pull/1911) adds this feature. Take a look at the PR and let me know if there's any feedback.

khellang:
Fixed by #2081

khellang:
Thanks, @jonathanfoster! :grin:

jonathanfoster:
Thank you guys for making an awesome framework. I'm just happy I could contribute.

PR #1911, opened 2015-04-25
Add content negotiation to the default error handler

jonathanfoster:
...r now supports application/json, application/xml, and text/html. The default content type is text/html when all media types are supported (\*/\*). If a non-supported media type is provided then the response status code will be 406 not acceptable.

I made a change to `ContextExtensions.GetExceptionDetails` that could be considered breaking. I changed the method to return `string.Empty` instead of "None" when no exception is found. I didn't see any other usages in the solution, but I could see a situation where a consumer of the library has a check for "None" that will now fail. I don't think it's a huge concern, but something to consider. I added a unit test to ensure the method returns `string.Empty` in the future. If there's an issue with this change then I could always revert it. I just thought that `string.Empty` is a more idomatic return value in this situation.


jonathanfoster:
@thecodejunkie: Any feedback on this PR? This resolves issue #1905.

khellang:
@jonathanfoster Currently you're duplicating *a lot* of code from the `IResponseNegotiator` service. Would it be possible to use that instead? The upside is that you'd then be able to leverage custom formatters etc. for errors :smile:

jonathanfoster:
Duplicate code is no bueno. I'll dive into the Content Negotiation wiki page so I can get a better idea of where the `IResponseNegotiator` fits in.

thecodejunkie:
@jonathanfoster did you ever resolve the code duplication issue? Thanks

thecodejunkie:
Ping @jonathanfoster :)

jonathanfoster:
I'm not dead yet! No, I haven't resolved the duplication issue yet. It took me some time to grok the code base and I think I have a better plan for integration now. I had done some spikes and I can't remember where I left off before getting pulled into something else. I'll take a look again this weekend to see if I can wrap things up. If not I'll close the PR and let someone else takeover.

khellang:
:+1:

jonathanfoster:
So I was able to remove the duplicate code by leveraging the `IResponseNegotiator`. Now I'm running into some challenges with rendering the default error pages due to the `ViewProcessor` handling HTML accept headers. I'm looking into the default view engine classes to see if there's an opportunity to short circuit the default view resolution and render a response from the static error files instead.

jonathanfoster:
@thecodejunkie, everything is working now. I wrapped the call to `responseNegotiator.NegotiateContent` in a try catch and if a `ViewNotFound` exception is thrown I render a response from the embedded HTML resource files. I had some reservations about this approach since it uses an exception for control flow, but in the end it was the cleanest solution I could find. 

This PR now includes 2 potential breaking changes. The `DefaultStatusCodeHandler` no longer has a parameter-less constructor, the constructor requires an instance of `IResponseNegotiator`. This may not be an issue since all instances are constructed using the IoC container. The `ContextExtension.GetExceptionDetails` method now returns `string.Empty` instead of 'None'. This is a more idiomatic return value especially when viewing the output in JSON or XML.

src/Nancy/ErrorHandling/DefaultStatusCodeHandler.cs
```diff
@@ -62,47 +66,60 @@ public void Handle(HttpStatusCode statusCode, NancyContext context)
                 return;
             }
 
-            string errorPage;
-
-            if (!this.errorPages.TryGetValue(statusCode, out errorPage))
+            if (!this.errorMessages.ContainsKey(statusCode) || !this.errorPages.ContainsKey(statusCode))
```
horsdal: Is this condition correct? We bail out if the `statusCode` is not in both dictionaries. Is that right?

src/Nancy/ErrorHandling/DefaultStatusCodeHandler.cs
```diff
@@ -62,47 +66,60 @@ public void Handle(HttpStatusCode statusCode, NancyContext context)
...
             {
-                context.Response = new Response() { StatusCode = statusCode };
+                context.Response = new Response { StatusCode = statusCode };
             }
 
+            var contents = this.errorPages[statusCode];
+
+            if (!string.IsNullOrEmpty(contents))
+            {
+                contents = contents.Replace("[MESSAGE]", result.Message).Replace("[DETAILS]", result.Details);
```
horsdal: I don't like that we go in and replace part of the content of the `.html` file. Why not keep the error pages static?

src/Nancy/ErrorHandling/DefaultStatusCodeHandler.cs
jonathanfoster: Yes, that's correct. This is a protection against calling this method without first checking `HandlesStatusCode`.

src/Nancy/ErrorHandling/DefaultStatusCodeHandler.cs
jonathanfoster: I could certainly do that instead. I was just extending the existing convention of replacing the `[DETAILS]` to keep the content consistent regardless of negotiation. I really have no strong opinions either way. Would you like to keep the message static and still dynamically render the details like the current implementation?

src/Nancy/ErrorHandling/DefaultStatusCodeHandlerResult.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// Default error handler result
+    /// </summary>
+    public class DefaultStatusCodeHandlerResult
+    {
+        /// <summary>
+        /// Gets or sets the HTTP status code
+        /// </summary>
+        public HttpStatusCode StatusCode { get; set; }
```
jonathanfoster: @horsdal, what are your thoughts on including the HTTP status code in the error handler result? Technically this is duplicated in the HTTP response status.

src/Nancy/ErrorHandling/DefaultStatusCodeHandler.cs
```diff
@@ -62,47 +66,60 @@ public void Handle(HttpStatusCode statusCode, NancyContext context)
...
+            try
+            {
+                context.Response = this.responseNegotiator.NegotiateResponse(result, context);
+                context.Response.StatusCode = statusCode;
                 return;
             }
-
-            Func<HttpStatusCode, NancyContext, string, string> expansionDelegate;
-            if (this.expansionDelegates.TryGetValue(statusCode, out expansionDelegate))
+            catch (ViewNotFoundException)
```
jonathanfoster: @horsdal, do you have any issue using the `ViewNotFoundException` here? Initially I was concerned about using an exception for control flow, but it was the simplest solution I could find to short circuit the view rendering.

src/Nancy/ErrorHandling/DefaultStatusCodeHandlerResult.cs
horsdal: That seems ok to me

src/Nancy/ErrorHandling/DefaultStatusCodeHandler.cs
horsdal: Sorry. I don't understand that. How is this protecting against calling without first checking the status code?
I read it as ensuring that there is always a message _and_ an error page for the status code.

src/Nancy/ErrorHandling/DefaultStatusCodeHandler.cs
horsdal: Well, an exception for control is a red flag of course. I haven't looked closely enough to see if there are alternatives.

src/Nancy/ErrorHandling/DefaultStatusCodeHandlerResult.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Nancy.ErrorHandling
+{
+    /// <summary>
+    /// Default error handler result
+    /// </summary>
+    public class DefaultStatusCodeHandlerResult
```
khellang: This is just an implementation detail of the `DefaultStatusCodeHandler`, right? Maybe this should just be a private class?

src/Nancy/ErrorHandling/DefaultStatusCodeHandler.cs
jonathanfoster: You're right, this protects against a status that doesn't have a message and an error page. I was referring to calling code that doesn't check `HandlesStatusCode` before calling `Handle`. Then someone could inadvertently pass in a non-supported status.

src/Nancy/ErrorHandling/DefaultStatusCodeHandler.cs
jonathanfoster: @horsdal, I reverted the change and now only the error details are dynamic.

src/Nancy/ErrorHandling/DefaultStatusCodeHandlerResult.cs
jonathanfoster: I made this an internal class since it's referenced in functional tests.

src/Nancy/ErrorHandling/DefaultStatusCodeHandler.cs
horsdal: OK, I think I finally understand :sleeping:

horsdal:
Replaced by #2081

Issue #1897, opened 2015-04-16
Streaming large requests


adamhathcock:
I'm implementing a file upload and the pay load goes directly to another service (S3 in this case.)

I was noticing everything gets buffered and it looks like Nancy always buffers requests.  I found this issue: https://github.com/NancyFx/Nancy/issues/1014 And it looks like the conclusion was just a bug but no real answer other than my choice is between a MemoryStream or a FileStream.

The only reason I see that a seekable stream is required is for the multipart processing.  Is that correct?  Is there a way to enable a non-seeking stream path for a request or is everything built around having a seekable request?

Thanks.

alanedwardes:
I am also interested in this.

adamhathcock:
Looking deeper at RequestStream, it looks like the primary reason this stream is seekable is for parsing multipart requests.  It seems if the buffering of the request stream could be lazy, then it would solve the problem.

However, it looks like the RequestStream object itself is writable primarily for test cases.  Being writable feels wrong.  Would a largish change to make the RequestStream lazy and readonly be welcome at all?

khellang:
@adamhathcock It's hard to say what the implications is without seeing any code. I assume there's a reason it's done as it is today. @grumpydev?

grumpydev:
It's as it is for multipart requests where you need to be able to seek back and forth, as long as it switches to a filestream I'm not sure I understand the problem - you'd never stream directly to the "destination file", you'd want to make sure the request was complete before you potentially overwrite what was already there? If what you're building is essentially just a proxy for a request to s3 why not just upload straight to s3 with a secure url? Failing that I'd probably fall back on something more lightweight like potentially a raw OWIN middleware for handling this - using a web framework, even a light one like nancy, as a proxy seems pretty heavy to me.

adamhathcock:
S3 is just an example really.  Though we do proxy files to various things.

Another scenario is that where we have a continuous stream that is an array of JSON objects.  On the response side, we gather the relevant data create the object and send it on the response with zero buffering as there can be a lot of this data.  We'd want to be able for a client to send a continuous stream for the server to process.  The server would only have to validate a single JSON object at a time.  If the stream dies early, it doesn't invalidate the successful objects received.

I guess I wanted to stick with Nancy as I can have non-buffered responses and code centered around Nancy and other things.  I wanted the request side to be able to look the same.

grumpydev:
[...]

PR #1920, opened 2015-04-30
Non buffered requests

adamhathcock:
Issue: https://github.com/NancyFx/Nancy/issues/1897

RequestStream doesn't buffer by default and is read only.  Buffering is done when requested (by Form and Multipart data).

Buffering is only done lazily when Form or Multipart data is accessed on the Request object.

Buffering to a file is still done as before.

Lots of tests relied on the fact that requests buffered and reset positions to 0.  Some tests were removed or majorly as they did stuff like test writing to the RequestStream


cvetomirtodorov:
I am applying this PR to a fork of mine to test it out. I think you should mention that buffering is done when using model binding, too.

I think it *may* be omitted, if the support for multiple binding is dropped. When I removed the buffering in the default binder there was just a single test failure related to that and nothing else. Or alternatively - a separate method is added `Bind(multiple: true)` which should trigger the buffering. While the default `Bind()` (delegating to `Bind(multiple: false)`) does not.

adamhathcock:
It looks like I have DefaultBinder buffering always.  I guess I'm not too familiar with the binding to make it lazy.  Though, I'm not sure why buffering for the model would be bad.  You're effectively buffering for the binding anyway.

If you want to make it lazy, I say go for it :)

cvetomirtodorov:
Well, additional API methods may be out of scope of the current PR, although it is related somehow. Nancy owners should have a major say about this before any implementation is to be added. But I do think that binding *once* should not *require* buffering. The question is what the API for multiple binding should look like.

cvetomirtodorov:
@adamhathcock any particular reason to explicitly not support seeking in request stream? I changed it to the old implementation which asks the underlying `this.stream` and it works fine for my case but wanted to know the reasoning behind this.

adamhathcock:
The old implementation always buffers to memory or a file.

I want to only buffer the request body when required because I have scenarios with large request payloads that shouldn't be buffered

cvetomirtodorov:
I meant these changes:

```c#
	public override bool CanSeek
	{		         
		get { return false; }
	}
	public override long Seek(long offset, SeekOrigin origin)
	{
		throw new NotSupportedException();
	}
```

My question was whether there is a need not to delegate these calls to `this.stream` as before. If the request stream is not buffered and does not support seeking then it would return `false` and `Seek` shouldn't be called by client. When the request stream is buffered though, it may be useful to be able to `Seek` in the stream.

adamhathcock:
OH I see what you're saying.

I guess I'm breaking pattern by allowing Position to be set when buffered but not allowing CanSeek and Seek to be used if buffered.

That should change, I'll do that shortly.

cvetomirtodorov:
I think you had it right and because of the unit tests failing it got reverted back instead of the tests being fixed.

The point is: even **non-buffered requests should support seeking** if the underlying stream does. That's just read-related and I can't think of a reason why it should be disallowed if the stream supports it. That was the behavior before.

adamhathcock:
The actual HTTP Request Stream doesn't support seeking.  You can only seek when the Stream gets buffered.  That's what I'm trying to codify anyway.  The fact that the implementation replaces the request stream with another stream object is an implementation detail.

cvetomirtodorov:
* If the actual HTTP Request Stream does not support streaming its `CanSeek` property will return `false` anyway and its `Seek` method will throw the exception, too. No need to write checks that are not required. And code gets simpler, too.
* If another underlying stream **does support** seeking though and request is not buffered you're limiting that stream capabilities.
* I think it's not good to be able to seek "only when the Stream gets buffered". You should be able to seek when the stream supports it. That's why there's `CanSeek` property in the first place.

adamhathcock:
The RequestStream class isn't supposed to be generic.  It's specific to a HTTP Request Stream.  It just can't be hardcoded to any specific request stream type.

I like the explicit "IsBuffered" to know that the stream isn't the http request stream anymore.

I get what you're saying but don't think it should support every scenario.

cvetomirtodorov:
Nancy supports lots of hosts already and adding different hosts is encouraged. So I think it should be as generic as possible and should support every scenario possible. The implementation I suggested is not hard and it's what it is in the main branch. Also, if you want to distinguish the stream - you can always just add an `IsBuffered` or `IsRaw` or something-like-that property.

adamhathcock:
Any new comments or thoughts on this?

khellang:
Assigning this to 2.0 so we can properly test this as a beta release or something.

thecodejunkie:
Just to elaborate on what @khellang said. This is touching a very sensitive part of the framework and we need to make sure that it is properly evaluated and tested before it hits a stable build.

adamhathcock:
No problem.  Thanks for looking.

jchannon:
@adamhathcock can you rebase this. if @NancyFx/most-valued-minions and @thecodejunkie have on issues lets merge this

adamhathcock:
rebased.  Though it looks like the Travis CI has a keyserver timeout issue?

thecodejunkie:
@jchannon @adamhathcock this https://github.com/NancyFx/Nancy/pull/1920#issuecomment-171224806 still applies :D It's going to require some focused testing and not just a code review. It's likely that this won't make it into 2.0 because of time constraints, but rather 2.1 or another point release

Issue #1865, opened 2015-03-12
Garbage added to multipart-formdata running selfhost on mono


umby24:
I am running a Nancy self host, using v1.10.0 from NuGet in my project. Everything runs fine under Windows with .NET 4.5, however issues arise when submitting larger files (~ 9 MB) to the nancy self host on mono.

The files become corrupted by Nancy when running in Mono. This is evidenced by several things. First is a wireshark capture running on the linux machine that is able to carve the submitted file successfully with no corruption.

Next is that the HTTP Headers content length field is of different size than the http body, because null bytes become added to the http body for some reason.

The corruption always seems to occur at the same place in the file, and is always null bytes (0x00)

This is example code for that should show the issue, when sending a ~9 MB file.

```C#
Post["/fileupload"] = x => {
                var myFile = Request.Files.FirstOrDefault();
                byte[] data;

                if (myFile != null) {
                    data = new byte[myFile.Value.Length];
                    Log(LogType.Debug, "Body Length: " + data.Length);
                    Log(LogType.Debug, myFile.Value.Read(data, 0, data.Length).ToString());

                    using (var fs = new FileStream("out.zip", FileMode.Create)) {
                        myFile.Value.CopyTo(fs);
                    }
                }
                else {
                    Log(LogType.Debug, "File is null.. " + Request.Headers.ContentLength);
                    Log(LogType.Debug, "Info: " + Request.Body.Length + " " + Request.Headers.ContentType);
                    data = new byte[Request.Headers.ContentLength];
                    Request.Body.Read(data, 0, data.Length);
                }
}
```

The else statement is always taken (File is null), if that information helps.

starock:
Me too...

cvetomirtodorov:
Occurs in Nancy 1.2 too with Mono 3.12.1 on Ubuntu 14.04.

cvetomirtodorov:
After debugging for a while I think the reason for the problem is explained in this bug I raised https://bugzilla.xamarin.com/show_bug.cgi?id=29337.

Problem is that I could not make things work without specifying `useAsync: false` when creating the `FileStream` in `RequestStream.CreateTemporaryFileStream` method. Looks like async operations on the `FileStream` in mono are not working well with large streams. One temporary solution would be to set `useAsync` to `false` when on mono, although that would degrade performance I suppose.

Also, I think the issue https://github.com/NancyFx/Nancy/issues/1796 is related to this one.

cvetomirtodorov:
[...]

PR #1925, opened 2015-05-04
Issue with RequestStream switching behavior on OSX.

cchrisman:
Disabling streamSwitching for HTTP request body data, this appears to cause problems when the body size is larger than the switching threshold
on OSX.

See issue: https://github.com/OmniSharp/omnisharp-sublime/issues/149

### DIFF
--- a/src/Nancy.Hosting.Self/NancyHost.cs
+++ b/src/Nancy.Hosting.Self/NancyHost.cs
@@ -279,7 +279,7 @@
             return new Request(
                 request.HttpMethod,
                 nancyUrl,
-                RequestStream.FromStream(request.InputStream, expectedRequestLength, false),
+                RequestStream.FromStream(request.InputStream, expectedRequestLength, true),
                 request.Headers.ToDictionary(), 
                 (request.RemoteEndPoint != null) ? request.RemoteEndPoint.Address.ToString() : null,
                 certificate,


cchrisman:
I initially found this as the source of my problems by reading the request.InputStream in its entirety, and then re-passing it back as a memory stream to RequestStream.FromStream.  This prevented the stream corruption, but passing request.InputStream directly causes it to fail; again this was only observed on OSX.

I'm not sure how important the stream switching functionality is in the general case.  My suggestion would be to provide some way to disable it (since I can imagine that some applications always want an in-memory stream, even when it's working as intended).

grumpydev:
I believe this is a Mono issue that's fixed in the latest version of Mono - see #1865 - making it an option on self host is probably a reasonable idea, although disabling it does leave you open to an easy denial of service attack unless you had additional middleware in place to prevent it.

Closing this for now, but happy to take another PR where this flag is taken from the selfhost configuration.

jchannon:
I'm reopening this. Mono 4 is unstable as hell and I think an addition to this PR to make the using of async configurable would be better. We may also need to change RequestStream as shown here https://github.com/vladlopes/Nancy/commit/f5e6d43cec6f344b76d5883a392fa7407f96a2a2

khellang:
@jchannon The request stream switching behavior is configurable using `StaticConfiguration.DisableRequestStreamSwitching` (See #2015) from the next version.

jchannon:
#2015 doesn't change RequestStream https://github.com/vladlopes/Nancy/commit/f5e6d43cec6f344b76d5883a392fa7407f96a2a2

khellang:
@jchannon What are you talking about? This PR modified the exact same line as #2015!?

khellang:
@jchannon The commit you linked changes [`FileStream`](https://msdn.microsoft.com/en-us/library/7db28s3c(v=vs.110).aspx)'s `useAsync` parameter from `true` to `false`. It has nothing to do with this PR, *or* #2015 :confused:

jchannon:
I'm going to do a new PR that modifies RequestStream to use the StaticConfiguration property made in #2015

khellang:
@jchannon No! Don't do that. The `DisableRequestStreamSwitching` config is for **request switching**. It has nothing to do with the `useAsync` parameter of the `FileStream`.

jchannon:
Ok I'll add another static configuration parameter then

PR #1928, opened 2015-05-05
Prevent body deserializers from throwing if IgnoreErrors is set to true

### LINKED ISSUE #1927: ModelBinding - Stop JSON.NET from erroring when bad input
If i have a model like this:

```
public class Filter
{
   public int[] Bedrooms{ get; set; }
}
```

With [model binding config](https://github.com/NancyFx/Nancy/wiki/Model-binding#binding-configuration) like this:

`this.Bind<Filter>(new BindingConfig { IgnoreErrors = true } );`

Why does a request like this:

```
{
   Bedrooms: 1
}
```

Cause the following error:
image

I thought the point of `IgnoreErrors = true` was to well, ignore binding errors? Or am i misunderstanding something?

I'm on Nancy v1.2.0.

Thanks!

khellang:
Fixes #1927

### DIFF
--- a/src/Nancy.Tests/Unit/ModelBinding/DefaultBinderFixture.cs
+++ b/src/Nancy.Tests/Unit/ModelBinding/DefaultBinderFixture.cs
@@ -62,6 +62,40 @@ namespace Nancy.Tests.Unit.ModelBinding
         }
 
         [Fact]
+        public void Should_throw_if_body_deserializer_fails_and_IgnoreErrors_is_false()
+        {
+            // Given
+            var deserializer = new ThrowingBodyDeserializer<FormatException>();
+            var binder = this.GetBinder(bodyDeserializers: new[] { deserializer });
+
+            var context = CreateContextWithHeader("Content-Type", new[] { "application/xml" });
+
+            var config = new BindingConfig { IgnoreErrors = false };
+
+            // When
+            var result = Record.Exception(() => binder.Bind(context, this.GetType(), null, config));
+
+            // Then
+            result.ShouldBeOfType<ModelBindingException>();
+            result.InnerException.ShouldBeOfType<FormatException>();
+        }
+
+        [Fact]
+        public void Should_not_throw_if_body_deserializer_fails_and_IgnoreErrors_is_true()
+        {
+            // Given
+            var deserializer = new ThrowingBodyDeserializer<FormatException>();
+            var binder = this.GetBinder(bodyDeserializers: new[] { deserializer });
+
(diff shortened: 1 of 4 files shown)


src/Nancy.Tests/Unit/ModelBinding/DefaultBinderFixture.cs
```diff
@@ -62,6 +62,40 @@ public void Should_throw_if_body_deserializers_is_null()
...
+        [Fact]
+        public void Should_not_throw_if_body_deserializer_fails_and_IgnoreErrors_is_true()
+        {
+            // Given
+            var deserializer = new ThrowingBodyDeserializer<FormatException>();
+            var binder = this.GetBinder(bodyDeserializers: new[] { deserializer });
+
+            var context = CreateContextWithHeader("Content-Type", new[] { "application/xml" });
+
+            var config = new BindingConfig { IgnoreErrors = true };
```
horsdal: You could extract these 4 setup lines from this test and the one above - what do you think? Going to far?

src/Nancy/ModelBinding/DefaultBinder.cs
```diff
@@ -93,20 +92,29 @@ public object Bind(NancyContext context, Type modelType, object instance, Bindin
             var bindingContext =
                 this.CreateBindingContext(context, modelType, instance, configuration, blackList, genericType);
 
-            var bodyDeserializedModel = this.DeserializeRequestBody(bindingContext);
+            var bindingExceptions = new List<PropertyBindingException>();
```
horsdal: Why is this decalration moved up? Are all declarations in the top of the method, or?

horsdal: Why is this decalration moved up? Are all declarations in the top of the method, or?

horsdal:
Overall :+1: Just a couple of nitpicks.

src/Nancy/ModelBinding/DefaultBinder.cs
khellang: That is a good question. I think I wanted to add an exception to the list at some point, but decided to just throw right away instead. I'll move it back :smile:

src/Nancy.Tests/Unit/ModelBinding/DefaultBinderFixture.cs
khellang: Meh. I'd rather don't do this, simply because I don't even want to try to come up with a good method name :stuck_out_tongue_closed_eyes:

khellang:
Rebased :dancers

PR #1945, opened 2015-05-26
Additional call to UpdateTypes in Bootstrap. Fixes #1942

### LINKED ISSUE #1942: FluentValidation doesn't work on vnext
Not sure what the policy is for reporting vnext bugs as it's currently not supported, but going to open this anyway. Maybe you want to create a label for vnext specific issues? 

When trying to use FluentValidation with Nancy on vnext an error is thrown when calling BindAndValidate<T>: 

'Nancy.Validation.ModelValidationException' occurred in Nancy.dll but was not handled in user code Additional information: No model validator factory could be located. Please ensure that you have an appropriate validation package installed, such as one of the Nancy.Validation packages.

I have Nancy.Validation.FluentValidation installed. I also have the appropriate registrations done in the Bootstrapper. You can see an example set of code in my Gist here: https://gist.github.com/awec/af94a240956b68e81e2b
 
Tried to debug this myself and couldn't figure it out cause I don't really know how Nancy works at that level. The modules window in Visual Studio suggests that Nancy.Validation.FluentValidation.dll is indeed loaded, but for some reason is not detected when the BindAndValidate method is called.

awec:
As per request.

### DIFF
--- a/src/Nancy/Bootstrapper/NancyBootstrapperBase.cs
+++ b/src/Nancy/Bootstrapper/NancyBootstrapperBase.cs
@@ -244,6 +244,9 @@
 
             this.ConfigureApplicationContainer(this.ApplicationContainer);
 
+            // We need to call this to fix an issue with assemblies that are referenced by DI not being loaded
+            AppDomainAssemblyTypeScanner.UpdateTypes();
+
             var typeRegistrations = this.InternalConfiguration.GetTypeRegistations()
                                         .Concat(this.GetAdditionalTypes());
 


phillip-haydon:
Possible Stack Overflow issue that may also be resolved in V1.3

http://stackoverflow.com/questions/30471563/simpleauthencation-throwing-failed-to-reflect-on-the-current-domains-assemblie

thecodejunkie:
@awec Hi, for our v2 drop (the current master) I think I am going to drop this call. In all aspects it a bandaid on what could be a broken leg. Part of v2 is that we will be replacing the `AppDomainAssemblyTypeScanner` (which is a static class, that loads stuff in a static constructor) and replace them with #2217 so we will be having instance based scanner instead, which are created inside the bootstrapper. I'm currently working on the PR to remove ADATS from the bootstrapper and it should go in during the day

Thanks

Issue #1938, opened 2015-05-20
DefaultModelBinder requires public default constructor


jvandertil:
The default model binder requires the existence of a public default constructor. 
Is it possible to allow non public default constructors as well?

Root cause seems to be:
https://github.com/NancyFx/Nancy/blob/master/src/Nancy/ModelBinding/DefaultBinder.cs#L432
and
https://github.com/NancyFx/Nancy/blob/master/src/Nancy/ModelBinding/DefaultBinder.cs#L436

I can work on a pull request to fix this, but I'm wondering if there is a good reason to not allow this.

khellang:
Landed with #1946

PR #1946, opened 2015-05-27
Allow binding to models with non public no args ctors

jvandertil:
Since my issue #1938 seems to be getting no attention, I went ahead and fixed it in my fork.

This resolves some issues for me when model binding directly to 'domain' commands, since I don't want to promote newing them up in an inconsistent manner I've hidden the no args constructor.


khellang:
Fixes #1938

PR #1952, opened 2015-06-06
Do not store nuget packages in git, use nuget to retrieve them

tparnell8:
You probably should not store packages in git. Packages stored in this manor can bloat repositories, and are often a source of merge conflicts.

Instead retrieving them at build time is more appropriate


tparnell8:
This is failing, because you are not running bundle with no params on start...

khellang:
I like this change! I wonder why the packages was checked in in the first place? I think it was something with Mono. @grumpydev

phillip-haydon:
Because at the time the project was created we couldn't restore from nuget on Linux. Now its just a decoding factor... Does it hurt having them checked in? I don't think so... 


Sent by Outlook for Android





On Mon, Jun 8, 2015 at 2:26 AM -0700, "Kristian Hellang" <notifications@github.com> wrote:












I like this change! I wonder why the packages was checked in in the first place? I think it was something with Mono. @grumpydev

grumpydev:
Yep, it was for Mono where nuget wasn't supported, and it's still a bit hit and miss.

tparnell8:
this should work on 2.0+ and is proved by the passing travis build. Avoiding binaries in the repo is key to keeping it small, and thus cheap to use.

phillip-haydon:
What becomes cheap to use?

grumpydev:
Cheap to use? What? :)

grumpydev:
And Nuget didn't work on 2.0 at all, it was 3.2 or so and above, which is fairly recent, where it was anything other than a right royal mess. The Travis build passes *because* the binaries are there :)

tparnell8:
I deleted the binaries in my PR, and I'm talking about git itself. Git is notoriously bad at storing binary files.

phillip-haydon:
Git is bad for changes on binaries. Yes. But it handles large files just fine. Its just not optimal to store large files like PSDs and such in Git...

phillip-haydon:
So I'm not sure what is gained by not having the binaries in the repo other than longer build times.

khellang:
I'm closing this for now. Thanks for the effort @tparnell8! We will most likely revisit this once we get on with 2.0. That means we're moving to newer version of Mono etc. and probably also to a project.json-based system.

Issue #1953, opened 2015-06-07
Builds do not run bundle


tparnell8:
I noticed that the builds on travis run `gem install albacore`, `gem install otherGems` instead you should run `bundle` this ensures that the gemfile is always correct, and if someone alters the bundle file as part of their PR, the requested gem(s) are pulled down as part of the build.

tparnell8:
you will of course still have to `gem install bundle`

thecodejunkie:
My brain might be a bit tired tonight, but could you elaborate a bit more on this? Thanks

PR #1956, opened 2015-06-08
use bundle instead of installing gems seperatly

tparnell8:
#1953 Probably want to run bundle and use the gem file...


tparnell8:
I assume someone will have to alter the Team City build. This lets PR's use other gems.

khellang:
I'm not really into Ruby, rake, bundle and gems. How come TC says this is all good?

tparnell8:
The main problem is that if someone submits a PR to alter the rake file that will require a new gem (ruby version of a nuget package) the build will fail. This is because you have to call bundle before you start calling rake. bundle is essentially a gem that will download all the gems listed in the gemfile. Very similar to calling a nuget restore. This PR contains the commands to alter travis, but you will also need to make the modification in TC, by simply having it call bundle before it does any rake tasks.

tparnell8:
As a side note, you are using the release candidate of version 1 of albacore, albacore is now on 2.4.2. You may want to consider updating at some point.

grumpydev:
The bundle file was there because it looked like Travis needed it but it isn't actually used.. Whether we delete it or change scripts to use it I am not too fussed. 

As for albacore, there are too many breaking changes to be worth the effort of switching versions, it would effectively mean a complete rewrite of our build scripts, which just isn't worth investing the time in when they work fine as they are.

thecodejunkie:
Did we reach a consensus on this one?

tparnell8:
this change fixes the gem file. Basically its like a packages.config. My change pins ruby to the specific version of albacore that is required by your rake file.

Anyone familiar with ruby will try to run bundle (which is like nuget restore) which will give them the latest version of albacore. The problem (which again is not too big of a deal) is that your rakefile needs a specific version of albacore. This was why you were having travis installing the specific version of albacore, but what you really want (what is best practice in ruby) is to fix the gemfile, and then have travis use the gemfile.

If you get a newer version of albacore (like version 2) the rake file will not work (which was also why your travis build was doing . 

When I checked this out on my ubuntu box I could not do a build with rake because I needed the proper gem (which was not specified in the gemfile).

khellang:
:+1:

PR #1957, opened 2015-06-08
Leverage the RequiresHttps method from SecurityHooks.

lijunle:
The method has already been [there](https://github.com/NancyFx/Nancy/blob/82b5a38a7c06cbc764a1899427fd8019603324ac/src/Nancy/Security/SecurityHooks.cs#L129), with the same content. I don't know why just use it.


horsdal:
thx

PR #1964, opened 2015-06-18
Exclude object methods from diagnostics

### LINKED ISSUE #1960: C# object functions (ToString, Finalize) incorrectly show up in diagnostic interface
Upgrading to Nancy 1.0 (from 0.23) caues diagnostics providers to include functions like ToString, MemberwiseClone, Equals, GetHashCode, GetType and Finalize. In short, all of the basic functions in C# object. It seems like it would be a good idea to filter out those functions before so they don't get shown in the diagnostic interface.

BindingFlags.DeclaredOnly changed to BindingFlags.FlattenHierarchy here: https://github.com/NancyFx/Nancy/commit/e346b627664a74ed1fe11d5e6d805baaa0527ad2#diff-6ff47c4d9e7ab5c48c0fbcdff8ca056dR12

khellang:
Fixes #1960


phillip-haydon:
:thumbs: looks good to me.

PR #1967, opened 2015-06-23
Fixed implicit cast from relative Uri to Url

### LINKED ISSUE #1966: Nancy.Testing.Browser / System.Uri throws InvalidOperationException for Relative Uris
I just stumbled upon a bit unexpected behaviour in Nancy's (.Testing) Browser:

When making calls on a Browser instance with relative string locations, i.e. "/users" everything works fine and as expected, however passing in a System.Uri with UriKind.Relative throws an InvalidOperationException.

Basically the problem is this:

```
var testBrowser = new Nancy.Testing.Browser(TestBootstrapper, with => { with.HttpRequest(); });
var targetPathAsString = "/users";
var targetPathAsRelativeUri = new System.Uri("/users", UriKind.Relative);

// works:
var responseFromStringPath = testBrowser.Post(targetPathAsString, browserContext => browserContext.JsonBody(someModelInstance, mySerializerInstance));

// does not work / throws InvalidOperationException:
var responseFromRelativeUri = testBrowser.Post(targetPathAsRelativeUri, browserContext => browserContext.JsonBody(someModelInstance, mySerializerInstance));
```

The problem seems to originate from https://github.com/NancyFx/Nancy/blob/master/src/Nancy.Testing/Browser.cs#L302 and https://github.com/NancyFx/Nancy/blob/master/src/Nancy.Testing/Browser.cs#L303 .. basically the two requestUrl Properties (.Scheme and .Hostname) cannot be accessed if the Uri Kind is Relative.. as the Uri property getters throw the exception if the Uri instance is not an Absolute one, see http://referencesource.microsoft.com/#system/net/System/URI.cs,1212 ..

Is this a Bug or a "Feature" of Nancy and I am doing it just wrong? 

Cheers,
-J

khellang:
Closes #1966

### DIFF
--- a/src/Nancy.Tests/Unit/UrlFixture.cs
+++ b/src/Nancy.Tests/Unit/UrlFixture.cs
@@ -227,7 +227,25 @@ namespace Nancy.Tests.Unit
         }
 
         [Fact]
-        public void Should_implicitliy_cast_to_uri()
+        public void Should_implicitly_cast_from_relative_uri()
+        {
+            // Given
+            var uri = new Uri("/hello", UriKind.Relative);
+
+            // When
+            Url result = uri;
+
+            // Then
+            result.Scheme.ShouldEqual(Uri.UriSchemeHttp);
+            result.HostName.ShouldEqual(string.Empty);
+            result.Port.ShouldEqual(null);
+            result.BasePath.ShouldEqual(string.Empty);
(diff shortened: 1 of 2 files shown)


grumpydev:
:shipit:

khellang:
FFFFFFFFUU! :sob:

khellang:
Phew :sweat_smile:

PR #1969, opened 2015-06-24
Added quotes to netsh command parameters

### LINKED ISSUE #1968: netsh url reservation fail for username with whitespaces
netsh is run with the user parameter specified without quotes.
When user contains whitespaces, like "Power Users", netsh fails.
A workaround is to specify the quotes :
User="\"Power Users\""
However, the main problem is that for example, on french computers, the default group "Everyone" resolves to "Tout le monde" and reservation fails.
Could be fixed by modifying NetSh.cs and add quotes if the user contains whitespaces.

khellang:
Fixes #1968

### DIFF
--- a/src/Nancy.Hosting.Self/NetSh.cs
+++ b/src/Nancy.Hosting.Self/NetSh.cs
@@ -7,7 +7,7 @@
     /// </summary>
     public static class NetSh
     {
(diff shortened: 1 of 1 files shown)


phillip-haydon:
Hah I ran into this issue ages ago and thought it was me screwing stuff up so never looked into it.

PR #1970, opened 2015-06-24
Make sure we don't use C# 6

khellang:
Added LangVersion = 5 to make sure we don't accidentally use C# 6 (for now).

Since I automated this job (there's no way in hell I'd manually edit or use the GUI to to this), I figured I might as well self-close some empty tags and remove those without attributes :lipstick:


phillip-haydon:
FAIL.

khellang:
Argh! Mono... Why do we still build for Mono 2.0?

grumpydev:
Because mono3 is a buggy pile of cack generally :)

phillip-haydon:
I think we should look at the latest version of Mono, anything under 3.10 I had problems with but after 3.10 haven't had much if any issues. 

Would be good to ditch 2.0.

khellang:
I won't bother with this. We *desperately* need to update our Mono infra to support newer stuff.

thecodejunkie:
@khellang seeing as we now target latest mono (4.2.2) and C# 6 syntax is supported as of Mono 4.0, maybe we should force the langversion to 6 ?

khellang:
I don't think there's a need to force `LangVersion`.

Since it's not specified anywhere, VS will assume you want to use latest, so if you have VS2015, it wants you to use C# 6.

The reason this was proposed was because we didn't want it to suggest using C# 6 because Mono didn't support it.

Since we're off the old Mono version, I don't see any reason not to go nuts with C# 6 :grin:

thecodejunkie:
@khellang :+1: let's have a chat about this tomorrow and make a united decision

jchannon:
Surely we just run R# Clean Code ;)

PR #1972, opened 2015-06-27
Performance Optimization

VictorTomaili:


khellang:
I'm sure it would be better to send this PR to https://github.com/grumpydev/TinyIoC :smile:

VictorTomaili:
Heheh yep. You are right. IsClass property has to much proccess time. And we have too much interface.  I optimize it for Nancy. Others libraries may be not use too much interfaces and it's take too much time for process on my code.

PR #1973, opened 2015-06-30
Parameterize `Csrf.GetFormToken`

Crisfole:
In many situations (esp. when using a lot of JSON) it is less convenient to send back CSRF tokens as form fields.

In those cases you may want to send back the plain-text token (not the encrypted cookie) in a cookie or custom Http Header.

This PR adds a (backward compatible) parameter to `ValidateCsrfToken` that allows you to specify how to fetch the plain text CSRF token.

It would be used like so:

```csharp
this.ValidateCsrfToken(getSentToken: (module) => {
    var tokenStr = module.Request.Headers["Csrf-Check"];
    return CsrfApplicationStartup.ObjectSerializer.Deserialize(tokenStr) as CsrfToken;
});
```

### DIFF
--- a/src/Nancy/Security/Csrf.cs
+++ b/src/Nancy/Security/Csrf.cs
@@ -104,8 +104,9 @@
         /// <param name="module">Module object</param>
         /// <param name="validityPeriod">Optional validity period before it times out</param>
         /// <exception cref="CsrfValidationException">If validation fails</exception>
-        public static void ValidateCsrfToken(this INancyModule module, TimeSpan? validityPeriod = null)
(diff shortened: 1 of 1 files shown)


Crisfole:
This is what happens when I edit from inside Github... :) I'm also realizing that it should probably be a `Func<INancyModule, IObjectSerializer, CsrfToken>`...

thecodejunkie:
@Crisfole Do you plan on re-visting this (anytime soon) ? :grin:

khellang:
I'm closing this as we're trying to clear out our PR queue in order to start work on 2.0. Feel free to open again later :smile:

Crisfole:
@thecodejunkie @khellang Sorry :( Lost the job that let me play with Nancy and don't have enough personal time between work (new job), school, and family obligations to hack on Nancy. Let me know if you need specific help...

PR #1981, opened 2015-07-10
Add DNX demo project

horsdal:
The demo project has a `WebListener` and `Kestrel` command and can run a super simple Nancy module on either. 
Right now it only works on `dnx451`, but when the Nancy package compiles on `dnxcore50` this should run there too.


jchannon:
Do you see us having 2 branches going forward so we don't have ifdefs? Just looking at the filesystemprovider changes

horsdal:
No. Those changes compile on both `DNX`'s

jchannon:
Yeah but not on current .net :)

horsdal:
ooh. Doh. Removing that commit.

horsdal:
Better?

jchannon:
Yup :)

PR #1988, opened 2015-07-21
Preventing Bootstrapper assembly scanner from crashing when encounter…

AdamFinley:
…ing incompatible DLLs

Extends exception handling added in #1041

Larger projects can contain resources that aren't all meant to be loaded concurrently.  Scanning every assembly in the executable directory without handling OutOfMemoryException causes initialization issues.  Nancy would fail to start even when Nancy references are present in other, valid DLLs.


khellang:
I'm curious what "initialization issues" this causes? If you get OutOfMemoryExceptions, then there's something seriously wrong and we should look at how to prevent those exceptions, instead of swallowing them. Swallowing exceptions is not something you'd want to do.

khellang:
There is actually (at least) an issue with the way we do assembly loading and scanning today;

 - First we load all assemblies in the application folder in reflection-only context.
 - Based on the references found in the reflection-only assemblies, we load a subset (or all) of them into the execution context.

Since everything is loaded into the same app domain, this takes up quite a bit more memory than needed.
A solution could be to spin up a separate domain, load the assemblies (reflection-only) in that and marshal back a list of assemblies referencing Nancy.

AdamFinley:
I agree with your view on this.  In my case, the directory contains a large number of assemblies that are never meant to be concurrently loaded.  When the `TypeScanner` is loading the full set of *.DLL, resource usage gets a little out of control.  Is there a way to target the assemblies better than just *.DLL ?  Probably not.

On the topic of swallowing `OutOfMemoryException`, there's a good read here: http://stackoverflow.com/questions/2117142/when-is-it-ok-to-catch-an-outofmemoryexception-and-how-to-handle-it

I absolutely agree that assemblies should be loaded in a separate `AppDomain` for the scan since its (required for unloading during runtime).  However - we could still end up loading a giant assembly and getting an `OutOfMemoryException`.  The above post on SO suggests spinning them up in a separate PROCESS.  That would be the only way to ensure the application is viable after post-exception.

khellang:
First pf all, there is no concurrent loading. And what are these "giant assemblies" that aren't meant to be loaded? Why are they there? This sounds like a very special use case. Maybe it's better for you to just add some convention to `AssembliesToScan`?

khellang:
@AdamFinley Did you figure out a solution? :smile:

PR #1990, opened 2015-07-24
Urlencodedcapturegroups

taharrison:
Currently Nancy url-decodes capture groups before identifying where the capture groups begin and end (i.e. splitting on "/")

This leads to odd behavior around encoded "/"s.

**Examples:**

With routes 

1. `"foo/{bar}/baz"`
2. `"foo/{fuzz}"`

* a request to `"foo/bar%2fbaz"` will resolve to route `"foo/{bar}/baz"`, despite the "/" having been encoded (and should therefore not be treated as a path separator).

* a request to `"foo/3-slashes-looks-like-"%2f%2f%2f"` resolves to `"foo/{fuzz*}"` (but not `"foo/{fuzz}"`) with a Fuzz parameter initialised to `"3-slashes-looks-like-"/""`, rather than the expected `"3-slashes-looks-like-"///""`

This merge introduces already passing tests around url decoding in the default route resolver fixture to confirm and preserve existing decoding behavior; then introduces tests that illustrate the incorrect behavior above.

The included fix is to delay url decoding until /after/ the path has been split on the "/" characters.

### DIFF
--- a/src/Nancy.Tests/Unit/Routing/DefaultRouteResolverFixture.cs
+++ b/src/Nancy.Tests/Unit/Routing/DefaultRouteResolverFixture.cs
@@ -72,6 +72,7 @@
             }
         }
 
+
         [Theory]
         [InlineData("/foo/testing/plop", true, "testing")]
         [InlineData("/foo/testing/plop", false, "testing")]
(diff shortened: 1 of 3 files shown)


taharrison:
Very willing to create an issue for this, if required, but since I discovered the issue and fix at the same time I'm assuming its not needed unless told otherwise.

taharrison:
Note - there is a case where incorrectly coded programs built on Nancy may break as a result of this fix.

If I _rely_ on Nancy treating %2F as a path separator, e.g. I request `"/foo%2Fbar" and expect to resolve to "/foo/bar" then I will break.  But this must be preferable to the current state where clients that expect url encoded separators to be ignored in path resolution are broken.

khellang:
Thanks for this! Could this be related to #1280?

taharrison:
I think this pull request is /correct/ but by the looks of the discussion on issue #1280, other devs are planning to change where decoding is done completely.

I think the plan on 1280 is incorrect, as the DRR won't be able to treat "/" correctly.

Will join the discussion there.

Maybe this merge request should be kept on the shelf until then.
 :(

taharrison:
*Sigh* 
@khellang lets close & reject this pull request.
The fix is way more complicated:
https://github.com/NancyFx/Nancy/issues/1280#issuecomment-54761233

khellang:
Yeah, the problem is pretty hairy. Also see #1676 and #1678

PR #2001, opened 2015-08-10
Changing it so that an empty string no longer is converted to a null.…

wtilton:
… Placed an option in the config if this was the actual intent.

### DIFF
--- a/src/Nancy/ModelBinding/BindingConfig.cs
+++ b/src/Nancy/ModelBinding/BindingConfig.cs
@@ -43,5 +43,11 @@
         /// </summary>
         /// <value><see langword="true" /> if the binder is allowed to overwrite non-default values, otherwise <see langword="false" />.</value>
         public bool Overwrite { get; set; }
+
+        /// <summary>
+        /// If the string passed in is empty the binder will not bind it.
+        /// </summary>
+        /// <value><see langword="true" /> if the binder will convert empty strings to null, otherwise <see langword="false" />.</value>
+        public bool ConvertEmptyStringToNull { get; set; }
     }
 }
--- a/src/Nancy/ModelBinding/DefaultBinder.cs
+++ b/src/Nancy/ModelBinding/DefaultBinder.cs
@@ -180,7 +180,7 @@ namespace Nancy.ModelBinding
 
         private bool BindingValueIsValid(string bindingValue, object existingValue, BindingMemberInfo modelProperty, BindingContext bindingContext)
         {
-            return (!String.IsNullOrEmpty(bindingValue) &&
+            return ((!String.IsNullOrEmpty(bindingValue) || (!bindingContext.Configuration.ConvertEmptyStringToNull && bindingValue != null)) &&
                     (IsDefaultValue(existingValue, modelProperty.PropertyType) ||
                      bindingContext.Configuration.Overwrite));
         }
--- a/src/Nancy/ModelBinding/DefaultBinder.cs
+++ b/src/Nancy/ModelBinding/DefaultBinder.cs
@@ -180,9 +180,13 @@ namespace Nancy.ModelBinding
 
         private bool BindingValueIsValid(string bindingValue, object existingValue, BindingMemberInfo modelProperty, BindingContext bindingContext)
         {
-            return ((!String.IsNullOrEmpty(bindingValue) || (!bindingContext.Configuration.ConvertEmptyStringToNull && bindingValue != null)) &&
-                    (IsDefaultValue(existingValue, modelProperty.PropertyType) ||
-                     bindingContext.Configuration.Overwrite));
+            var passedValueContainsSomething = !String.IsNullOrEmpty(bindingValue);
+            var passedValueNotNull = bindingValue != null;
+            var doNotConvertEmptyStringsToNull = !bindingContext.Configuration.ConvertEmptyStringToNull && passedValueNotNull;
+            var passedValueIsTheDefaultValue = IsDefaultValue(existingValue, modelProperty.PropertyType);
+            var overwriteValuesThatDoNotHaveADefaultValue = bindingContext.Configuration.Overwrite;
+
+            return ((passedValueContainsSomething || doNotConvertEmptyStringsToNull) && (passedValueIsTheDefaultValue || overwriteValuesThatDoNotHaveADefaultValue));
         }
 
         /// <summary>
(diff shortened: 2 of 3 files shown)


src/Nancy/ModelBinding/DefaultBinder.cs
```diff
@@ -180,7 +180,7 @@ public object Bind(NancyContext context, Type modelType, object instance, Bindin
 
         private bool BindingValueIsValid(string bindingValue, object existingValue, BindingMemberInfo modelProperty, BindingContext bindingContext)
         {
-            return (!String.IsNullOrEmpty(bindingValue) &&
+            return ((!String.IsNullOrEmpty(bindingValue) || (!bindingContext.Configuration.ConvertEmptyStringToNull && bindingValue != null)) &&
```
khellang: It's probably partly because it's late, but this boolean thingy right here is getting WAAAY too complex for my poor brain to parse. Think you could split it up into a couple of locals that has some good names? :smile:

thecodejunkie: What he said :D

wtilton:
Oh sure, I can do that. I've been wrestling with a server the past few days.

wtilton:
I'm not able to run the test that appears to be failing. The xUnit runner is not working for me and won't install. I have no idea why it is failing (which concerns me) because I don't think that I was anywhere near that code with these changes. I'd also write a new test to test this new functionality but without a runner really hard to do.

khellang:
Mmyeah. I pulled down your code and ran the test in question, and there's definitely something fishy going on with that logic. I tried to revert your changes in the binding logic and that made it pass. I'm actually not really groking what that boolean expression is supposed to convey, even with the variable names. Maybe a solution would be to just return early where you can? That way you'd also avoid doing lots of calculations etc. if you don't have to.

src/Nancy/ModelBinding/BindingConfig.cs
```diff
@@ -43,5 +43,11 @@ public BindingConfig()
         /// </summary>
         /// <value><see langword="true" /> if the binder is allowed to overwrite non-default values, otherwise <see langword="false" />.</value>
         public bool Overwrite { get; set; }
+
+        /// <summary>
+        /// If the string passed in is empty the binder will not bind it.
+        /// </summary>
+        /// <value><see langword="true" /> if the binder will convert empty strings to null, otherwise <see langword="false" />.</value>
+        public bool ConvertEmptyStringToNull { get; set; }
```
thecodejunkie: A better name would perhaps be `IgnoreEmptyStringValues` and update the XML comments accordingly

thecodejunkie:
*IMPORTANT*

Once this has been pulled in, then the [Binding configuration](https://github.com/NancyFx/Nancy/wiki/Model-binding#binding-configuration) section, of the Model Binding documentation, has to be updated to include the new member

src/Nancy/ModelBinding/BindingConfig.cs
wtilton: ConvertEmptyStringToNull was what was used in ASP.NET MVC, it's what I found when I was trying to find a solution to this issue. Not that it is any better but that's why I named it that.

wtilton:
@khellang OK, I can try that. That does sound like it would make it a bit cleaner, good suggestion.

I totally agree that a test, testing the new functionality is pretty important, I just can't get the test runner to install or work on my machine. I updated to Windows 10, maybe that's what's going on. My nUnit tests on my own project still work fine though.

That crazy/awesome that the test is catching something there. I'm fascinated to see how those things are tied together. :)

src/Nancy/ModelBinding/BindingConfig.cs
thecodejunkie: Still prefer `IgnoreEmptyStringValues` because `ConvertEmptyStringToNull` is a bit wrong. It does not convert an empty string to null, it simply ignores setting the values, i.e leaving it at the default which happens to be null for a string =)

wtilton: I can just tell you that when I went searching for the problem the only thing I could think of searching for was ConvertEmptyStringToNull because of that being in the ASP.NET MVC project. It seems to have the same behavior. I agree that your name is probably better suited, and after the change there probably won't be anyone searching for it anymore maybe.

I just wonder how no one ran into this before?

wtilton:
@khellang Not sure how to rework it. A bunch of if statements did not look better. Maybe the original author can step in here, if they exist?

khellang:
@wtilton I'll see if I can take a look at it tonight :smile:

wtilton:
I think I might know why the test is failing. If you have an object but the data being passed in does not contain the property that the object has.

psuedo code:

class Person {
    prop HiThere
}

request header: <completely blank>

class MyModule {
    var boundObject = this.Bind<Person>(); // ERROR
}

thecodejunkie:
@wtilton @khellang Nope. Having members on your model, that _does not_ have binding data specified in the request (forms, body, query string, parameters) does not cause a failure in the model binding. I am sure of this and I also just verified it locally that we'd not broken any in a recent pull. However if you can reproduce the behavior and share it with us, then I'd be happy to take a look.

That said, I don't think this is the cause of your failing tests. 

Going to move this to the 1.4 milestone because we want to bake 1.3 asap

wtilton:
@thecodejunkie No, I'm saying with my change it is happening. The error cropped up in my current project. It doesn't happen with the unchanged code, so it was my guess that this might be why the test is failing.

thecodejunkie:
@wtilton if that's the case then you've introduced a breaking change and a blocker which is going to have to be resolved before we consider pulling this PR in

thecodejunkie:
@wtilton any updates on this? Thanks

thecodejunkie:
Hi,

After long consideration, we've decided not to take this pull-request. I am sorry. The main reason is just a gut feeling, there's just something that doesn't quite seem to sit right. Also, as it stands now, we couldn't pull it anyway.

Should we find ourselves wanting to change this behaviour then we might return to this pull-request and see what we can make out of it.

Thank you for taking the time to submit it!

wtilton:
Bummer. I guess I can just keep pulling updates into my fork, but... kind of stinks.

I honestly have no idea how people aren't running into this. There is a pretty big difference between null and an empty string which was really negatively affecting what I'm working with. Not sure how everyone is working around this issue.

I did fix the issue of the exception when the parameter wasn't specified but I don't really see the point of submitting if no one can help me get the tests working. The one that was breaking ran fine, now there's something else. This tell me that this issue runs pretty deep so would probably require someone more experienced than me with the code also taking a look at this.

I'll try running xUnit on a different machine to see if I can get this thing into a future release. Our project won't work without it.

Issue #1999, opened 2015-08-10
Create new unified configuration API


thecodejunkie:
_THIS WILL BE A NANCY v2.x FEATURE ONLY DUE TO THE BREAKING CHANGES_

Overtime we have gotten a bit of fragmentation when it comes to configuring various parts of Nancy. We need to re-evaluate this and create a unified approach. I have created a [pull-request](
https://github.com/NancyFx/Nancy/pull/2003) with a design that we've talked about in the team.

The base for the new approach is a new method, on the bootstrapper, called [Configure()](https://github.com/NancyFx/Nancy/blob/nancy-configuration-feature/src/Nancy/Bootstrapper/NancyBootstrapperBase.cs#L316)

```c#
public virtual void Configure(INancyEnvironment environment)
{
}
```

The [INancyEnvironment](https://github.com/NancyFx/Nancy/blob/nancy-configuration-feature/src/Nancy/Configuration/INancyEnvironment.cs) instance, that is passed in, is essentially an `IReadOnlyDictionary<string, object>` with a couple of more methods added. In short it's a property bag, more or less. You can take a dependency on this interface wherever you want to read configuration values.

To provide a configuration API you simply create extension methods on `INancyEnvironment` which sticks a config (your own type, tailored to your needs), in environment.

Here is what it could look (just sample code) like when configuring Nancy diagnostics, using the new approach

```c#
public virtual void Configure(INancyEnvironment environment)
{
    environment.Diagnostics("password", "the/path/to/use");
}
```

The `Diagnostics` extension method could manufacture an instance of a class that looks something like this

```c#
public class DiagnosticsConfiguration
{
    public DiagnosticsConfiguration(string password, string path)
    {
        this.Password = password;
        this.Path = path;
    }

    public string Password { get; private set; }

    public string Path { get; private set; }
}
```

A very basic, immutable, object that contains the configuration value. To get the value back out you take a dependency on `INancyEnvironment` and use one of the `GetValue` extension methods.

```c#
var config = environment.GetValue<DiagnosticsConfiguration>();

// Do something with config.Password and config.Path
```

There are a couple of more bits such as
* [INancyEnvironmentFactory](https://github.com/NancyFx/Nancy/blob/nancy-configuration-feature/src/Nancy/Configuration/INancyEnvironmentFactory.cs) - Creates an `INancyEnvironment` instance
[...]

PR #2003, opened 2015-08-11
Implemented new configuration API

thecodejunkie:
See #1999 for more details on the contents of this post


jchannon:
:shipit:

src/Nancy.Tests/Unit/Bootstrapper/NancyBootstrapperBaseFixture.cs
```diff
@@ -465,6 +475,14 @@ public void Dispose()
         public bool Disposed { get; private set; }
     }
 
+    internal class FakeEnvironmentalConfigurator : INancyEnvironmentConfigurator
```
horsdal: `Environmental` should just be `Environment`

src/Nancy.Tests/Unit/Configuration/DefaultNancyEnvironmentFixture.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        [Theory]
+        [InlineData(2)]
+        [InlineData(4)]
+        [InlineData(6)]
+        public void Should_add_value_to_environment_when_invoking_addvalue(int numberOfElementsToAdd)
+        {
+            // Given, When
+            var environment = CreateEnvironment(numberOfElementsToAdd);
+            IReadOnlyDictionary<string, object> setter = environment;
```
horsdal: No reason to have the `setter` variable` AFAICT

src/Nancy.Tests/Unit/Configuration/DefaultNancyEnvironmentFixture.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            // When
+            var result = environment.ContainsKey(keyToLookFor);
+
+            // Then
+            result.ShouldEqual(expectedResult);
+        }
+
+        [Fact]
+        public void Should_return_true_if_key_exists_when_invoking_trygetvalue()
```
horsdal: This test and `Should_return_false_if_key_does_not_exist_when_invoking_trygetvalue()` could be unified with some `InilineData` like above.

horsdal:
How about `INacyEnvironment` implements 2 interfaces. A mutable one and an immutable one. The environment is passed as the mutable interface into `.Configure`, but only registered in the container as the immutable one. Defaults could then be overriden in `Configure` but module etc. taking a dependency on the environment could only read.

thecodejunkie:
Thanks @horsdal updated the code according to your review comments!

phillip-haydon:
:+1:

thecodejunkie:
Rebased against latest `v2` branch. We need to pull this before we merge `v2` into `master` .. same with #2012 after I've rebased it

PR #2004, opened 2015-08-11
Perform case-insensitive lookups in EmbeddedFileResource

anton-gogolev:
Perform case-insensitive lookups in EmbeddedFileResource

Previously, `EmbeddedFileResource` was overly picky with regards to embedded resource name casing. Particularly, this was causing problems when `AddDirectory()` on `EmbeddedStaticContentConventionBuilder` was given an incorrectly-capitalized `contentPath` argument, i.e.:

    _.AddDirectory("content", GetType().Assembly, contentPath: "content"));

Provided that embedded resources were under "Foo.Content", all requests to `/content/...` would fail.

With this PR, `EmbeddedFileResource` becomes much less picky about casing, which it has been all along nevertheless, except for a couple places.

### DIFF
--- a/src/Nancy/Responses/EmbeddedFileResponse.cs
+++ b/src/Nancy/Responses/EmbeddedFileResponse.cs
@@ -48,12 +48,10 @@
         {
             var resourceName = assembly
                 .GetManifestResourceNames()
-                .Where(x => GetFileNameFromResourceName(resourcePath, x).Equals(name, StringComparison.OrdinalIgnoreCase))
-                .Select(x => GetFileNameFromResourceName(resourcePath, x))
-                .FirstOrDefault();
+                .FirstOrDefault(x => GetFileNameFromResourceName(resourcePath, x).Equals(name, StringComparison.OrdinalIgnoreCase));
 
-            resourceName =
-                string.Concat(resourcePath, ".", resourceName);
+            if (resourceName == null)
+                return null;
 
             return assembly.GetManifestResourceStream(resourceName);
         }
--- a/src/Nancy/Responses/EmbeddedFileResponse.cs
+++ b/src/Nancy/Responses/EmbeddedFileResponse.cs
@@ -6,6 +6,7 @@
     using System.Reflection;
     using System.Security.Cryptography;
     using System.Text;
+    using System.Text.RegularExpressions;
 
     public class EmbeddedFileResponse : Response
     {
@@ -59,7 +60,7 @@
 
         private static string GetFileNameFromResourceName(string resourcePath, string resourceName)
         {
-            return resourceName.Replace(resourcePath, string.Empty).Substring(1);
+            return Regex.Replace(resourceName, resourcePath, string.Empty, RegexOptions.IgnoreCase).Substring(1);
         }
 
         private static string GenerateETag(Stream stream)
--- a/src/Nancy.Embedded.Tests/Unit/EmbeddedStaticContentConventionBuilderFixture.cs
+++ b/src/Nancy.Embedded.Tests/Unit/EmbeddedStaticContentConventionBuilderFixture.cs
@@ -58,7 +58,7 @@
             result.ShouldEqual("Embedded Text");
         }
 
-        private static string GetEmbeddedStaticContent(string virtualDirectory, string requestedFilename, string root = null)
+        private static string GetEmbeddedStaticContent(string virtualDirectory, string requestedFilename, string contentPath = null)
         {
             var resource =
                 string.Format("/{0}/{1}", virtualDirectory, requestedFilename);
--- a/src/Nancy.Embedded.Tests/Unit/EmbeddedStaticContentConventionBuilderFixture.cs
+++ b/src/Nancy.Embedded.Tests/Unit/EmbeddedStaticContentConventionBuilderFixture.cs
@@ -58,6 +58,17 @@
             result.ShouldEqual("Embedded Text");
         }
 
+        [Fact]
+        public void Should_retrieve_static_content_ignoring_casing()
+        {
+            // Given
+            // When
+            var result = GetEmbeddedStaticContent("Foo", "Subfolder/embedded2.txt", "resources");
+
+            // Then
+            result.ShouldEqual("Embedded2 Text");
+        }
+
         private static string GetEmbeddedStaticContent(string virtualDirectory, string requestedFilename, string contentPath = null)
         {
             var resource =
--- a/src/Nancy.Embedded.Tests/Unit/EmbeddedStaticContentConventionBuilderFixture.cs
+++ b/src/Nancy.Embedded.Tests/Unit/EmbeddedStaticContentConventionBuilderFixture.cs
@@ -77,7 +77,7 @@
                 Assembly.GetExecutingAssembly();
 
             var resolver =
-                EmbeddedStaticContentConventionBuilder.AddDirectory(virtualDirectory, assembly, "Resources");
+                EmbeddedStaticContentConventionBuilder.AddDirectory(virtualDirectory, assembly, contentPath ?? "Resources");
 
             var response =
                 resolver.Invoke(context, null) as EmbeddedFileResponse;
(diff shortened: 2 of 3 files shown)


src/Nancy/Responses/EmbeddedFileResponse.cs
```diff
@@ -47,19 +48,17 @@ private Stream GetResourceContent(Assembly assembly, string resourcePath, string
         {
             var resourceName = assembly
                 .GetManifestResourceNames()
-                .Where(x => GetFileNameFromResourceName(resourcePath, x).Equals(name, StringComparison.OrdinalIgnoreCase))
-                .Select(x => GetFileNameFromResourceName(resourcePath, x))
-                .FirstOrDefault();
+                .FirstOrDefault(x => GetFileNameFromResourceName(resourcePath, x).Equals(name, StringComparison.OrdinalIgnoreCase));
 
-            resourceName =
-                string.Concat(resourcePath, ".", resourceName);
```
jchannon: Don't you still need the `string.Concat`?

anton-gogolev: @jchannon Nope.

`Assembly.GetManifestResourceStream()` requires [case-sensitive name](https://msdn.microsoft.com/en-us/library/xc4235zt(v=vs.110).aspx) of the manifest resource, and the `FirstOrDefault()` call on line 51 returns just that: full verbatim name of the matching embedded resource.

jchannon: Ah yeah I missed the `GetFileNameFromResourceName ` change.

jchannon:
LGTM, other @NancyFx/most-valued-minions want to approve?

damianh:
Something about this is not sitting well with me. Maybe because unix filesystems are case sensitive, as are URLs. Not sure hmm.

src/Nancy/Responses/EmbeddedFileResponse.cs
```diff
@@ -47,19 +48,17 @@ private Stream GetResourceContent(Assembly assembly, string resourcePath, string
         {
             var resourceName = assembly
                 .GetManifestResourceNames()
-                .Where(x => GetFileNameFromResourceName(resourcePath, x).Equals(name, StringComparison.OrdinalIgnoreCase))
-                .Select(x => GetFileNameFromResourceName(resourcePath, x))
-                .FirstOrDefault();
```
damianh: Any per-request request performance concerns with this select?

jchannon: The `Select` has been removed?

src/Nancy/Responses/EmbeddedFileResponse.cs
```diff
@@ -47,19 +48,17 @@ private Stream GetResourceContent(Assembly assembly, string resourcePath, string
         {
             var resourceName = assembly
                 .GetManifestResourceNames()
-                .Where(x => GetFileNameFromResourceName(resourcePath, x).Equals(name, StringComparison.OrdinalIgnoreCase))
-                .Select(x => GetFileNameFromResourceName(resourcePath, x))
-                .FirstOrDefault();
+                .FirstOrDefault(x => GetFileNameFromResourceName(resourcePath, x).Equals(name, StringComparison.OrdinalIgnoreCase));
```
damianh: There is still a select here (I know this is not specific to the PR, just an observation)

src/Nancy/Responses/EmbeddedFileResponse.cs
damianh: eh yup :) but it's still there...

src/Nancy/Responses/EmbeddedFileResponse.cs
```diff
@@ -47,19 +48,17 @@ private Stream GetResourceContent(Assembly assembly, string resourcePath, string
...
+            if (resourceName == null)
+                return null;
 
             return assembly.GetManifestResourceStream(resourceName);
         }
 
         private static string GetFileNameFromResourceName(string resourcePath, string resourceName)
         {
-            return resourceName.Replace(resourcePath, string.Empty).Substring(1);
+            return Regex.Replace(resourceName, resourcePath, string.Empty, RegexOptions.IgnoreCase).Substring(1);
```
damianh: Is this worth a performance review?

src/Nancy/Responses/EmbeddedFileResponse.cs
anton-gogolev: @damianh Sorry, I don't quite get what are your concerns here?

src/Nancy/Responses/EmbeddedFileResponse.cs
anton-gogolev: @damianh I highly doubt that. Serving static content from embedded resources is not fast by itself, so a few additional milliseconds will not make a difference here. Otherwise, we'll have to invent our very own

    public static string Replace(this string source, string oldValue, string newValue, 
        StringComparison comparisonType)

src/Nancy/Responses/EmbeddedFileResponse.cs
damianh: Per-request performance concerns. 

Just a smell that needs to be thought about....

anton-gogolev:
@damianh Basically, all the Embedded subsystem is case-insensitive now: string comparisons and other operations are performed with `...IgnoreCase`. Plus, this PR is not as much about case sensitivity in URLs but rather about Nancy being able to locate static content without being too strict on embedded resource names.

anton-gogolev:
@jchannon @damianh So guys, what are your thoughts on this?

thecodejunkie:
With regards to the case-(in)sensitive yay or nay discussion. The feature could respect the [StaticConfiguration.CaseSensitive](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/StaticConfiguration.cs#L49) settings. There are several features that use the value of this to decide if it should ignore case or not

damianh:
Will there ever be a situation where you want Case sensitivity in one
place, but not another?

In anycase, I definitely believe this should be opt-in.

anton-gogolev:
@thecodejunkie I don't quite get where it would be appropriate to use `CaseSensitive` property. [EmbeddedStaticContentConventionBuilder](https://github.com/anton-gogolev/Nancy/blob/case-insensitive-embedded-file-response/src/Nancy.Embedded/Conventions/EmbeddedStaticContentConventionBuilder.cs) is already case-insensitive, which means that the essense of this PR (ad37b8014c05481c4d07309f2e7f89fbcd55ab3f) just reinforces preexising behavior.

src/Nancy/Responses/EmbeddedFileResponse.cs
```diff
@@ -47,19 +48,17 @@ private Stream GetResourceContent(Assembly assembly, string resourcePath, string
...
             var resourceName = assembly
                 .GetManifestResourceNames()
-                .Where(x => GetFileNameFromResourceName(resourcePath, x).Equals(name, StringComparison.OrdinalIgnoreCase))
-                .Select(x => GetFileNameFromResourceName(resourcePath, x))
-                .FirstOrDefault();
+                .FirstOrDefault(x => GetFileNameFromResourceName(resourcePath, x).Equals(name, StringComparison.OrdinalIgnoreCase));
 
-            resourceName =
-                string.Concat(resourcePath, ".", resourceName);
+            if (resourceName == null)
```
thecodejunkie: We should replace this with

```c#
return (resourceName != null) ?
    assembly.GetManifestResourceStream(resourceName) :
    null;
```

thecodejunkie:
@anton-gogolev The way I've understood your explanation is that, currently, the `EmbeddedFileResource` is case-sensitive - causing issues if you are not aware of this? Your change made it case-insensitive. My suggestion was to support both the existing and new behavior by having it use the value of [StaticConfiguration.CaseSensitive](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/StaticConfiguration.cs#L49) to determine in which "mode" it should be operating.

Or am I misunderstanding?

thecodejunkie:
@anton-gogolev thoughts?

Pushing this to 1.4 since we want to bake 1.3 asap

thecodejunkie:
@anton-gogolev Giving his one final poke. See https://github.com/NancyFx/Nancy/pull/2004#issuecomment-133566471

Thanks

horsdal:
I agree with @damianh. This doesn't feel right. Even with a config setting, it would still feel like a hack to me.

Issue #1837, opened 2015-02-27
Nancy.Testing.BrowserResponseBodyWrapper does not support gzipped responses


rasmus-beck:
I was trying to write a test for my REST API tonight, that tested a gzipped response. But using the Nancy test framework isn't possible, as the `BrowserResponseBodyWrapper` sees the response stream as being closed.

This is not a very detailed issue description, but I will write a failing test for this sometime in the next couple of days.

horsdal:
@rasmus-beck did you get around to that failing test?

rasmus-beck:
Sorry got sidetracked. We decided to let the load balancer do the GZipping, and I lost all focus on this. Will see if I can get around to this soon.

horsdal:
:cool:

yannisgu:
Heres a test for reproduction: https://github.com/NancyFx/Nancy/compare/master...yannisgu:issue1837
Didn't find a solution yet...

rasmus-beck:
@yannisgu Thanks!

xt0rted:
When creating the `GZipStream` ([DeflateStream](https://msdn.microsoft.com/en-us/library/system.io.compression.deflatestream(v=vs.110).aspx) as well) you need to pass`true` as the [third parameter](https://msdn.microsoft.com/en-us/library/27ck2z1y(v=vs.110).aspx#parameters). The remarks section has more information as to why.

Here's a passing version of the above test.
```c#
[Fact]
public void Should_support_gzipped_response()
{
    // Given
    bootstrapper.AfterRequest.AddItemToEndOfPipeline(ctx =>
    {
        ctx.Response.Headers["Content-Encoding"] = "gzip";
        var contents = ctx.Response.Contents;
        ctx.Response.Contents = responseStream =>
        {
            using (var gzip = new GZipStream(responseStream, CompressionMode.Compress, leaveOpen: true))
            {
                contents.Invoke(gzip);
            }
        };
    });

    // When
    var result = browser.Get("/");

    // Then
    using (var responseStream = new MemoryStream(result.Body.ToArray()))
    using (var zipStream = new GZipStream(responseStream, CompressionMode.Decompress))
    using (var zipStreamReader = new StreamReader(zipStream))
    {
        zipStreamReader.ReadToEnd().ShouldEqual("Hello World");
    }
}
```

thecodejunkie:
Related to #2011 

https://github.com/NancyFx/Nancy/pull/2011#issuecomment-132979264

PR #2011, opened 2015-08-14
Is any real reason for contentsStream.Position=0?

AIexandr:
contentsStream.Position = 0;
- fails when response's stream is closed already, i.e. something like:

```csharp
var response = new Response();
response.ContentType = file.ContentType;
response.Contents = s =>
{
    using (var bw = new System.IO.BinaryWriter(s))
    using (var inputStream = new System.IO.FileStream(file.PhisicalFilePath, System.IO.FileMode.Open))
    using (var br = new System.IO.BinaryReader(inputStream))
    {
        var bufferMaxLength = 1048576; // 1Mb
        var bufferLength = file.Size;
        if (bufferLength > bufferMaxLength) bufferLength = bufferMaxLength;

        while (br.BaseStream.Position < br.BaseStream.Length)
        {
            var cnt = br.BaseStream.Length - br.BaseStream.Position;
            if (cnt > bufferLength) cnt = bufferLength;
            if (cnt > 0)
            {
                var buffer = br.ReadBytes((int)cnt);
                bw.Write(buffer);
             }
             else
             {
                 break;
              }
              bw.Flush();
          }
      }
};
```

### DIFF
--- a/src/Nancy.Testing/BrowserResponseBodyWrapper.cs
+++ b/src/Nancy.Testing/BrowserResponseBodyWrapper.cs
@@ -30,7 +30,7 @@ namespace Nancy.Testing
         {
             var contentsStream = new MemoryStream();
             response.Contents.Invoke(contentsStream);
-            contentsStream.Position = 0;
+            //contentsStream.Position = 0;
             return contentsStream;
         }
 


thecodejunkie:
The problem is not that the position is set back to 0, the real problem here is that the `BinaryWriter` is closing the underlaying stream when you call `Dispose`. You are, in essence, closing a stream you do not own, which is bad practice. As of .NET 4.5 they added an overload with a `leaveOpen` parameter that you can use to prevent this behaviour  https://msdn.microsoft.com/en-us/library/gg712841.aspx

For pre-4.5 you simply avoid calling `Dispose` and just call `Flush`.

This is related to #1837 

Thanks

AIexandr:
Thanks for the explanation!
Yes, it is a bad practice to close the stream of course. My problem was due to I fogot to call bw.Flush(); (in my project, not in this example). My bad.

PR #2020, opened 2015-08-20
Fix TextResponse encoding

### LINKED ISSUE #2018: special characters response issue.
var str= "✈️ From Osaka to Okinawa 🎌";
return new TextResponse(str,"text/html");

result
?? From Osaka to Okinawa ??

khellang:
Fixes #2018 

### Breaking Change

This is a breaking change because it adds `; charset={encoding}` to the `Content-Type` header of `TextResponse`.
It also changes `Response`'s implicit operator from `string` to use `TextResponse`, which in turn changes the `Content-Type` from `text/html` to `text/plain`.

This means that if you have tests currently asserting on, or anything else relying on the format of the `ContentType` property of a `TextResponse`, it might break.

### Fix

The fix is trivial; change `text/html` to `text/plain; charset=utf-8` (or if you use other encodings, use it's `WebName` property, i.e. `Encoding.Unicode.WebName`, or simply `utf-16`).

### DIFF
--- a/src/Nancy/Responses/TextResponse.cs
+++ b/src/Nancy/Responses/TextResponse.cs
@@ -10,29 +10,36 @@
     /// </summary>
     public class TextResponse : Response
     {
+        private const string TextPlainContentType = "text/plain";
+
         /// <summary>
         /// Creates a new instance of the TextResponse class
         /// </summary>
         /// <param name="contents">Text content - defaults to empty if null</param>
         /// <param name="contentType">Content Type - defaults to text/plain</param>
         /// <param name="encoding">String encoding - UTF8 if null</param>
-        public TextResponse(string contents, string contentType = "text/plain", Encoding encoding = null)
+        public TextResponse(string contents, string contentType = null, Encoding encoding = null)
         {
             if (encoding == null)
             {
                 encoding = Encoding.UTF8;
             }
 
-            this.ContentType = contentType;
+            if (string.IsNullOrEmpty(contentType))
+            {
+                contentType = TextPlainContentType;
+            }
+
+            this.ContentType = GetContentType(contentType, encoding);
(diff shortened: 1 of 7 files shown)


src/Nancy/Responses/TextResponse.cs
```diff
@@ -15,23 +15,28 @@ public class TextResponse : Response
...
+                contentType = "text/plain";
+            }
+
+            this.ContentType = GetContentType(contentType, encoding);
             this.StatusCode = HttpStatusCode.OK;
 
             if (contents != null)
             {
                 this.Contents = stream =>
-                                {
```
thecodejunkie: Bonus points for fixing this ;)

zephyrweb:
If this is a "breaking change", shouldn't it be moved to 2.0?

thecodejunkie:
@zephyrweb Indeed. Thank you :heart:

src/Nancy/Responses/TextResponse.cs
```diff
@@ -10,29 +10,36 @@
...
         /// </summary>
         /// <param name="contents">Text content - defaults to empty if null</param>
         /// <param name="contentType">Content Type - defaults to text/plain</param>
         /// <param name="encoding">String encoding - UTF8 if null</param>
-        public TextResponse(string contents, string contentType = "text/plain", Encoding encoding = null)
+        public TextResponse(string contents, string contentType = null, Encoding encoding = null)
         {
             if (encoding == null)
             {
                 encoding = Encoding.UTF8;
```
thecodejunkie: I was thinking that perhaps we should add an `Encoding` settings to the new config API (probably along with stuff from static configuration, such as case sensitivity) that it would use unless an encoding was explicitly declared. This pull-request can still go in, but we should add it to the task list in #2000

PR #2023, opened 2015-08-20
Preventing user code from closing the response stream while testing

thecodejunkie:
Wraps the underlaying `MemoryStream`, in `BrowserResponseBodyWrapper`, in an `UnclosableStreamWrapper` to prevent users from closing the underlaying stream by mistake (hint: you should never close a stream you do not own) :-D

Related to #1837 and #2011

### DIFF
--- a/src/Nancy.Testing/BrowserResponseBodyWrapper.cs
+++ b/src/Nancy.Testing/BrowserResponseBodyWrapper.cs
@@ -3,6 +3,7 @@ namespace Nancy.Testing
     using System.Collections;
     using System.Collections.Generic;
     using System.IO;
+    using Nancy.IO;
 
     /// <summary>
     /// Wrapper for the HTTP response body that is used by the <see cref="BrowserResponse"/> class.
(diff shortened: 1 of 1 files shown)


src/Nancy.Testing/BrowserResponseBodyWrapper.cs
```diff
@@ -13,24 +14,35 @@ public class BrowserResponseBodyWrapper : IEnumerable<byte>
...
+        /// Gets the content type of the wrapped response.
+        /// </summary>
+        /// <returns>A string containing the content type.</returns>
+        public string ContentType
         {
             get { return this.contentType; }
         }
 
-        private static MemoryStream GetContentStream(Response response)
+        private static UnclosableStreamWrapper GetContentStream(Response response)
```
khellang: Why change this return type? Couldn't you just do

```csharp
var contentsStream = new MemoryStream();

response.Contents.Invoke(new UnclosableStreamWrapper(contentsStream));
contentsStream.Position = 0;

return contentsStream;
```

And avoid that nasty casting in the ctor?

thecodejunkie: Urg.. yes.. for some reason I got it in my head that it was being used elsewhere as well.. will fix

PR #2026, opened 2015-08-21
Added support to control the use of IApplicationStartup and IRequestStartup implementations from Browser based tests

### LINKED ISSUE #1793: Nancy Testing - prevent autoregistration of IApplicationStartup and IRequestStartup
I have a NancyFx app which uses the default bootstrapper. This automatically discovers and registers at startup all classes which implement `IApplicationStartup` interface and `IRequestStartup`, respectively.

For unit testing my modules I am using Nancy Testing with the `ConfigurableBootstrapper` which uses the same conventions as the "production" bootstraper thus making the same automatic discovery and registrations of the above classes.

How can I prevent automatic registration of `IApplicationStartup` and `IRequestStartup` assignable classes in unit testing scenarios? For example, I have a NewRelic app startup logic which I don't want to use in my unit tests.

thecodejunkie:
We will always register the ones bundled with Nancy, but from tests you can not control how user-defined implementations should be registered (automatic, never, user-defined implementations)

New methods on the ´ConfigurableBootstrapperConfigurator`

```c#
ApplicationStartupTask<T>() where T : IApplicationStartup
ApplicationStartupTasks(params Type[] types)

RequestStartupTask<T>() where T : IApplicationStartup
RequestStartupTasks(params Type[] types)

DisableAutoApplicationStartupRegistration();
DisableAutoRequestStartupRegistration();
```

Resolves #1793

### DIFF
--- a/src/Nancy.Testing.Tests/BrowserFixture.cs
+++ b/src/Nancy.Testing.Tests/BrowserFixture.cs
@@ -6,7 +6,6 @@ namespace Nancy.Testing.Tests
     using System.Security.Cryptography.X509Certificates;
     using System.Text;
     using System.Linq;
-
     using Nancy.Extensions;
     using Nancy.Tests;
     using Nancy.Helpers;
@@ -14,6 +13,7 @@ namespace Nancy.Testing.Tests
     using Xunit;
     using FakeItEasy;
     using Nancy.Authentication.Forms;
+    using Nancy.Bootstrapper;
 
     public class BrowserFixture
     {
@@ -26,7 +26,7 @@ namespace Nancy.Testing.Tests
 
             CookieBasedSessions.Enable(bootstrapper);
 
-            browser = new Browser(bootstrapper);
+            this.browser = new Browser(bootstrapper);
         }
 
         [Fact]
@@ -538,7 +538,7 @@ namespace Nancy.Testing.Tests
 
             // Then
             header.ShouldEqual(expectedHeaderValue);
-        }
+        }     
 
         public class EchoModel
         {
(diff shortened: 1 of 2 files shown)


src/Nancy.Testing.Tests/BrowserFixture.cs
```diff
@@ -538,7 +538,7 @@ public void Should_override_default_user_agent_when_explicitly_defined()
 
             // Then
             header.ShouldEqual(expectedHeaderValue);
-        }
+        }
```
khellang: :fire:

src/Nancy.Testing/ConfigurableBootstrapper.cs
```diff
@@ -187,7 +202,7 @@ private static void LoadReferencesForAssemblyUnderTest(string testAssemblyName)
 
         private static string GetSafePathExtension(string name)
         {
-            return Path.GetExtension(name) ?? String.Empty;
+            return Path.GetExtension(name) ?? string.Empty;
```
khellang: :100: bonus points :+1:

src/Nancy.Testing.Tests/BrowserFixture.cs
thecodejunkie: Not even going to give me a chance to remove the whitespace?! :)

khellang: Ah, the rest was nice, so why bother :stuck_out_tongue:

thecodejunkie: https://en.wikipedia.org/wiki/Broken_windows_theory :smile:

khellang: https://visualstudiogallery.msdn.microsoft.com/a204e29b-1778-4dae-affd-209bea658a59 :wink:

thecodejunkie: I am actually running that, but I was not on the machine I used to send the PR :P

PR #2039, opened 2015-08-28
Url acl prefix fix

### LINKED ISSUE #2038: UrlReservations.CreateAutomatically fails if port is 80
If uri port is 80 Nancy tries to execute the following command:
`netsh http add urlacl url=http://+/ user=Everyone`
which produces:
Url reservation add failed, Error: 87
The parameter is incorrect.
netsh requires the port number. instead of just using `uri.ToString()` which omits the port number if port is 80 one should do something like this: `uri.Scheme + Uri.SchemeDelimiter + uri.Host + ":" + uri.Port + uri.PathAndQuery`

khellang:
Fixes #2038

### DIFF
--- a/src/Nancy.Hosting.Self.Tests/NancySelfHostFixture.cs
+++ b/src/Nancy.Hosting.Self.Tests/NancySelfHostFixture.cs
@@ -3,6 +3,7 @@ namespace Nancy.Hosting.Self.Tests
 {
     using System;
     using System.IO;
+    using System.Linq;
     using System.Net;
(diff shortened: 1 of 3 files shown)


phillip-haydon:
@khellang can you stop breaking Team City...

khellang:
@phillip-haydon Totally unrelated. Rebuilt and all is :ok:

phillip-haydon:
Pfft I know you're trying to sabotage the Nancy project!

Issue #2043, opened 2015-08-29
How to build instructions reference incorrect version of Albacore


jonathanfoster:
The how to build instructions (`/src/how_to_build.txt`) recommend installing the latest version of Albacore, but Rake fails when using the latest version with the error message `NameError: uninitialized constant Albacore::Task`. The [Travis build script](https://github.com/NancyFx/Nancy/blob/5dc9e2a179c5f3a50f6caf238d0341cdd2dd1474/.travis.yml#L12) forces version 1.0.0.rc2. The build instructions should be updated to reference the correct version.

jonathanfoster:
[PR 2045](https://github.com/NancyFx/Nancy/pull/2045) resolves this issue.

PR #2045, opened 2015-08-30
Update how to build instructions with missing prerequisites

jonathanfoster:
The rake build is not compatible with the latest version of albacore, it's only compatible with v1.0.0.rc2. If you don't have Visual Studio 2010 installed then you'll need to install the Visual Studio 2010 shell in order to build projects that reference Microsoft.WebApplication.targets. The Windows SDK is required because Microsoft.Common.targets reference AL.exe using the SdkToolsPath. These prerequisites were confirmed on a bare Windows 8.1 install with Visual Studio 2015. This PR resolves [issue 2043](https://github.com/NancyFx/Nancy/issues/2043).

### DIFF
--- a/how_to_build.txt
+++ b/how_to_build.txt
@@ -6,19 +6,17 @@ How to build Nancy
 Prerequisites
 -------------
 
-1. Download and install Ruby 1.8.7+ from http://www.ruby-lang.org/en/downloads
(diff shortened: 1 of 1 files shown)


jchannon:
What does the SDK bring?

jchannon:
Thanks

PR #2057, opened 2015-09-14
Use Enumerable as intended

jongleur1983:
StringExtensions.GetParameterDetails() returns `Enumerable<ParameterSegmentInformation>`.
If there is no parameter, it threw a `FormatException`.
IMHO there is no reason not to return an empty enumerable instead.

# Drawback of the old solution
With the exception approach using this function required to check if there is a parameter at all, which has been done by `IsParametrized` in the only "root" usage of the function (DefaultRoutePatternMatcher.cs line 107); on the other hand, `ParameterizeSegment`(DefaultRouteMatcher line 128) does not require the segment to contain at least one parameter, to function, so that check wouldn't even be required.

# Benefit of this change
The additional check is still possible, but not required to prevent throwing an exception.

(side note: removed Select-self statement (line 36)

### DIFF
--- a/src/Nancy/Extensions/StringExtensions.cs
+++ b/src/Nancy/Extensions/StringExtensions.cs
@@ -33,15 +33,9 @@ namespace Nancy.Extensions
 
             var nameMatch = matches
                 .Cast<Match>()
-                .Select(x => x)
                 .ToList();
 
-            if (nameMatch.Any())
-            {
-                return nameMatch.Select(x => new ParameterSegmentInformation(x.Groups["name"].Value, x.Groups["default"].Value, x.Groups["default"].Success));
-            }
-
-            throw new FormatException("The segment did not contain any parameters.");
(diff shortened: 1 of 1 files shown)


06b:
If this is a breaking change, shouldn't it be added to the 2.0 milestone instead of 1.3?

grumpydev:
Is this a breaking change? It's an extension method only used inside our default implementation of routing isn't it?

thecodejunkie:
Yes and No. It's a public extension method, so technically someone could use it for whatever reason. I could go two ways on this ?

grumpydev:
I guess, but it's pretty tenuous imo, anyone using this would be writing their own routing, I'm not sure that's an "end user contract" that we need to be concerned about for a breaking change - semver for me is more about whether you break your public, documented API contracts, not just whether a type that happens to be public has changed - but I know this is kinda woolly :)

thecodejunkie:
Yeah let's fly with that :)

thecodejunkie:
All CI-checks are :white_check_mark: 
Status :shipit:

PR #2059, opened 2015-09-15
Update ResourceBasedTextResource.cs

duszekmestre:
When you have EndsWith then you can have problems with resources like that:
Orders
ExternalOrders.
End with this case you have error!


khellang:
Thanks for this! :heart:

This particular class is not well tested, and I'm afraid this might break someone. Would you be able to add a couple of tests for the indexer?
You can add them to [`ResourceBasedTextResourceFixture`](https://github.com/NancyFx/Nancy/blob/fec9d1e94c4486d6fb5bd20c4991baf10cfdc664/src/Nancy.Tests/Unit/Localization/ResourceBasedTextResourceFixture.cs) :smile:

thecodejunkie:
Been a while since I looked at this, but won't this actually break stuff?

If I recall correctly, the reason we used `EndsWith` with because resources gets "namespaces"

So you could have

`Foo.Name` and `Bar.Name` and you ask for `Name` then we will find both and throw an exception telling you to provide a more specific key.. so you'd have to request it using either `Foo.Name` or `Bar.Name` to get a unique match. The more specific key is **only** required if you run into this issue with your resources.

However, by turning it into an `Equals` check, then you're **always* going to have to provide the full key. I'e even if you only had a `Foo.Name` you could no longer just ask for it using `Name`

duszekmestre:
@thecodejunkie So you can change to this: 
```
x => x.Key.EndsWith("." + components.Item1, StringComparison.OrdinalIgnoreCase)
```
And now we are able to specify correct name without mistakes.

I understand when we have resources with both exact name. But i have resources with different names but ending exactly the same.

khellang:
Will the resource *always* be prefixed with a namespace?
What about a resource named `Orders`? That won't match the `EndsWith("." + "Orders")` check...

Maybe we should split on `.` and do `Equals` on the last item?

khellang:
And again. If you want ut to pull this, I think we need some tests covering a couple of these scenarios.

PR #2066, opened 2015-10-01
ExceptionExtensions Flattens Incorrectly

joebuschmann:
ExceptionExtensions.FlattenInnerExceptions() was returning inner
exceptions of non-aggregate exceptions. This caused the wrong exception to
be passed to the error pipeline.

I found this bug after upgrading from Nancy 1.2 to 1.3. My application has an OnError pipeline handler to take the exception thrown by the app and return the exception text in the response body. Automated tests that were checking error messages returned by the service started failing but only for exceptions where the InnerException property was set to a non-null value. Oddly, the inner exception message text was returned by the service. The bug was in ExceptionExtensions.FlattenInnerExceptions(). It was returning the inner exception for non-aggregate exceptions and throwing away the original exception. This means the OnError pipeline was receiving the incorrect exception.

### DIFF
--- a/src/Nancy.Tests/Helpers/ExceptionExtensionsFixture.cs
+++ b/src/Nancy.Tests/Helpers/ExceptionExtensionsFixture.cs
@@ -1,0 +1,41 @@
+﻿namespace Nancy.Tests.Helpers
+{
+    using System;
+    using System.Collections.Generic;
+    using Nancy.Helpers;
+    using Xunit;
+
+    public class ExceptionExtensionsFixture
+    {
+        [Fact]
+        public void Should_return_exception_if_not_aggregate_exception()
+        {
+            var exception = new Exception("Not an aggregate exception.", new Exception("Inner exception."));
+            var result = exception.FlattenInnerExceptions();
+
+            Assert.Equal(exception, result);
+        }
+
+        [Fact]
+        public void Should_flatten_aggregate_exceptions()
+        {
+            var exception1 = new Exception("Exception 1", new Exception("Inner exception of exception 1"));
+            var exception2 = new Exception("Exception 2", new Exception("Inner exception of exception 2"));
+            var exception3 = new Exception("Exception 3", new Exception("Inner exception of exception 3"));
+            
+            // Aggregate exceptions nested three levels deep.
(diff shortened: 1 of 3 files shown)


khellang:
Hmm. Weird. This fails on Mono. Seems like an issue with the ordering of the exception.

khellang:
> Nancy.Tests.Helpers.ExceptionExtensionsFixture.Should_flatten_aggregate_exceptions [FAIL]
>   Assert.Equal() Failure
>   Position: First difference is at position 0
>   Expected: ReadOnlyCollection<Exception> { System.Exception: Exception 3 ---> System.Exception: Inner exception of exception 3
>               --- End of inner exception stack trace ---, System.Exception: Exception 2 ---> System.Exception: Inner exception of exception 2
>               --- End of inner exception stack trace ---, System.Exception: Exception 1 ---> System.Exception: Inner exception of exception 1
>               --- End of inner exception stack trace --- }
>   Actual:   List<Exception> { System.Exception: Exception 1 ---> System.Exception: Inner exception of exception 1
>               --- End of inner exception stack trace ---, System.Exception: Exception 2 ---> System.Exception: Inner exception of exception 2
>               --- End of inner exception stack trace ---, System.Exception: Exception 3 ---> System.Exception: Inner exception of exception 3
>               --- End of inner exception stack trace --- }

khellang:
I think maybe something like this might work:

```csharp
var innerExceptions = ((AggregateException) result).InnerExceptions;
var expectedExceptions = new[] { exception1, exception2, exception3 };

Assert.True(innerExceptions.SequenceEquals(expectedExceptions));
```

joebuschmann:
Your suggestion should work. I'll try it out and update the pull request.

khellang:
It probably won't give the best failure message, but it does the job.

joebuschmann:
The test is still failing in Mono. I'll try a manual comparison of each item.

Also, I apologize for the extra commits in the pull request. I did a rebase of remote/master on my feature branch before pushing to Github. Not sure how to fix.

joebuschmann:
Passing now. Test assertions are ugly, but like you said, they do the job.

khellang:
Thanks for the contribution @joebuschmann! :heart:

PR #2070, opened 2015-10-06
fix Test LoginFixture.Should_return_unauthorized_without_user_agent()

jongleur1983:
This test did not define not to send a user agent. 
As the user-agent header entry was missing, Nancy.Testing.Browser.CreateRequest() added a default test user agent "Nancy.Testing.Browser" - thus the defining part of the fact to test was missing.

I simply defined user-agent to be null, thus the test sends the request intended and passes again.

### DIFF
--- a/src/Nancy.Demo.Authentication.Token.TestingDemo/LoginFixture.cs
+++ b/src/Nancy.Demo.Authentication.Token.TestingDemo/LoginFixture.cs
@@ -98,6 +98,7 @@ namespace Nancy.Demo.Authentication.Token.TestingDemo
                 with.Accept("application/json");
                 with.FormValue("UserName", "demo");
                 with.FormValue("Password", "demo");
+                with.Header("User-Agent", null);
             });
 
             // Then


khellang:
I'm not sure I understand. This test was failing?

jongleur1983:
it failed here, yes. It was the only test, but it failed. debugging why it did, I found out that the stuff I wrote in the PR comment was the reason.

The test called Browser.POST() calling HandleRequest() (calling HandleRequest) calling Browser.CreateRequest in turn, using the context created before.
Browser.CreateRequest itself contains the following lines:

    if (!contextValues.Headers.ContainsKey("user-agent"))
    {
        contextValues.Headers.Add("user-agent", new[] { "Nancy.Testing.Browser" });
    }

As these add the user agent, the request was "correct" and no error response was returned, but 200 (OK)

khellang:
The thing I find weird is that we run a CI build on each pull request and on each merge to master, and all those are passing :confused: Why is it failing locally on your machine?

khellang:
Anyway, it doesn't matter. This didn't break anyone, so if it fixes the tests on your local machine, we're all happy :smile: Thanks for the contribution! :+1:

PR #2078, opened 2015-10-23
Enabled casting DynamicDictionaryValue to primitives when it is missing a value + to nullable primitives

### LINKED ISSUE #1918: DynamicDictionaryValue does not have implicit operator for nullable types
When handling multiple routes the same way, it could be preferable to cast capture groups to nullable types.

``` csharp
Get["/sizes/{accountId:int}"] =
Get["/sizes/{accountId:int}/{siteId:int}"] =
Get["/sizes/{accountId:int}/{siteId:int}/{year:int}"] =
Get["/sizes/{accountId:int}/{siteId:int}/{year:int}/{month:int}"] =
Get["/sizes/{accountId:int}/{siteId:int}/{year:int}/{month:int}/{day:int}"] = x => { ... }
```

Here I cant simply put
``` csharp
long? siteId = x.siteId;
```
Because if the first route is matched, then siteId isn't set and DynamicDictionaryValue will try to cast the value as a long and throw a NullReferenceException.

Instead I have to do something like
``` csharp
long? siteId = x.siteId.HasValue ? x.siteId : null;
```
Which just seem a bit excessive when an implicit operator could have been implemented instead.

So would it be possible to add implicit operators to DynamicDictionaryValue that supports nullable types?

thecodejunkie:
Fixes #1918

### DIFF
--- a/src/Nancy/DynamicDictionaryValue.cs
+++ b/src/Nancy/DynamicDictionaryValue.cs
@@ -320,7 +320,7 @@
         {
             if (!dynamicValue.HasValue)
             {
-                return new bool?();
+                return default(bool?);
             }
 
             return (bool)dynamicValue;
@@ -358,7 +358,7 @@
         {
             if (!dynamicValue.HasValue)
             {
-                return new int?();
+                return default(int?);
             }
 
             return (int)dynamicValue;
@@ -383,7 +383,7 @@
         {
             if (!dynamicValue.HasValue)
             {
-                return new Guid?();
+                return default(Guid?);
             }
 
             return (Guid)dynamicValue;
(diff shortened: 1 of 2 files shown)


src/Nancy/DynamicDictionaryValue.cs
```diff
@@ -315,6 +316,16 @@ public override string ToString()
             return this.value == null ? base.ToString() : Convert.ToString(this.value);
         }
 
+        public static implicit operator bool?(DynamicDictionaryValue dynamicValue)
+        {
+            if (!dynamicValue.HasValue)
+            {
+                return new bool?();
```
khellang: `null`? or even `default(bool?)`?

khellang:
:heart:

PR #2083, opened 2015-10-23
Support early Flush() of AspNet responses

khellang:
Replaces #1525


mloenow:
Interesting. How's this used in a view?

khellang:
Ask @csainty :wink: TBH, I think this first and foremost meant to be used in a view *engine*, but I might be mistaken. To make it work in your own views, we might have to add some extra plumbing. Want to file an issue?

csainty:
Veil has explicit support for it built in to the view language - https://github.com/csainty/Veil

Razor does not, it renders templates inside out, so you need to use tricks like those employed here - https://github.com/nikmd23/CourtesyFlush - if you want to do this with razor.

Before this PR, the way Nancy worked with the AspNet hosting, it was simply impossible to early flush. This removes that barrier, so now it is up to view engines or other libraries to take advantage.

PR #2084, opened 2015-10-25
Made is possible to disable asp.net output buffering

thecodejunkie:
The ASP.NET config section, for Nancy, now supports a new element
`<disableoutputbuffer value="true/false" />` which controls if output
buffer, on the underlying response stream, should be disabled or not.

Replaces #1549


thecodejunkie:
I updated the documentation to include the new configuration element https://github.com/NancyFx/Nancy/wiki/Hosting-Nancy-with-asp.net#controlling-output-buffering

PR #2088, opened 2015-10-27
INancyEngine now implements IDisposable 

### LINKED ISSUE #1864: INancyEngine should be disposable and be disposed when the boostrapper and container is disposed.
Main problem [is here](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Owin/NancyMiddleware.cs#L98). When the middleware ( -> boostrapper -> container) is disposed, Nancy is still handling requests. Sometimes successfully. 

`INancyEngine` should be `IDisposable` and [here](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Owin/NancyMiddleware.cs#L98) should be seeing an ObjectDisposedException.

damianh:
It will no longer handle subsequent requests and will attempt to cooperatively cancel existing in-flight requests. Fixes #1864.

### DIFF
--- a/src/Nancy/NancyEngine.cs
+++ b/src/Nancy/NancyEngine.cs
@@ -19,7 +19,7 @@
     /// <summary>
     /// Default engine for handling Nancy <see cref="Request"/>s.
     /// </summary>
-    public class NancyEngine : INancyEngine
+    public sealed class NancyEngine : INancyEngine
     {
         public const string ERROR_KEY = "ERROR_TRACE";
         public const string ERROR_EXCEPTION = "ERROR_EXCEPTION";
@@ -30,6 +30,7 @@
         private readonly IEnumerable<IStatusCodeHandler> statusCodeHandlers;
         private readonly IStaticContentProvider staticContentProvider;
         private readonly IResponseNegotiator negotiator;
+        private readonly CancellationTokenSource engineDisposedCts;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="NancyEngine"/> class.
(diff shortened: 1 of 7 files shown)


src/Nancy/NancyEngine.cs
```diff
@@ -19,7 +19,7 @@
     /// <summary>
     /// Default engine for handling Nancy <see cref="Request"/>s.
     /// </summary>
-    public class NancyEngine : INancyEngine
+    public sealed class NancyEngine : INancyEngine
```
damianh: Either `sealed` or fully implement the dispose pattern? Unless designed for inheritance, I'm on the side of 'sealed'.

damianh:
Took slightly different track than reported in the issue - if after the engine is disposed, instead of throwing an `ObjectDisposedException`, the Task is immediately cancelled. Caller tend to treat cancelled tasks differently that other exception types. 

I think this is cleaner and matches what would happen when a request is cancelled under other terms also.

src/Nancy/NancyEngine.cs
```diff
@@ -93,6 +95,10 @@ public class NancyEngine : INancyEngine
 
         public Task<NancyContext> HandleRequest(Request request, Func<NancyContext, NancyContext> preRequest, CancellationToken cancellationToken)
         {
+            var cts = CancellationTokenSource.CreateLinkedTokenSource(this.engineDisposedCts.Token, cancellationToken);
+
+            cts.Token.ThrowIfCancellationRequested();
```
damianh: POI

src/Nancy/NancyEngineExtensions.cs
```diff
@@ -66,8 +66,6 @@ public static NancyContext HandleRequest(this INancyEngine nancyEngine, Request
             nancyEngine
                 .HandleRequest(request, preRequest, cancellationToken)
                 .WhenCompleted(t => onComplete(t.Result), t => onError(t.Exception));
-
-            //this.HandleRequest(request, null, onComplete, onError);
```
damianh: Removed commented out line...

PR #2089, opened 2015-10-27
In proc sessions

DavidLievrouw:
# In-process sessions

This feature was developed in order to be able to work with in-process memory sessions. The current NancyFX version only works with cookie-based sessions, that contain all the session data. All the session data is sent over-the-wire with every request.

A couple of reasons for doing this:
- In my case, I keep a non-serializable object in the session. This is not serializable to a string to put in the cookie. So I would like to keep it in memory, and only send a lightweight correlation ID around.
- When working with larger objects, the cookie data gets quite large. And that's going over the wire with every request. 
- I would like to be able to use cookieless sessions, over HTTPS.

### Super-Duper-Happy-Path

```cs
    public class Bootstrapper : DefaultNancyBootstrapper
    {
        protected override void ApplicationStartup(TinyIoCContainer container, IPipelines pipelines)
        {
            base.ApplicationStartup(container, pipelines);
            InProcSessions.Enable(pipelines);
        }
    }
```
This command enables the in-process sessions, using the default configuration options:
- A cookie is created, containing **only** the unique identifier of the session.
- This cookie is encrypted using CryptographyConfiguration.Default.
- The cookie name is "_nsid".
- There is no Domain or Path setting for the cookie.
- The session timeout is 20 minutes.
- Expired sessions are cleaned automatically every 30 minutes.

### Configuration options
```cs
    public class Bootstrapper : DefaultNancyBootstrapper
    {
        protected override void ApplicationStartup(TinyIoCContainer container, IPipelines pipelines)
        {
            base.ApplicationStartup(container, pipelines);
            
            var sessionConfig = new InProcSessionsConfiguration
            {
                SessionTimeout = TimeSpan.FromMinutes(3),
                CacheTrimInterval = TimeSpan.FromMinutes(10),
                SessionIdentificationMethod = new BySessionIdCookieIdentificationMethod(CryptographyConfiguration.NoEncryption)
                {
                    CookieName = "Mycookie",
                    Domain = ".nascar.com",
                    Path = "/schedule/"
                }
            };
            InProcSessions.Enable(pipelines, sessionConfig);
        }
    }
```

You can specify several configuration options when enabling this feature:
- SessionTimeout: The timeout after which sessions expire.
- CacheTrimInterval: Time interval between automatic cleanup actions of expired sessions (async).
- SessionIdentificationMethod:
  - Defines the method by which the session is identified. Two implementations are provided:
    - BySessionIdCookie (default):
      - CookieName, CookiePath and CookieDomain.
      - CryptographyConfiguration.
    - ByQueryStringParam (for cookieless sessions, more on that later)
      - ParameterName: The name of the querystring parameter in which the session identifier is stored. 
      - CryptographyConfiguration.

### Cookieless sessions
The default session identification method is BySessionIdCookie. This is very similar with how the current CookieBasedSessions feature works: It creates a cookie that is sent with each request, in order to be able to load the session.
The big difference: The session data is not in the cookie value itself. Only a correlation ID.

When I worked with ASP.NET, there was a feature called cookieless sessions. There are a lot of pros and cons to consider, when implementing this. But I found it useful in the past. So I thought it would be a nice feature to add to NancyFX.

```cs
public class Bootstrapper : DefaultNancyBootstrapper
    {
        protected override void ApplicationStartup(TinyIoCContainer container, IPipelines pipelines)
        {
            var sessionConfig = new InProcSessionsConfiguration
            {
                SessionIdentificationMethod = new ByQueryStringParamIdentificationMethod(CryptographyConfiguration.Default)
            };
            InProcSessions.Enable(pipelines, sessionConfig);

            base.ApplicationStartup(container, pipelines);
        }
    }
```

When navigating to a page without the session ID querystring parameter, a 302-FOUND response with a location header is sent back to the client. This location header redirects to the same location, but includes the session ID querystring parameter.

### Sample
I added a tiny sample project to the solution: Nancy.Demo.InProcSessions.

### Roll your own
Currently, two session identification methods are implemented:
- BySessionIdCookie
- ByQueryStringParam

You can create your own method. Implement the interface and wire it up.

```cs
    /// <summary>
    /// Identification method for in-process memory based sessions.
    /// </summary>
    public class MySessionIdentificationMethod : IInProcSessionIdentificationMethod
    {
        /// <summary>
        /// Load the session identifier from the specified context.
        /// </summary>
        /// <param name="context">The current context.</param>
        /// <returns>The identifier of the session for the current request.</returns>
        public SessionId GetCurrentSessionId(NancyContext context)
        {
            // Load the session identifier from the context.
            // Or create a new one, if no session data is sent with the request.
        }

        /// <summary>
        /// Save the session identifier in the specified context.
        /// </summary>
        /// <param name="sessionId">The identifier of the session.</param>
        /// <param name="context">The current context.</param>
        public void SaveSessionId(SessionId sessionId, NancyContext context)
        {
            // Make sure that the next request from the client
            // contains this session ID.
        }
    }
```

And modify your bootstrapper:
```cs
    public class Bootstrapper : DefaultNancyBootstrapper
    {
        protected override void ApplicationStartup(TinyIoCContainer container, IPipelines pipelines)
        {
            var sessionConfig = new InProcSessionsConfiguration
            {
                SessionIdentificationMethod = new MySessionIdentificationMethod()
            };
            InProcSessions.Enable(pipelines, sessionConfig);

            base.ApplicationStartup(container, pipelines);
        }
    }
```

### Breaking changes
None, it's completely new, in a separate namespace.

### Not happy about...
I needed to update the SharedAssemblyInfo.cs file, in order to be able to fake internal interfaces using FakeItEasy. I added this line:

```cs
[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
```

Not really happy about that. But I didn't want to make all the feature internals publicly visible. Users should not be bothered with a bunch of internal interfaces, in my opinion.

### Not interested?
When this feature doesn't get pulled, I would like to provide it as a NuGet package.

### Contact
Feel free to contact me. I would like some feedback on this.
david.lievrouw at gmail.com


phillip-haydon:
Wow, what a huge PR. :D

Is there any reason why you baked it into core of Nancy and not as a separate library? That would keep the plugable design of Nancy, and not matter if it's pulled or not.

phillip-haydon:
Also, before Mr Style Cop comes. All the braces would need to be fixed. (need to start on a new line)

```
if (....)
{

}
```

Not

```
if (....) {

}
```

DavidLievrouw:
No, I just put it next to CookieBasedSessions, that was already there. I can move it to a separate project, of course. Just let me know and I'll do it.
I would move it to a new project called "Nancy.Session.InProcSessions", and create a nuspec file.

Meanwhile, I'll look into the curly braces. Shouldn't we store this in the .dotsettings file?

phillip-haydon:
Yes it it could be in it's own project that would be preferred. Couple of the main reasons for having Cookie sessions is:

1) Load Balancer means cookies are controlled by the client, not the server, so no need for centralized cookies like State Server or SQL Server. 

2) We (I think I speak for all, or most contributors of Nancy) believe Sessions are... more or less an anti pattern, or a bad solution to an almost non-existent problem.

So having InProcSessions baked into the Core would be bad.

However there are people who would love this PR so I think it would be great to pull it in and have an opt-in NuGet package for it. 

It could also set the stages up for other session implementations in the future.

khellang:
The problem of pulling this into the main Nancy repo is that we have to maintain the code until the end of time.

I'd rather see this sitting somewhere else where the code can get the :heart: it deserves.

DavidLievrouw:
OK, I'll move it into separate projects.

By the way, I totally agree about sessions being an anti-pattern. But we are migrating from a legacy application that relies heavily on sessions. So I needed this anyway, and I thought I'd push it into the world.

If you decide not to pull it, please let me know and I'll create my personal repo and Nuget package.

damianh:
> By the way, I totally agree about sessions being an anti-pattern. 

Yes, this is bit of a philosophical concern. If we pull this into the core library we're saying 'this is good to use' for _all_ nancy users. I think sessions, while supportable, should be discouraged at bit. Having to add another nuget package, as a deliberate decision, is enough of a speed bump to make people think.

@DavidLievrouw This is an awesome PR, but I think this as an external nuget package would be a better way to proceed.

thecodejunkie:
While this indeed is an impressive pull-request (the test coverage alone is enough to earn our respect), I do not feel that this is something that we can accept. The simple fact is maintenance. Historically we've taken a board large features and they have always turned out to be a massive undertaking when it comes to maintaining and extending, especially since we've not written a single line of it ourselves.

However, I would very much like to see you maintain this as a community project and help grow the Nancy ecosystem. We would support this in any way we can, just not through ownership. We would also avoid housing it on the NancyFx organisation here at GitHub, because our experience has show that we do gain ownership by proxy when we do

So, I'm going to be closing this, but please tell us if you need any help to promote it. 

Thank you!

DavidLievrouw:
OK, available here:
https://www.nuget.org/packages/Nancy.Session.InProc/

and here:
https://github.com/DavidLievrouw/Nancy.Session.InProc

Thanks for your feedback!

khellang:
Lovely! :heart:

phillip-haydon:
Awesome stuff! <3

PR #2096, opened 2015-11-01
Delete wcf host and test projects.

damianh:
We're not supporting this in v2 :)


phillip-haydon:
YES!. 

Next do the SelfHost one too!

khellang:
This is the kind of diffs I like the most :heart:

PR #2097, opened 2015-11-01
Delete Hosting.Self projects. 

damianh:
Self hosting will be via owin (and later) hosts in v2.

Should probably still have a self hosting sample projects, but I'd recommend a different repo would be better imho.


phillip-haydon:
Yayyyyy no more decoding errors.

grumpydev:
Woah there, I think some people are jumping the gun a bit and conflating "vnext" with "v2", don't recall discussing or agreeing to drop self host for v2 - I know v2 will be a "breaking change", but self hosting is pretty much equal with the asp.net hosting as the most popular hosting package - no way should we be dropping this without discussions and a proper alternative in place.

damianh:
We got our wires cross there then, fair enough to the revert.

thecodejunkie:
The only hosting option we've talked about _dropping_ is the WCF one. All other hosting options have been discussed if we should make Nancy speak OWIN natively and turn the other packages into meta packages. Even here we need to decide if/how this is going to be done. For example, do we really want to take a dependency on any Microsoft OWIN packages (like Microsoft.Owin.SelfHost) or roll our own. Baby steps :)

Issue #2076, opened 2015-10-20
BindTo(...) within inherited class constructor does not populate properties


JC008:
The following illustrates the issue:

```c#
using System;
using Nancy;
using Nancy.ModelBinding;
using Nancy.Testing;

namespace Question
{
    public class BaseNancyHandler
    {
        protected BaseNancyHandler(INancyModule nancyModule)
        {
            nancyModule.BindTo(this);
        }
    }

    public class ExampleHandler
        : BaseNancyHandler
    {
        public ExampleHandler(INancyModule nancyModule)
            : base(nancyModule)
        {
            // nancyModule.BindTo(this); // <-- works when this is enabled

            if (string.IsNullOrEmpty(UserName))
            {
                throw new ArgumentNullException(nameof(UserName));
            }

            if (string.IsNullOrEmpty(Password))
            {
                throw new ArgumentNullException(nameof(Password));
            }
        }

        public string UserName { get; set; }
        public string Password { get; set; }
    }

    public class ExampleModule 
        : NancyModule
    {
        public ExampleModule()
        {
            Post["/login/"] = _ => new ExampleHandler(this);
        }
    }

    public static class TestExample
    {
        public static void Main()
        {
            Run();
        }

        public static void Run()
        {
            var bootstrapper = new DefaultNancyBootstrapper();
            var browser = new Browser(bootstrapper);

            var response = browser.Post("/login/", (with) => {
                with.HttpRequest();
                with.FormValue("username", "badusername");
                with.FormValue("password", "badpassword");
            });

        }
    }
}
```

khellang:
That's because `BindTo` is generic (`public static TModel BindTo<TModel>(this INancyModule module, TModel instance)`) and the type is inferred to `BaseNancyHandler` (which `this` is). There's no way you'll get that to work because you don't have enough information about the derived type in the base class.

I'm not really sure what you gain by doing this. Why are you returning a handler, which in reality is just a DTO that has the resposiblity to deserialize itself using a Nancy module? It all sounds very weird.

thecodejunkie:
Like @khellang said, this is not a Nancy "problem". There is no way that the base class could know which sub-class `this` would reference. I'm closing this issue, but feel free to continue with the discussions

JC008:
See #2102

PR #2102, opened 2015-11-05
Fix for #2076

JC008:
The dynamic object receives a reference to the parent type in the binder
and not the inherited type. As the instance type is detectable and is
the eventual target, it should be used to call Bind on the model binder.


jchannon:
Could you amend the test to show that the base class and inherited class have properties that can both get set by one request

khellang:
Thanks for this, @JC008! :heart:

PR #2105, opened 2015-11-10
Using async/await

damianh:
Removes the .net 4.0 era TPL code and uses .net 4.5's async / await.


src/Nancy.Demo.Authentication.Forms.TestingDemo/LoginFixture.cs
```diff
@@ -17,10 +17,10 @@ public LoginFixture()
         }
 
         [Fact]
-        public void Should_redirect_to_login_with_error_querystring_if_username_or_password_incorrect()
+        public async Task Should_redirect_to_login_with_error_querystring_if_username_or_password_incorrect()
         {
             // Given, When
-            var response = browser.Post("/login/", (with) =>
+            var response = await browser.Post("/login/", (with) =>
```
damianh: `Browser` is now entirely async.

src/Nancy.Testing/Browser.cs
```diff
@@ -219,7 +219,9 @@ public BrowserResponse HandleRequest(string method, Url url, Action<BrowserConte
             var request =
                 CreateRequest(method, url, browserContextValues);
 
-            var response = new BrowserResponse(this.engine.HandleRequest(request), this, (BrowserContext)browserContextValues);
+            var context = await this.engine.HandleRequest(request);
```
damianh: **Question**: anywhere we do an await within Nancy, should we also `.ConfigureAwait(false)`. My understanding is that all libraries should be doing that.

src/Nancy.Tests/Unit/AfterPipelineFixture.cs
```diff
@@ -96,43 +96,12 @@ public void Pipeline_containing_another_pipeline_will_invoke_items_in_both_pipel
...
+            await pipeline.Invoke(CreateContext(), new CancellationToken());
 
             Assert.True(item1Called);
             Assert.True(item2Called);
             Assert.True(item3Called);
             Assert.True(item4Called);
         }
-
-        [Fact]
-        public void Pipeline_containing_method_returning_null_throws_InvalidOperationException()
```
damianh: Deleted because internally the await will never return a null task.

src/Nancy.Tests/Unit/NancyEngineFixture.cs
```diff
@@ -42,7 +43,7 @@ public NancyEngineFixture()
             this.negotiator = A.Fake<IResponseNegotiator>();
 
             A.CallTo(() => this.requestDispatcher.Dispatch(A<NancyContext>._, A<CancellationToken>._))
-                .Returns(CreateResponseTask(new Response()));
+                .Returns(Task.FromResult(new Response()));
```
damianh: Got rid of `TaskHelpers.CompletedTask` and used `Task.FromResult` where I could.

src/Nancy/NancyEngineExtensions.cs
```diff
@@ -13,75 +11,9 @@ public static class NancyEngineExtensions
...
-        }
-
-        /// <summary>
-        /// Handles an incoming <see cref="Request"/>.
-        /// </summary>
-        /// <param name="nancyEngine">The <see cref="INancyEngine"/> instance.</param>
-        /// <param name="request">An <see cref="Request"/> instance, containing the information about the current request.</param>
-        /// <param name="preRequest">Delegate to call before the request is processed</param>
-        /// <returns>A <see cref="NancyContext"/> instance containing the request/response context.</returns>
-        public static NancyContext HandleRequest(this INancyEngine nancyEngine, Request request, Func<NancyContext, NancyContext> preRequest)
```
damianh: Working with `NancyEngine` is _async only_.

src/Nancy.Testing/Browser.cs
grumpydev: Yep
________________________________

damianh: already made that change, see last commit

khellang:
Assigning to @grumpydev :smile:

src/Nancy.Demo.Authentication.Forms.TestingDemo/LoginFixture.cs
grumpydev: Not 100% sure this is wise - it's a sneaky breaking change because stuff may well still build, it just (potentially) won't work properly.

We are planning on bringing in the LegacyNancyModule and making NancyModule async only, perhaps we could do a similar thing here?

src/Nancy.Testing.Tests/ConfigurableBootstrapperFixture.cs
```diff
@@ -219,5 +219,21 @@ public BlowUpModule()
                 Get["/"] = _ => { throw new InvalidOperationException("Oh noes!"); };
             }
         }
+
+        private static class Record
```
grumpydev: Surprised xunit doesn't support this directly tbh.

src/Nancy.Tests/xUnitExtensions/RecordAsync.cs
```diff
@@ -0,0 +1,39 @@
+﻿namespace Nancy.Tests.xUnitExtensions
+{
+    using System;
+    using System.Threading.Tasks;
+    using Xunit;
+
+    public static class RecordAsync
```
grumpydev: Be nice to not have to repeat this, but probably the lesser of two evils compared to adding references.

src/Nancy/AfterPipeline.cs
```diff
@@ -160,7 +74,7 @@ private static Action<Task> ExecuteTasksContinuation(NancyContext context, Cance
             {
                 try
                 {
-                    syncDelegate.Invoke(ctx);
+                    syncDelegate(ctx);
```
grumpydev: Personally I prefer .Invoke, more meaningful imo.

grumpydev:
Made some minor comments and one "major" one around the breaking change to Browser - I know 2.0 is a breaking release, but seems a bit out of kilter with the approach we've discussed for NancyModule - not precious about it though, we can always add a LegacyBrowser wrapper around the new Browser ?

Other than that, and the fact you've deleted so much of my carefully crafted code, looks good! :shipit:

src/Nancy.Testing.Tests/ConfigurableBootstrapperFixture.cs
khellang: I think it does? `Assert.Throws<T>` and `Assert.ThrowsAsync<T>` both first assert the correct exception and then returns it.

khellang: There's also the `Record` class which does exactly this; https://github.com/xunit/xunit/blob/dd20821043bcb23eeecd03d0468f08c4b5378c0b/src/xunit.core/Record.cs

src/Nancy.Demo.Authentication.Forms.TestingDemo/LoginFixture.cs
damianh: I'm not fond with wrapping a sync api (`.Wait` `.Result` ) around and async api at a library level. Funny things happen with deadlock and what not. See `HttpClient` 

The methods _could_ be renamed `PostAsync()` so it more of a clear breaking change?

src/Nancy.Tests/xUnitExtensions/RecordAsync.cs
damianh: I choose the lesser of two evils....

src/Nancy.Demo.Authentication.Forms.TestingDemo/LoginFixture.cs
grumpydev: Waiting is exactly what the existing code does/did though so I don't think it will be an issue - obviously it will be marked as deprecated, but I think if *everyone's* tests break with a 2.0 upgrade, and it means editing every test to fix, it will piss off and put off an awful lot of our users.

I would rather that shim than postasync personally - @thecodejunkie ?

src/Nancy/AfterPipeline.cs
damianh: Disagree but don't really care. Will revert. :)

src/Nancy.Testing.Tests/ConfigurableBootstrapperFixture.cs
damianh: It's in xunit 2.x, but I failed to port to that because Mono

src/Nancy.Demo.Authentication.Forms.TestingDemo/LoginFixture.cs
damianh: Another thought - keeping the sync methods but marked as `[Obsolete(false)]` and add the async overloads?

grumpydev: Cant' be an overload can it? Only differs on return type (T vs Task<T>)

damianh:
```
[Obsolete(false)]
public BrowserResponse Post(..)
{
    return PostAync(..).Result;
}

public Task<BrowserResponse> PostAsync()
{
}
```

khellang:
```csharp
[Obsolete(false)]
public BrowserResponse Post(..)
{
    return PostAync(..).GetAwaiter().GetResult();
}

public Task<BrowserResponse> PostAsync()
{
}
```

damianh:
Renamed the browser async methods to `*Async(..)` . Added sync overloads (`Post()` etc) that calls their async counterparts and marked them as obsolete.

I didn't write tests over the obsolete sync methods... do I have to?

grumpydev:
XAsync methods? Yuck

src/Nancy.Testing.Tests/BrowserDefaultsFixture.cs
```diff
@@ -25,7 +25,7 @@ public void Should_pass_default_headers_in_get_request_when_using_configurable_b
             var sut = new Browser(with => with.Module(this.captureRequestModule), defaults: to => to.Accept(this.expected));
             
             // When
-            sut.Get("/");
+            sut.GetAsync("/");
```
khellang: Shouldn't all of these be `await`ed?

src/Nancy.Testing.Tests/BrowserFixture.cs
```diff
@@ -318,15 +322,13 @@ public void Should_be_able_to_continue_with_another_request()
...
         [Fact]
-        public void Should_maintain_cookies_when_chaining_requests()
+        public async Task Should_maintain_cookies_when_chaining_requests()
         {
             // Given
             // When
-            var result = browser.Get(
-                             "/session",
-                             with => with.HttpRequest())
-                .Then
```
khellang: What happened to `.Then`?

src/Nancy.Testing/Browser.cs
```diff
@@ -54,9 +52,10 @@ public Browser(INancyBootstrapper bootstrapper, Action<BrowserContext> defaults
         /// <param name="path">The path that is being requested.</param>
         /// <param name="browserContext">An closure for providing browser context for the request.</param>
         /// <returns>An <see cref="BrowserResponse"/> instance of the executed request.</returns>
+        [Obsolete("Use async overload instead.", false)]
         public BrowserResponse Delete(string path, Action<BrowserContext> browserContext = null)
         {
-            return this.HandleRequest("DELETE", path, browserContext);
+            return this.DeleteAsync(path, browserContext).Result;
```
khellang: I still think these should be `GetAwaiter().GetResult()` to not wrap exceptions in `AggregateException`...

src/Nancy.Testing/Browser.cs
```diff
@@ -54,9 +52,10 @@ public Browser(INancyBootstrapper bootstrapper, Action<BrowserContext> defaults
         /// <param name="path">The path that is being requested.</param>
         /// <param name="browserContext">An closure for providing browser context for the request.</param>
         /// <returns>An <see cref="BrowserResponse"/> instance of the executed request.</returns>
+        [Obsolete("Use async overload instead.", false)]
```
khellang: `error: false`?

src/Nancy.Testing/Browser.cs
```diff
@@ -109,9 +155,33 @@ public BrowserResponse Head(string path, Action<BrowserContext> browserContext =
...
+        /// <summary>
+        /// Performs a HEAD request against Nancy.
+        /// </summary>
+        /// <param name="url">The url that is being requested.</param>
+        /// <param name="browserContext">An closure for providing browser context for the request.</param>
+        /// <returns>An <see cref="BrowserResponse"/> instance of the executed request.</returns>
+        public Task<BrowserResponse> HeadAsync(Url url, Action<BrowserContext> browserContext = null)
+        {
             return this.HandleRequest("HEAD", url, browserContext);
+
```
khellang: :fire:

src/Nancy.Tests/NoAppStartupsFixture.cs
```diff
@@ -69,12 +69,12 @@ public void When_AutoRegistration_Is_Enabled_Should_Throw()
                 var browser = new Browser(bootstrapper);
 
                 // When
-                browser.Get("/");
+                browser.GetAsync("/");
```
khellang: `await`?

khellang:
Just a couple of nits from me :+1: :shipit:

damianh:
You're not giving me any options here.... browser interacts with nancy
engine which is pure async therefore browser is pure async. If you want
sync and async "overloads" on same class, you must name them differently.

src/Nancy.Testing.Tests/BrowserDefaultsFixture.cs
damianh: yep, slipped past.

src/Nancy.Testing/Browser.cs
damianh: See @grumpydev 's comment regarding forcing people to rewrite / fix all their tests. `True` here would make such happen too.

src/Nancy.Testing.Tests/BrowserFixture.cs
damianh: extensions methods around awaited methods = fugly. Have to surround the method in more brackets first:

```
(await browser.Get("/session", with => with.HttpRequest())).Then()
```

src/Nancy.Testing/Browser.cs
damianh: ok, nice tip

src/Nancy.Testing/Browser.cs
khellang: I'm not saying we set it to `true`, it just wasn't obvious what `false` meant...

src/Nancy.Testing.Tests/BrowserFixture.cs
khellang: You could add an extension method to `Task<BrowserResponse>` though, so you can chain them?

grumpydev:
Who said they had to be the same class? My suggestion was a sync wrapper class - a breaking change, but one that's easy to fix with a find replace from Browser to LegacyBrowser, same as it will be for LegacyNancyModule when that change is made.

khellang:
In the cases where we can break stuff and still provide a decent migration path, like this, I think we should just do it.

So :+1: to `LegacyBrowser` - possibly with an `ObsoleteAttribute` as well.

just do it

damianh:
What is here now is not a breaking change. Upgrading to 2.x will mean a user will not have to modifiy their tests at all, until we set `[Obsolete(..., true)]`. They will get compiler warnings, but test will (should) fully work. This is a good migration path.

My original proposal was a breaking change but, in your words, 'sneaky'. That is, tests would not be obviously compilation broken (though there will be lots [CS4104 warnings](https://msdn.microsoft.com/en-us/library/hh873131.aspx) ). Here a user would have to make all their tests async, which is going to be the recommended path anyway.

With `LegacyBrowser` ( :-1: ) they will have to update every single test anyway but...  it's still sneaky in that it would not be obvious unless you treat compilation warnings as errors.

No opinion yet on `LegacyModule`, I think that is a different concern.

@khellang it make no sense to me to put an `[Obsolete]` immediately on a new type.

phillip-haydon:
I'm happy with the `XAsync` methods. Would rather that than legacy Browser.

grumpydev:
I think people are missing the point on the XAsync stuff - that's not the API we want to end up with, it looks nasty, so if we go down that road we are introducing a non-breaking change now and saying to people "move all this to these new async methods because the sync ones are going away", then we will make *another* breaking change to remove the sync ones and rename the async ones to Get etc as they should be (calling them XAsync is ugly and pointless when the sync ones are removed)

With the Legacy classes there should be compiler errors (possible there won't be depending on the tests), but rather than having to go in and add .Result or .Wait() into every test (which they'd have to if we just made the existing methods async) it's a simple Ctrl+H "new Browser", "new LegacyBrowser".

khellang:
> @khellang it make no sense to me to put an [Obsolete] immediately on a new type.

Why? It's a migration path. We don't want to keep the `LegacyBrowser` around forever and it will be removed in the next major or something.

src/Nancy.Demo.Authentication.Forms.TestingDemo/LoginFixture.cs
horsdal: I would rather go with a new `SyncBrowser` or `LegacyBrowser` type for the sync versions and have `Browser` be fully async. I don't like introducing the `*Async` methods - once the sync ones are removed we'll want to remove the `Async` suffixes, but that'll be another round of breaking changes.

src/Nancy/AfterPipeline.cs
```diff
@@ -60,94 +52,16 @@ public AfterPipeline(int capacity)
...
-                tcs.SetResult(null);
-            }
-
-            return tcs.Task;
-        }
-
-        private static void ExecuteTasksInternal(NancyContext context, CancellationToken cancellationToken, IEnumerator<Func<NancyContext, CancellationToken, Task>> enumerator, TaskCompletionSource<object> tcs)
-        {
-            while (true)
+            while (enumerator.MoveNext())
```
horsdal: Any reason not to foreach over `PipelineDelegates`?

src/Nancy/BeforePipeline.cs
```diff
@@ -50,113 +49,19 @@ public BeforePipeline(int capacity)
...
-        {
-            // Endless loop to try and optimise the "main" use case of
-            // our tasks just being delegates wrapped in a task.
-            //
-            // If they finish executing before returning them we will
-            // just loop around this while loop running them one by one,
-            // as soon as we have to return, or a task is actually async,
-            // then we will bale out and set a continuation.
-            while (true)
+            while (enumerator.MoveNext())
```
horsdal: do.

grumpydev:
Should we merge this in now or wait until 3.0? ;)

damianh:
I swear to god :)

PR #2108, opened 2015-11-11
Added ISerializerFactory + implementation

thecodejunkie:
The idea is that this what you will take a dependency on, instead of `IEnumerable<ISerializer>`, and it will give you the serializer that matches the provided `MediaRange`.

The default implementation will take the default serializers into account and only return them if no other match could be made. 

If there are multiple matched (not counting the defaults) then it will throw an `InvalidOperationException` with a detailed error message, telling you about the types of the implementations that it matched.

I have also updated the `ISerializer.CanSerialize` and `Serialize` methods to accept a `MediaRange` instance, instead of string, for the `contentType` parameter. This works fine since it can implicitly be case to/from a string.

--

Except for adding this in to `NancyInternalConfiguration` I've not made any use of this yet. I wanted to get some feedback before I started ripping stuff apart and plugging this in. The idea is that this both makes things more testable, but also reduces the need for all `ISerializer` users to do the matching themselves. We currently have a couple of places (like the `AsXml` and `AsJson` extension methods) where we explicitly new up instances of `DefaultXmlSerializer` and `DefaultJsonSerializer`, creating hidden dependencies in the code.

This is a bit of yakshaving before migrating the `JsonSettings` and `XmlSettings` into the new unified configuration API.

### DIFF
--- a/src/Nancy.Demo.Hosting.Aspnet/PngSerializer.cs
+++ b/src/Nancy.Demo.Hosting.Aspnet/PngSerializer.cs
@@ -26,7 +26,7 @@
         /// <returns>True if supported, false otherwise</returns>
         public bool CanSerialize(MediaRange mediaRange)
         {
-            return ((string)mediaRange).Equals("image/png", StringComparison.OrdinalIgnoreCase);
+            return mediaRange.Matches("image/png");
         }
 
         /// <summary>
--- a/src/Nancy.Demo.Hosting.Aspnet/PngSerializer.cs
+++ b/src/Nancy.Demo.Hosting.Aspnet/PngSerializer.cs
@@ -5,6 +5,7 @@
     using System.Drawing;
     using System.Drawing.Imaging;
     using System.IO;
+    using Nancy.Responses.Negotiation;
 
     /// <summary>
     /// If you request /negotiated with an accept header with the value image/png you will get an image back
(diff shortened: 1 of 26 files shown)


src/Nancy.Demo.Hosting.Aspnet/PngSerializer.cs
```diff
@@ -21,11 +22,11 @@ public PngSerializer(IRootPathProvider rootPathProvider)
...
         /// Whether the serializer can serialize the content type
         /// </summary>
-        /// <param name="contentType">Content type to serialise</param>
+        /// <param name="mediaRange">Content type to serialise</param>
         /// <returns>True if supported, false otherwise</returns>
-        public bool CanSerialize(string contentType)
+        public bool CanSerialize(MediaRange mediaRange)
         {
-            return contentType.Equals("image/png", StringComparison.OrdinalIgnoreCase);
+            return ((string)mediaRange).Equals("image/png", StringComparison.OrdinalIgnoreCase);
```
khellang: Ugh, dat cast :mask:

thecodejunkie:
Seems to work. All tests are green, demo project runs and json/xml serializer routes works :)

khellang:
I assume that meant it was good to merge :smile:

khellang:
Now time to update the serializers :wink:

PR #2115, opened 2015-11-17
Migrated XmlSettings to the configuration API

thecodejunkie:


src/Nancy.Testing/BrowserContext.cs
```diff
@@ -14,10 +14,12 @@
     public class BrowserContext : IBrowserContextValues
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="BrowserContext"/> class.
+        /// Initializes a new instance of the <see cref="BrowserContext"/> class,
+        /// with the provided <see cref="INancyEnvironment"/>.
         /// </summary>
-        public BrowserContext()
+        public BrowserContext(INancyEnvironment environment)
```
khellang: No `param` doc?

src/Nancy/Responses/DefaultXmlSerializer.cs
```diff
@@ -64,19 +77,19 @@ public void Serialize<TModel>(MediaRange mediaRange, TModel model, Stream output
             }
         }
 
-        private static bool IsXmlType(string contentType)
+        private static bool IsXmlType(MediaRange mediaRange)
```
khellang: What's the reason for this? This means we're ToString'ing the `mediaRange` twice here. Once in the `IsNullOrEmpty` implicit conversion and once below.

src/Nancy/Xml/XmlSettingConfigurationExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public static class XmlSettingConfigurationExtensions
+    {
+        /// <summary>
+        /// Configures <see cref="XmlSettings"/>.
+        /// </summary>
+        /// <param name="environment"><see cref="INancyEnvironment"/> that should be configured.</param>
+        /// <param name="enableEncoding"><see langword="true"/> if encoding should be enabled, otherwise <see langword="false"/>.</param>
+        /// <param name="defaultEncoding">The default <see cref="Encoding"/>.</param>
+        public static void XmlSettings(this INancyEnvironment environment, bool enableEncoding, Encoding defaultEncoding = null)
```
khellang: Should we make this an extra overloads? Optional arguments version so poorly...

khellang:
LGTM! Just a few minor comments :shipit:

src/Nancy/Responses/DefaultXmlSerializer.cs
thecodejunkie: The signature was updates since stuff now speaks `MediaRange`, but yeah that method is probably better of accepting a string so we only `ToString` once. Will revert

src/Nancy/Xml/XmlSettingConfigurationExtensions.cs
thecodejunkie: The idea here is that `enableEncoding` is never optional, while `defaultEncoding` is.

src/Nancy.Testing/BrowserContext.cs
thecodejunkie: Whoopsie! Good catch =)

thecodejunkie:
Fixed review comments + renamed `XmlSettings` -> `XmlConfig`

thecodejunkie:
Bunch of changes just went in

src/Nancy/Xml/XmlConfigurationExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public static class XmlConfigurationExtensions
+    {
+        /// <summary>
+        /// Configures <see cref="XmlConfiguration"/>.
+        /// </summary>
+        /// <param name="environment"><see cref="INancyEnvironment"/> that should be configured.</param>
+        /// <param name="enableEncoding"><see langword="true"/> if encoding should be enabled, otherwise <see langword="false"/>.</param>
+        /// <param name="defaultEncoding">The default <see cref="Encoding"/>.</param>
+        public static void Xml(this INancyEnvironment environment, bool enableEncoding, Encoding defaultEncoding = null)
```
khellang: I'd still like to see this without the optional argument, just for versioning's sake. You can bet we'll have to change this (pr an overload) once the settings change.

thecodejunkie: This is per design and per the discussions we had during the creation of the config API. The idea is to promote the use of named parameters when calling these methods to make the configuration more explicit and clear

```c#
env.Foo.(10, "John Doe", new DateTime(2015, 11, 18), () => Guid.NewGuid());

vs.

env.Foo(
   maxNumberOfRequests: 10,
   defaultUserName: "John Doe",
   earliestDate: new Date(2015, 11, 18),
   userIdFactory: () => Guid.NewGuid()
);
```

It is already in place for the `DiagnosticsConfiguration` as well. Because this is a config API it adds a lot of clarity when reading the config code.

khellang: You know you can used named parameters without them being optional? The difference is that non-optional parameters version much better. IMO, we should change both this and the `DiagnosticsConfiguration` extensions.

grumpydev: If you have 6 or 7 parameters you either need a bazillion overloads to cover all eventualities, or optional parameters - assuming some of the config options are actually optional this makes much more sense.

khellang: I think *most* config options are atullay optional. They should have sensible defaults. I'm just worried by all the people we're going to break every time we add a config option.

PR #2116, opened 2015-11-17
Diagnostics configuration modifications

thecodejunkie:
Fixed some XML comments and inconsistencies that I found

### DIFF
--- a/src/Nancy/Diagnostics/DiagnosticsConfiguration.cs
+++ b/src/Nancy/Diagnostics/DiagnosticsConfiguration.cs
@@ -3,7 +3,7 @@
     using Nancy.Cryptography;
 
     /// <summary>
-    /// Settings for the diagnostics dashboard.
+    /// Configuration for the diagnostics dashboard.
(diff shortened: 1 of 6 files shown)


src/Nancy/Diagnostics/DiagnosticsConfiguration.cs
```diff
@@ -3,7 +3,7 @@
     using Nancy.Cryptography;
 
     /// <summary>
-    /// Settings for the diagnostics dashboard.
+    /// Configuration for the diagnostics dashboard.
     /// </summary>
     public class DiagnosticsConfiguration
```
thecodejunkie: Opps, missed an important change in this class

khellang: Does that mean that this isn't good to merge?

thecodejunkie: It is now =)

PR #2121, opened 2015-11-20
Virtual Dispose in NancyBootstrapperBase

asbjornu:
This PR makes made `NancyBootstrapperBase.Dispose()` `virtual` so it can be overridden in deriving classes. This makes it easier for `StructureMapNancyBootstrapper` and others to dispose resources other than the `ApplicationContainer`.

### DIFF
--- a/src/Nancy/Bootstrapper/NancyBootstrapperBase.cs
+++ b/src/Nancy/Bootstrapper/NancyBootstrapperBase.cs
@@ -396,7 +396,7 @@
         /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
         /// </summary>
         /// <filterpriority>2</filterpriority>
-        public virtual void Dispose()
+        public void Dispose()
         {
             // Prevent StackOverflowException if ApplicationContainer.Dispose re-triggers this Dispose
             if (this.disposing)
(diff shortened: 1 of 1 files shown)


khellang:
This way you'd have to either re-implement the `disposing` and `initialised` logic and make sure you dispose the container (unless you call base as well). Wouldn't it be better to have a separate overload that's called for additional resources?

asbjornu:
@khellang True. I've changed it to be more according to [this guide](https://msdn.microsoft.com/en-us/library/fs2xkftw(v=vs.110).aspx), although there might be some confusion over the collision between the field `disposing` and the parameter to `Disposing()`. However, since this is a known and common pattern, I think naming the parameter `disposing` is important and that if anything should change, then it is the name of the field.

src/Nancy/Bootstrapper/NancyBootstrapperBase.cs
```diff
@@ -426,6 +426,8 @@ public void Dispose()
             catch (ObjectDisposedException)
             {
             }
+
+            Dispose(true);
```
khellang:
```csharp
if (container == null)
{
    return;
}
```

This call won't be made if the above code kicks in.

asbjornu: Of course.

khellang:
Thanks @asbjornu! :heart:

PR #2125, opened 2015-11-22
[WIP]Migrated StaticConfiguration to config API

thecodejunkie:
_Opening this pull-request early for the sake of discussions_

## Background

The `StaticConfiguration` design has served us well over the years, but the time has come for us to remove this static class. We need to move away from static instances for various reasons (multi-tenancy, DNX, improved testing and so on). The class will be migrated over the new config API that was implemented in #1999 

## Overview

This one is probably going to be broken up into several different configurations (one pull-request each), because right now `StaticConfiguration` is a mix of things.
The following members are on the class today

- `DisableErrorTraces` (Tracing)
- `DisableMethodNotAllowedResponses` (Routing)
- `CaseSensitive` (???)
- `EnableHeadRouting` (Routing)
- `IsRunningDebug` (Discussion in the comments [here](https://github.com/NancyFx/Nancy/pull/2125#issuecomment-158758527))
- `RequestQueryFormMultipartLimit` (Request file uploads)
- `EnableRequestTracing` (Tracing)
- `DisableRequestStreamSwitching` (Request file uploads)
- `AllowFileStreamUploadAsync` (Request file uploads)
- `EnableRuntimeViewDiscovery` (Views)
- `EnableRuntimeViewUpdates` (Views)

These need to be

1. Decided if any of them should be deleted, i.e does it still make sense to have all of them around?
1. Grouped into different configs
1. Renamed to make more sense in their new context

I have a feeling that `IsRunningDebug` probably should not be a config thing, but rather be placed in a class that you can take a dependency on if you need to know

```#
public interface IEnvironmentInformation
{
    bool IsDebug();
}
```

I know the name sucks (suggestions?), but it illustrates the point ;)
Update: See the [comment below](https://github.com/NancyFx/Nancy/pull/2125#issuecomment-158758527) for a discussion / spike on this

## Naming

Right now the static configurations uses a mix of `EnableXXXX` and `DisableXXXX`. I think all of these should follow the same naming convention and use appropriate default values instead.

## Progress

Removed all use of the `static` ketword in the `StaticConfiguration` class and ended up with 32 errors

image

### DIFF
--- a/src/Nancy/StaticConfiguration.cs
+++ b/src/Nancy/StaticConfiguration.cs
@@ -7,14 +7,14 @@ namespace Nancy
     using Nancy.Bootstrapper;
     using Nancy.Diagnostics;
 
-    public static class StaticConfiguration
+    public class StaticConfiguration
     {
-        private static bool? isRunningDebug;
-        private static bool? disableCaches;
+        private bool? isRunningDebug;
+        private bool? disableCaches;
 
-        private static bool? disableErrorTraces;
+        private bool? disableErrorTraces;
 
-        static StaticConfiguration()
+        public StaticConfiguration()
         {
             disableErrorTraces = !(disableCaches = IsRunningDebug);
             CaseSensitive = false;
@@ -26,7 +26,7 @@ namespace Nancy
         /// Gets or sets a value indicating whether or not to disable traces in error messages
         /// </summary>
         [Description("Disables trace output in the default 500 error pages.")]
-        public static bool DisableErrorTraces
+        public bool DisableErrorTraces
         {
             get
             {
@@ -42,26 +42,26 @@ namespace Nancy
         /// Gets or sets a value indicating whether or not to respond with 405 responses
         /// </summary>
         [Description("Disables 405 responses from being sent to the client.")]
-        public static bool DisableMethodNotAllowedResponses { get; set; }
+        public bool DisableMethodNotAllowedResponses { get; set; }
 
         /// <summary>
         /// Gets or sets a value indicating whether or not to enable case sensitivity in query, parameters (DynamicDictionary) and model binding. Enable this to conform with RFC3986.
         /// </summary>
         [Description("Enable case sensitivity in query, parameters (DynamicDictionary) and model binding. Enable this to conform with RFC3986.")]
-        public static bool CaseSensitive { get; set; }
+        public bool CaseSensitive { get; set; }
 
         /// <summary>
         /// Gets or sets a value indicating whether or not to route HEAD requests explicitly.
         /// </summary>
         [Description("Enables explicit HEAD routing and disables the usage of GET routes for HEAD requests.")]
-        public static bool EnableHeadRouting { get; set; }
+        public bool EnableHeadRouting { get; set; }
 
         /// <summary>
         /// Gets a value indicating whether we are running in debug mode or not.
         /// Checks the entry assembly to see whether it has been built in debug mode.
         /// If anything goes wrong it returns false.
         /// </summary>
-        public static bool IsRunningDebug
+        public bool IsRunningDebug
         {
             get
             {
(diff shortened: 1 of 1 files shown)


thecodejunkie:
I spiked out an approach for extracting `StaticConfiguration.IsDebugMode()`.

```c#
/// <summary>
/// Defines functionality for getting information about the runtime execution environment.
/// </summary>
public interface IRuntimeEnvironmentInformation
{
    /// <summary>
    /// Gets a value indicating if the application is running in debug mode.
    /// </summary>
    /// <returns><see langword="true"/> if the application is running in debug mode, otherwise <see langword="false"/>.</returns>
    bool IsDebug { get; }
}

/// <summary>
/// Default implementation of the <see cref="IRuntimeEnvironmentInformation"/> interface.
/// </summary>
public class DefaultRuntimeEnvironmentInformation : IRuntimeEnvironmentInformation
{
    private bool? isDebug;

    /// <summary>
    /// Gets a value indicating if the application is running in debug mode.
    /// </summary>
    /// <returns><see langword="true"/> if the application is running in debug mode, otherwise <see langword="false"/>.</returns>
    public virtual bool IsDebug
    {
        get { return this.isDebug ?? (bool)(this.isDebug = GetDebugMode()); }
    }

    private static bool GetDebugMode()
    {
        try
        {
            var assembliesInDebug = AppDomainAssemblyTypeScanner
                .TypesOf<INancyModule>(ScanMode.ExcludeNancy)
                .Select(x => x.Assembly.GetCustomAttributes(typeof(DebuggableAttribute), true))
                .Where(x => x.Length != 0);

            return assembliesInDebug.Any(d => ((DebuggableAttribute)d[0]).IsJITTrackingEnabled);
        }
        catch
        {
            return false;
        }
    }
}
```

I consider this to be `Phase 1` of this implementation. `Phase 2` would happen after the scanning has moved away from the static `AppDomainAssemblyTypeScanner` class.

khellang:
I'd rather just make it a `Lazy<bool>` than doing `return this.isDebug ?? (bool)(this.isDebug = GetDebugMode());`.

thecodejunkie:
@khellang agreed. Changed the comment to reflect this

thecodejunkie:
Ping @NancyFx/most-valued-minions @NancyFx/nancy-core-contributors need feedback / suggestions on how to split up `StaticConfiguration` :heart:

jchannon:
Excuse thenaivety but what does making in Lazy give you?

I agree on the naming making everything `EnableX`

khellang:
What about `UseX`?

thecodejunkie:
@jchannon I think I'd like to convert them one-by-one myself because it's going to be easier. What I need help with is suggestions on how to logically split them into different configs + names of said configs

thecodejunkie:
@jchannon `Lazy<bool>` gives us the same functionality with both less code and with less complexity.

khellang:
Ok, `CaseSensitive` is used for all kinds of things. I think it's worth having a `CaseSensitive` property for each area where it's used, so you get more granular control.

The areas I see it's being used right now are
 - Routing
 - Tracing
 - Query string parsing
 - DynamicDictionary
 - Form parsing

It's used some other places as well, but I think we can manage to kill those. I'm not sure they make sense.

khellang:
I think a pattern is emerging:

 - RoutingConfiguration
 - TracingConfiguration
 - FileUploadConfiguration
 - ViewConfiguration

From my comment above, I think form, query and dynamic dictionary are closely related, maybe they could have a different configuration?

phillip-haydon:
I don't really understand why there's an upload file config.

khellang:
@phillip-haydon Because the current `StaticConfiguration` has 3 properties for configuring file upload:

 - RequestQueryFormMultipartLimit
 - DisableRequestStreamSwitching
 - AllowFileStreamUploadAsync

phillip-haydon:
Yes but I don't understand what they are for. Really (IMO) the only file upload setting is limiting the request size which IMO should be done in middleware before it ever reaches Nancy. The rest are pointless.

thecodejunkie:
Moving this to `2.1` as it's not a PR that will be pulled in anymore, but serves as a reminder that we still have a couple of static configs to migrate. Most of the old members have been converted to the new configuration system in other pull-requests

thecodejunkie:
Closing this.. we can revisit if/when we decide to remove the remaining stuff out

PR #2127, opened 2015-11-23
NullReferenceException in pipeline

asbjornu:
If a pipeline method returning `Task` for some reason returns `null`, `Nancy.AfterPipeline.ExecuteTasksInternal()` blows up with a `NullReferenceException`. This PR fixes this by instead throwing an informative `InvalidOperationException` as well as a couple of tests ensuring the expected behavior.

### DIFF
--- a/src/Nancy.Tests/Unit/AfterPipelineFixture.cs
+++ b/src/Nancy.Tests/Unit/AfterPipelineFixture.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Linq;
     using System.Threading;
+    using System.Threading.Tasks;
 
     using Xunit;
 
@@ -101,6 +102,37 @@
             Assert.True(item2Called);
             Assert.True(item3Called);
             Assert.True(item4Called);
+        }
+
+        [Fact]
+        public void Pipeline_containing_method_returning_null_throws_InvalidOperationException()
+        {
+            pipeline.AddItemToEndOfPipeline(ReturnNull);
+
+            var exception = Assert.Throws<InvalidOperationException>(() =>
+            {
+                pipeline.Invoke(CreateContext(), new CancellationToken());
+            });
+
+            Assert.Equal("An after-pipeline action must not return null, a Task was expected.", exception.Message);
+        }
+
+        [Fact]
+        public void Pipeline_containing_lambda_returning_null_throws_InvalidOperationException()
+        {
+            pipeline.AddItemToEndOfPipeline((context, ct) => null);
+
+            var exception = Assert.Throws<InvalidOperationException>(() =>
+            {
+                pipeline.Invoke(CreateContext(), new CancellationToken());
+            });
+
+            Assert.Equal("An after-pipeline action must not return null, a Task was expected.", exception.Message);
+        }
+        
+        private static Task ReturnNull(NancyContext context, CancellationToken ct)
+        {
+            return null;
         }
     }
 }
--- a/src/Nancy/AfterPipeline.cs
+++ b/src/Nancy/AfterPipeline.cs
@@ -83,6 +83,11 @@
             {
                 var current = enumerator.Current.Invoke(context, cancellationToken);
 
+                if (current == null)
+                {
+                    throw new InvalidOperationException("An after-pipeline action must not return null, a Task was expected.");
+                }
+
                 if (current.Status == TaskStatus.Created)
                 {
                     current.Start();


src/Nancy.Tests/Unit/AfterPipelineFixture.cs
```diff
@@ -102,5 +103,40 @@ public void Pipeline_containing_another_pipeline_will_invoke_items_in_both_pipel
             Assert.True(item3Called);
             Assert.True(item4Called);
         }
+
+        [Fact]
+        public void Pipeline_containing_method_returning_null_from_method_throws_InvalidOperationException()
+        {
+            pipeline.AddItemToEndOfPipeline(ReturnNull);
+
+
```
khellang: :fire:

src/Nancy.Tests/Unit/AfterPipelineFixture.cs
```diff
@@ -102,5 +103,40 @@ public void Pipeline_containing_another_pipeline_will_invoke_items_in_both_pipel
...
+        {
+            pipeline.AddItemToEndOfPipeline(ReturnNull);
+
+
+            var exception = Assert.Throws<InvalidOperationException>(() =>
+            {
+                pipeline.Invoke(CreateContext(), new CancellationToken());
+            });
+
+            Console.WriteLine(exception);
```
khellang: :fire:

src/Nancy.Tests/Unit/AfterPipelineFixture.cs
```diff
@@ -102,5 +103,40 @@ public void Pipeline_containing_another_pipeline_will_invoke_items_in_both_pipel
...
+        public void Pipeline_containing_method_returning_null_from_lambda_throws_InvalidOperationException()
+        {
+            pipeline.AddItemToEndOfPipeline((context, ct) => null);
+
+            var exception = Assert.Throws<InvalidOperationException>(() =>
+            {
+                pipeline.Invoke(CreateContext(), new CancellationToken());
+            });
+
+            Console.WriteLine(exception);
```
khellang: :fire:

src/Nancy.Tests/Unit/AfterPipelineFixture.cs
```diff
@@ -102,5 +103,40 @@ public void Pipeline_containing_another_pipeline_will_invoke_items_in_both_pipel
...
+        }
+
+
```
khellang: :fire:

src/Nancy/AfterPipeline.cs
```diff
@@ -83,6 +83,12 @@ private static void ExecuteTasksInternal(NancyContext context, CancellationToken
             {
                 var current = enumerator.Current.Invoke(context, cancellationToken);
 
+                if (current == null)
+                {
+                    var method = enumerator.Current.Method ?? (object)enumerator.Current;
+                    throw new InvalidOperationException(string.Format("'{0}' returned null; a non-null {1} expected.", method, typeof(Task)));
```
khellang: Do you have an example of this error message? Both for the `Task.Method` and the `Task` itself. Something's telling me that this won't be very helpful :confused:

asbjornu: Yes, it's being written to the console in the accompanying tests. It is a lot more useful than the currently thrown `NullReferenceException`, at least.

src/Nancy.Tests/Unit/AfterPipelineFixture.cs
asbjornu: You don't want the `Console.WriteLine()`?

khellang: Convince me why it should stay :stuck_out_tongue_winking_eye:

src/Nancy/AfterPipeline.cs
khellang: That doesn't help me when reviewing this PR :stuck_out_tongue:

khellang: Found it in the CI build log, and as I thought - not very nice :cry: 

> 'System.Threading.Tasks.Task <Pipeline_containing_method_returning_null_from_lambda_throws_InvalidOperationException>b__1a(Nancy.NancyContext, System.Threading.CancellationToken)' returned null; a non-null System.Threading.Tasks.Task expected.

I'd rather remove the parameters and give a hard coded message.

src/Nancy.Tests/Unit/AfterPipelineFixture.cs
asbjornu: The point of this PR is to improve the exception being thrown when a pipeline method returns `null`. Writing the exception to the `Console` is the only way I can think of to actually see the caught exception.

asbjornu:
@khellang A last attempt at formatting the method name in a user friendly way before removing it. This is the output from `Pipeline_containing_lambda_returning_null_throws_InvalidOperationException`:

```c#
System.InvalidOperationException: The pipeline method '(context, ct) => Task' returned null; a non-null System.Threading.Tasks.Task expected.
```

And this is from `Pipeline_containing_method_returning_null_throws_InvalidOperationException`:

```c#
System.InvalidOperationException: The pipeline method 'Task AfterPipelineFixture.ReturnNull(context, ct)' returned null; a non-null System.Threading.Tasks.Task expected.
```

khellang:
@asbjornu I'm not a fan. I still think we're doing way to much (reflection) work for a message that's still not super helpful. Would you mind just hard coding it to `An after-pipeline action must not return null, a Task was expected.` or something? And kill the `Console.WriteLine` while you're at it. We're moving to xUnit 2.x very soon, and that won't capture the output by default.

asbjornu:
@khellang Ok, I'll remove it. I think that the current exceptions are super useful and I would love having them thrown in my face. Not having the name of the method makes debugging much harder, especially in production environments where line numbers aren't available.

khellang:
@asbjornu `(context, ct) => Task`, isn't very helpful to identify the method, is it? I agree that `Task AfterPipelineFixture.ReturnNull(context, ct)` is better, but it's still too much fuzz. What if it just contained just `method.Name` (given the name is sensible)?

khellang:
Also, a squash is probably in order here :wink:

asbjornu:
@khellang Squashed. I kept the constant message. Although I agree `method.Name` would be useful, I don't have more time to spend on this at the moment. Perhaps in a future PR. :)

khellang:
@asbjornu Thanks for this! Yes, a future PR would be welcome :+1:

Issue #2113, opened 2015-11-15
Nancy 1.4.1 memory leak


cusnirati:
While testing for high-load, I discovered a memory leak issue directly linked to either Nancy or Nancy.Hosting.Self .
Downgrading to the 1.2.0 version solves this.
I uploaded a small demo for this : https://github.com/cusnirati/Nancy1.4.1MemoryLeakDemo

I tested this using JMeter, the project for repeating my steps is inside the solution linked above.

![screenshot 2015-11-15 22 28 26](https://cloud.githubusercontent.com/assets/10377229/11171623/ab90f5a0-8bec-11e5-9856-9bdeb8ddd7f3.png)
![screenshot 2015-11-15 22 29 04](https://cloud.githubusercontent.com/assets/10377229/11171622/ab901bd0-8bec-11e5-87ee-fe9d513c53dd.png)
![screenshot 2015-11-15 22 29 41](https://cloud.githubusercontent.com/assets/10377229/11171621/ab9007f8-8bec-11e5-9e2a-e94e9589d364.png)
![screenshot 2015-11-15 22 30 13](https://cloud.githubusercontent.com/assets/10377229/11171620/ab8eaffc-8bec-11e5-82c4-6d8bd6eefe47.png)
![screenshot 2015-11-15 22 30 52](https://cloud.githubusercontent.com/assets/10377229/11171624/ab9151e4-8bec-11e5-9b1c-2f050c5ea581.png)

khellang:
It would be nice to narrow down the version even more. Are you seeing the same on 1.3?

cusnirati:
For 1.3.0 everything looks good, no leaks.
Let me check for the 1.4.1.
I just mentioned the 1.2.0 because that was my previous working version.

khellang:
OK, I've identified the regressing commit; https://github.com/NancyFx/Nancy/commit/4aeaf318e9710f9a0c3ac00befb35ee109ec82a2.

It introduced the `CancellationTokenSource.CreateLinkedTokenSource` call which allocates per request, without disposing it afterwards.

![image](https://cloud.githubusercontent.com/assets/582487/11171758/8abc1b64-8bf8-11e5-8218-ed7e379d0f88.png)

![image](https://cloud.githubusercontent.com/assets/582487/11171766/a9337f1a-8bf8-11e5-845d-64bf69e91e63.png)

I'll send a PR for it. Thanks for filing this! :+1:

cusnirati:
No problem :)

[...]

PR #2128, opened 2015-11-23
Added test for memory leaks across requests

khellang:
This should easily detect cases like #2113 in the future.

The only thing left is to hook up dotMemoryUnit.exe to the build script, but my Rake-fu is lacking, so I punted on it until we decide if this is something we want.

So... Do we want/need this? :smile:


jchannon:
nifty!

PR #2132, opened 2015-11-24
Add the method name to the InvalidOperationException

asbjornu:
This PR adds the method name to the `InvalidOperationException` thrown when a pipeline action returns null, and use semicolon instead of comma to separate the clauses in the message.


khellang:
:grin::+1:

Issue #2131, opened 2015-11-24
Empty negotiated response with accept header of application/xml or text/xml throws serialization exception


phoenixx:
Given a simple model which simply returns a 200 response:

```c#
 public class DemoModule : NancyModule
    {
        public DemoModule() : base("/demo")
        {
            Get["ok", "/"] =
                _ => Negotiate.WithStatusCode(HttpStatusCode.OK);
        }
    }
```

Making a GET request to this endpoint with an accept header of "application/json" returns an empty response with a status code of 200, as expected. 

Making a GET request to this endpoint with an accept header of "text/xml" or "application/xml" returns a 500 internal server error with the following exception:

````
The type arguments for method 'Nancy.ISerializer.Serialize<TModel>(string, TModel, System.IO.Stream)' cannot be inferred from the usage. Try specifying the type arguments explicitly.
````

khellang:
Closed by #2133

khellang:
Version 1.4.3 was just pushed to [NuGet](https://www.nuget.org/packages/Nancy/1.4.3).

PR #2133, opened 2015-11-24
Fixes issue #2131 with XmlProcessor

phoenixx:
When the model passed to CreateResponse is null don't try to serialize
it as XML, effectively pass an empty body.


khellang:
Awesome! Thanks @phoenixx! :heart:

khellang:

PR #2139, opened 2015-12-02
Added IRuntimeEnvironmentInformation

thecodejunkie:
The idea for `IRuntimeEnvironmentInformation` is to give us a place to stick information about the runtime environment. Currently it contains support for knowing if the application is being run in debug or not. 

This is the first step for splitting up the `StaticConfiguration` class. It will be rewritten once we change over to instance based scanning.

**PS** This does not remove the same functionality in `StaticConfiguration` - that will happen one at a time when `StaticConfiguration` is split up, but we need this in place first.

### DIFF
--- a/src/Nancy/DefaultRuntimeEnvironmentInformation.cs
+++ b/src/Nancy/DefaultRuntimeEnvironmentInformation.cs
@@ -10,7 +10,7 @@ namespace Nancy
     /// </summary>
     public class DefaultRuntimeEnvironmentInformation : IRuntimeEnvironmentInformation
     {
-        private readonly Lazy<bool> isDebug = GetDebugMode();
+        private readonly Lazy<bool> isDebug = new Lazy<bool>(GetDebugMode);
 
         /// <summary>
         /// Gets a value indicating if the application is running in debug mode.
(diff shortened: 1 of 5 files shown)


src/Nancy/DefaultRuntimeEnvironmentInformation.cs
```diff
@@ -0,0 +1,44 @@
...
+    using System.Diagnostics;
+    using System.Linq;
+    using Nancy.Bootstrapper;
+
+    /// <summary>
+    /// Default implementation of the <see cref="IRuntimeEnvironmentInformation"/> interface.
+    /// </summary>
+    public class DefaultRuntimeEnvironmentInformation : IRuntimeEnvironmentInformation
+    {
+        private readonly Lazy<bool> isDebug = GetDebugMode();
```
khellang: `new Lazy<bool>(GetDebugMode)` and kill the nesting within the method?

src/Nancy/DefaultRuntimeEnvironmentInformation.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        private static Lazy<bool> GetDebugMode()
+        {
+            return new Lazy<bool>(() =>
+            {
+                try
+                {
+                    var assembliesInDebug = AppDomainAssemblyTypeScanner
+                        .TypesOf<INancyModule>(ScanMode.ExcludeNancy)
+                        .Select(x => x.Assembly.GetCustomAttributes(typeof(DebuggableAttribute), true))
```
khellang: `IsDefined<DebuggableAttribute>(inherit: true)`?

khellang: Oh, wait, you need the attribute below. What about `GetCustomAttribute<DebuggableAttribute>(inherit: true)` then?

khellang:
LGTM. Just a couple of nits :wink:

thecodejunkie:
Updated the init of `Lazy<bool>` and left the rest. Basically the code in `GetDebugMode` is a straight copy from the code in `StaticConfiguration` and I decided to leave it as is because it will be rewritten once we move over to instance based scanning

khellang:
:+1:

PR #2149, opened 2015-12-07
Keep existing reason phrase in DefaultStatusCodeHandler

### LINKED ISSUE #2148: ReasonPhrase is discarded on application/json processsor
If you have a route such as:

    Get["/notfound"] = parameters =>
    {
        return new NotFoundResponse
        {
            ReasonPhrase = "foobar",
        };
    };

And you call the API with accept header `application/json`, then "ReasonPhrase" will be `Not Found` instead of `foobar`.

If accept header is `text/html`, then "ReasonPhrase" will be `foobar` (as expected).

This behavior has changed in some version between Nancy 0.22.2 and 1.4.2.

I have tried looking into the code, but could not find a solution in my available time frame.

khellang:
Currently the `DefaultStatusCodeHandler` discards the `ReasonPhrase` property of a potential existing response.

Fixes #2148.

### DIFF
--- a/src/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
+++ b/src/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
@@ -159,7 +159,7 @@ namespace Nancy.Tests.Functional.Tests
 
                     var negotiator =
                         new Negotiator(context);
-                    negotiator.WithReasonPhrase("The test is passing!");
+                    negotiator.WithReasonPhrase("The test is passing!").WithStatusCode(404);
 
                     return negotiator;
                 });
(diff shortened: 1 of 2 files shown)


src/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
```diff
@@ -167,13 +167,13 @@ public void Should_set_reason_phrase_on_response()
...
             {
+                with.StatusCodeHandler<DefaultStatusCodeHandler>();
                 with.ResponseProcessor<TestProcessor>();
-
                 with.Module(module);
             });
 
             // When
-            var response = brower.Get("/customPhrase");
+            var response = brower.Get("/customPhrase", with => with.Accept("application/json"));
```
damianh: wee typo 'brower'

damianh: oh wait it was there previously

damianh:
Change looks good, typo can be fixed later.

jchannon:
I'm actually not sure about this PR. A phrase is usually tied to a status code? So if you return some data with a phrase of "here's some awesome data" but then it throws, having it changed to 500 but with a phrase that relates to something that was meant for when there was no exception doesn't make sense?

damianh:
Anything that sets the status code (i.e. error handler) should also set the reason phrase

jchannon:
Agree but there will be a mismatch?

khellang:

Issue #2141, opened 2015-12-03
CancellationToken Disposed before async request is complete


jeffthiele-iq:
After updating to 1.4.2, I started getting `ObjectDisposedException: The CancellationTokenSource has been disposed.` when calling async methods using the `CancellationToken` provided by Nancy.  In my case it was EntityFramework methods, but I believe it would happen with any async methods.

The error does not exist in 1.4.1, and I think I tracked it down to the changes in pull request #2114.  It might be because the `lifeCycleTask` task isn't allowed to complete before the the token is disposed.

khellang:
> The error does not exist in 1.4.1

That's because the CTS wasn't disposed *at all*, so we had a nice 40 byte memory leak per request :stuck_out_tongue:

I'll take a look at it. @damianh, feel free to beat me to it :wink:

khellang:
@jeffthiele-iq Thanks for filing and digging into this, BTW! :+1:

jeffthiele-iq:
Here's a test module which recreates the issue with only file access.  The test file is just the text for War and Peace to slow down the file read enough that the token gets disposed part-way through.
```
using System;
using System.IO;
using System.Reflection;
using System.Threading;
using System.Threading.Tasks;
using System.Web;
using Nancy;

namespace NancyAsyncTesting
{
	public class TestModule : NancyModule
	{
		public TestModule()
		{
			Get["/test", true] = TokenTest;
		}

		private async Task<dynamic> TokenTest(dynamic o, CancellationToken cancellationToken)
		{
			using (var memoryStream = new MemoryStream())
			using (var dataFile = new FileStream(Path.Combine("C:/code", "TestFile.txt"), FileMode.Open, FileAccess.Read, FileShare.Read, 4096, true))
			{
				await dataFile.CopyToAsync(memoryStream, 4096, cancellationToken);
			}

			return "test";
		}
	}
}
```

damianh:
Fixed via #2150

damianh:
Personal note: will need to reimplement on the async branch

khellang:
Version 1.4.3 was just pushed to [NuGet](https://www.nuget.org/packages/Nancy/1.4.3).

PR #2150, opened 2015-12-07
Change disposal of CancellationTokenSource to after task is complete

jeffthiele-iq:
The previous implementation was allowing the disposal of the `CancellationTokenSource` to occur before the `lifeCycleTask` was actually complete.  I moved the disposal to the `completeTask` continuation so it would only be disposed when the task was complete.

Also included is a test that can detect this in the future and also verifies that the token is disposed after the call.

I also ran through the memory leak test in pull request #2128 and (assuming I had my CLI commands correct for dotMemoryUnit) it passed.

### DIFF
--- a/src/Nancy/NancyEngine.cs
+++ b/src/Nancy/NancyEngine.cs
@@ -101,61 +101,63 @@
         /// <returns>The task object representing the asynchronous operation.</returns>
         public Task<NancyContext> HandleRequest(Request request, Func<NancyContext, NancyContext> preRequest, CancellationToken cancellationToken)
         {
-            using (var cts = CancellationTokenSource.CreateLinkedTokenSource(this.engineDisposedCts.Token, cancellationToken))
+	        var cts = CancellationTokenSource.CreateLinkedTokenSource(this.engineDisposedCts.Token, cancellationToken);
+            cts.Token.ThrowIfCancellationRequested();
+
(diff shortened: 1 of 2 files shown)


jeffthiele-iq:
I forgot to mention, this is to fix issue #2141

src/Nancy/NancyEngine.cs
```diff
@@ -101,61 +101,63 @@ public class NancyEngine : INancyEngine
...
+		                this.SaveTraceInformation(completeTask.Result);
+	                }
+	                catch (Exception ex)
+	                {
+		                tcs.SetException(ex);
+		                return;
+	                }
+	                finally
+	                {
+		                cts.Dispose();
```
damianh: Nice :+1:

damianh:
There is a PR where the async await stuff is cleaned up for .net 4.5 : https://github.com/NancyFx/Nancy/pull/2105/files#diff-3711e290f8f06e8bd6ce86e1e6ce7dbfR102 

I don't think I need to do any special here?

jeffthiele-iq:
The async/await implementation looks correct to me too

damianh:
Thanks @jeffthiele-iq

thecodejunkie:
@jeffthiele-iq appreciate your feedback on `async/await` !

khellang:

PR #2154, opened 2015-12-08
Restructured repository

khellang:
This'll ease the trasition to project.json. It also makes it easier to separate production/test/sample projects.

If you want to take a look at the new structure, see https://github.com/khellang/Nancy/tree/restructure


jchannon:
LGTM

khellang:
Same with this. Not worth keeping rebased.

PR #2158, opened 2015-12-09
Fixes #2157 with RequestStream.AsString extension modifying stream's position

### LINKED ISSUE #2157: RequestStream's AsString() extension modifies stream position
If you use the AsString() extension to get the textual representation of a request's body...

```
Console.WriteLine(this.Request.Body.AsString());
```

...the RequestStream position gets left at the end of the stream because the StreamReader being used by the extension reads the stream until the end...

```
public static string AsString(this RequestStream stream, Encoding encoding = null)
{
    using (var reader = new StreamReader(stream, encoding ?? Encoding.UTF8))
    {
        return reader.ReadToEnd();
    }
}
```

...but doesn't update the position back to the initial one.

It can get confusing if you try to read again from Request.Body afterwards. A solution might be resetting the stream's position before returning.

aryelfnd:
Changes RequestStream.AsString extension method to not modify the stream's position.

Fixes #2157


jchannon:
LGTM. @NancyFx/most-valued-minions

PR #2159, opened 2015-12-10
SwitchoverStream does not work when %TEMP% is full

danbarua:
I'm not sure if this is of much value as I suspect we're in the minority of users to run into this issue - feel free to ignore and close.

The award for most asinine error message surely goes to ```Path.GetTempFileName()``` throwing "The file exists." when %TEMP% has more than 65k items in it.

Catching the exception and wrapping it in one with a useful error message.

### DIFF
--- a/src/Nancy/IO/RequestStream.cs
+++ b/src/Nancy/IO/RequestStream.cs
@@ -358,9 +358,21 @@
 
         private static FileStream CreateTemporaryFileStream()
         {
-            var filePath = Path.GetTempFileName();
+            // we could use Path.GetTempFilePath() but this is problematic on Windows and more so on Mono
+            // symptoms will show when this method has been called > 65k times
+            // see docs: https://msdn.microsoft.com/en-us/library/system.io.path.gettempfilename(v=vs.110).aspx
+            // comments on Win32 implementation: https://msdn.microsoft.com/en-us/library/windows/desktop/aa364991(v=vs.85).aspx
+            // mono implementation: https://github.com/mono/mono/blob/master/mcs/class/corlib/System.IO/Path.cs#L490
 
-            return new FileStream(filePath, FileMode.Create, FileAccess.ReadWrite, FileShare.None, 8192, StaticConfiguration.AllowFileStreamUploadAsync);
+            var filePath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("N") + ".tmp");
+
+            return new FileStream(
+                filePath,
+                FileMode.Create,
+                FileAccess.ReadWrite,
+                FileShare.None,
+                8192,
+                StaticConfiguration.AllowFileStreamUploadAsync);         
         }
 
         private Stream CreateDefaultMemoryStream(long expectedLength)


src/Nancy/IO/RequestStream.cs
```diff
@@ -358,9 +358,28 @@ private void ShiftStreamToFileStreamIfNecessary()
...
+                    filePath,
+                    FileMode.Create,
+                    FileAccess.ReadWrite,
+                    FileShare.None,
+                    8192,
+                    StaticConfiguration.AllowFileStreamUploadAsync);
+            }
+            catch (System.IO.IOException ex)
+            {
+                if (ex.Message == "The file exists.")
```
thecodejunkie: Probably break this out to a const and also use a case-insensitive string comparison

danbarua:
It turns out we don't need to check the exception message, so I removed it.

https://msdn.microsoft.com/en-us/library/system.io.path.gettempfilename(v=vs.110).aspx
https://github.com/mono/mono/blob/master/mcs/class/corlib/System.IO/Path.cs#L490

src/Nancy/IO/RequestStream.cs
```diff
@@ -358,9 +358,23 @@ private void ShiftStreamToFileStreamIfNecessary()
...
+
+                return new FileStream(
+                    filePath,
+                    FileMode.Create,
+                    FileAccess.ReadWrite,
+                    FileShare.None,
+                    8192,
+                    StaticConfiguration.AllowFileStreamUploadAsync);
+            }
+            catch (System.IO.IOException ex)
```
jchannon: Remove the fully qualified exception name :smile:

danbarua:
Nancy aside, wtf? If I call ```Path.GetTempFileName()``` 65k times, then call it again and get a name collision it will chuck an IOException? Lame.

jchannon:
LGTM

danbarua:
Hmm maybe we get rid of ```Path.GetTempFileName``` as it's going to crumple under heavy load: https://msdn.microsoft.com/en-us/library/windows/desktop/aa364991(v=vs.85).aspx

danbarua:
Ignore my last comment, you'd have to be hammering Nancy with >65k concurrent requests >85kb in size to hit that problem.

src/Nancy/IO/RequestStream.cs
```diff
@@ -358,9 +358,23 @@ private void ShiftStreamToFileStreamIfNecessary()
...
+
+                return new FileStream(
+                    filePath,
+                    FileMode.Create,
+                    FileAccess.ReadWrite,
+                    FileShare.None,
+                    8192,
+                    StaticConfiguration.AllowFileStreamUploadAsync);
+            }
+            catch (IOException ex)
```
jchannon: Shouldn't this just be around `GetTempFilename`?

src/Nancy/IO/RequestStream.cs
```diff
@@ -358,9 +358,23 @@ private void ShiftStreamToFileStreamIfNecessary()
...
+            {
+				throw new IOException(
+                        "Unable to create temporary file, the %TEMP% folder has more than 65535 items. Try clearing out the folder.", ex);
```
damianh: Can the `IOException` be thrown for reasons _other_ than going over the 65535 items limit? Would this obscure those reasons?

src/Nancy/IO/RequestStream.cs
danbarua: I'd have to declare a FileStream variable outside the try{} block and then return it after, this is, imo, cleaner.

Even better would be 
```
return new FileStream(Path.GetTempFileName(), FileMode.Create, FileAccess.ReadWrite, FileShare.None, 8192, StaticConfiguration.AllowFileStreamUploadAsync);
```

src/Nancy/IO/RequestStream.cs
danbarua: Actually, yes on Mono that might be the case. We can inspect the HResult on Mono.

damianh: https://msdn.microsoft.com/en-us/library/system.io.path.gettempfilename(v=vs.110).aspx

"The GetTempFileName method will raise an IOException if no unique temporary file name is available. To resolve this error, delete all unneeded temporary files."

danbarua:
Since ```Path.GetTempFileName()``` seems to be problematic on Windows and even more so on Mono, I've voted to replace it entirely.

src/Nancy/IO/RequestStream.cs
```diff
@@ -358,9 +358,15 @@ private void ShiftStreamToFileStreamIfNecessary()
 
         private static FileStream CreateTemporaryFileStream()
         {
-            var filePath = Path.GetTempFileName();
-
-            return new FileStream(filePath, FileMode.Create, FileAccess.ReadWrite, FileShare.None, 8192, StaticConfiguration.AllowFileStreamUploadAsync);
+			var filePath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("N") + ".tmp");
```
khellang: Can we just add a comment stating why we're doing this? Otherwise, you can bet someone comes along a year from now with a PR saying "why don't you just use `Path.GetTempFileName` and no one remembers :stuck_out_tongue:

khellang:
LGTM modulo a comment and a squash :+1:

danbarua:
Cool, commented and squashed :+1:

jchannon:
Cheers big ears!

khellang:

PR #2163, opened 2015-12-14
[WIP] Simple json spike

jasonmead:
Definitely not complete. Need some guidance on how to continue.


khellang:
> Need some guidance on how to continue.

What specifically do you need guidance on? What are the issues?

phillip-haydon:
@jasonmead 

In regard to your 4 points

1. Don't care about circular dependencies, anyone running into that issue are doing it wrong. People run into that returning EF models directly.
2. We have no care for WCF so we can rip that crap out and support the dates properly.
3. Again like circular dependencies, recursion depth isn't our responsibility, people should return small concise models.
4. This is the only concern, but this should be relatively easy to fix with a static cache or something along those lines.

thecodejunkie:
@jasonmead Any more assistance that we can provide on this spike?

jasonmead:
@thecodejunkie No. Just got sidetracked with life. I'll revisit and update this weekend.

jasonmead:
I was able to get this updated tonight. As far as the issues are concerned I address each of them the following ways:

1. Remove circular reference tests
2. Removed WCF DateTime tests
3. Recursion limit tests were either removed or are no longer failing since I originally made the change.
4. Made some internal changes to SimpleJson to make the Constructor and Property caches static. They were already thread safe. These are internal to SimpleJson.PocoSerializationStrategy so should not impact the work to remove statics from Nancy's public interfaces.

jchannon:
@jasonmead do you consider this ready?

jchannon:
@jasonmead *bump* :smile:

Yantrio:
@jasonmead bumpity bump :)

jasonmead:
@jchannon I believe that it is. Had some merge conflicts trying to pull in latest master but I think I got it. Closing this spike PR and opened a clean PR: #2313

PR #2172, opened 2015-12-22
Added Status Code: HTTP 451 Unavailable For Legal Reasons

06b:
This status code indicates that the server is denying access to the resource as a consequence of a legal demand.

More Details: https://www.mnot.net/blog/2015/12/18/451


phillip-haydon:
Beaten to it :) was reading about 451 this morning and figured I would send a PR.

PR #2173, opened 2015-12-23
Don't swallow exception in stateless auth

### LINKED ISSUE #2112: Stateless Authentication - Swallows User Validator Exceptions
Hi,

I'm having intermittent exceptions from the user validator and would have expected those to have bubbled through the pipeline to the internal server error handler but instead the exceptions are swallowed and the request ends up on the unauthorised handler as if everything is fine but authorisation failed.

I could be missing something so, is that the intended behaviour? 

Thanks,

Joao

khellang:
Fixes #2112

### DIFF
--- a/src/Nancy.Authentication.Stateless/StatelessAuthentication.cs
+++ b/src/Nancy.Authentication.Stateless/StatelessAuthentication.cs
@@ -65,7 +65,7 @@ namespace Nancy.Authentication.Stateless
         /// </summary>
         /// <param name="configuration">Stateless authentication configuration to use</param>
         /// <returns>Pre request hook delegate</returns>
-        static Func<NancyContext, Response> GetLoadAuthenticationHook(StatelessAuthenticationConfiguration configuration)
+        private static Func<NancyContext, Response> GetLoadAuthenticationHook(StatelessAuthenticationConfiguration configuration)
         {
             if (configuration == null)
             {
@@ -74,15 +74,9 @@ namespace Nancy.Authentication.Stateless
 
             return context =>
             {
-                try
-                {
-                    context.CurrentUser = configuration.GetUserIdentity(context);
-                    return context.Response;
-                }
-                catch (Exception)
-                {
-                    return context.Response;
-                }
+                context.CurrentUser = configuration.GetUserIdentity(context);
+
+                return null;
             };
         }
     }


src/Nancy.Authentication.Stateless/StatelessAuthentication.cs
```diff
@@ -74,16 +74,10 @@ public static void Enable(INancyModule module, StatelessAuthenticationConfigurat
...
-                    context.CurrentUser = configuration.GetUserIdentity(context);
-                    return context.Response;
-                }
-                catch (Exception)
-                {
-                    return context.Response;
-                }
+                context.CurrentUser = configuration.GetUserIdentity(context);
+
+                return null;
```
bsommardahl: Why return null here?

khellang: Because if you don't return `null` in a before-pipeline-hook, you short-circuit the pipeline and return the response directly.

khellang: You can see [here](https://github.com/NancyFx/Nancy/blob/665a59a23ce3df74672f3334bf653947151f11a3/src/Nancy/NancyEngine.cs#L239-L240), that if the pre-request hooks return a response, the request dispatcher (the whole module pipeline) will never be called. Further, you can see in the [BeforePipeline](https://github.com/NancyFx/Nancy/blob/665a59a23ce3df74672f3334bf653947151f11a3/src/Nancy/BeforePipeline.cs#L56-L60) that it short-circuits if the response is non-null.

khellang: BTW, this is what the other auths do as well:
 - [Basic](https://github.com/NancyFx/Nancy/blob/665a59a23ce3df74672f3334bf653947151f11a3/src/Nancy.Authentication.Basic/BasicAuthentication.cs#L76)
 - [Forms](https://github.com/NancyFx/Nancy/blob/665a59a23ce3df74672f3334bf653947151f11a3/src/Nancy.Authentication.Forms/FormsAuthentication.cs#L211)
 - [Token](https://github.com/NancyFx/Nancy/blob/d2d0240327efe8875025c4a3c68c093ac9bfb877/src/Nancy.Authentication.Token/TokenAuthentication.cs#L71)

bsommardahl: Cool. I was unaware. Maybe we can improve the framework in the future where we done have to give special meaning to "null". It worries me any time I see it used.

khellang: +:100: from me there :smile:

PR #2184, opened 2016-01-09
Migrated safe path configuration to new config API

jchannon:


src/Nancy/RootPathApplicationStartup.cs
```diff
@@ -9,13 +10,16 @@ namespace Nancy
...
+        private readonly INancyEnvironment envronment;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="RootPathApplicationStartup"/> class.
         /// </summary>
         /// <param name="rootPathProvider">An <see cref="IRootPathProvider"/> instance.</param>
         public RootPathApplicationStartup(IRootPathProvider rootPathProvider)
         {
-            GenericFileResponse.SafePaths.Add(rootPathProvider.GetRootPath());
+            envronment = new DefaultNancyEnvironment();
```
khellang: WTF?

src/Nancy/SafePathConfigurationExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// Contains <see cref="SafePathConfiguration"/> configuration for <see cref="INancyEnvironment"/>.  
+    /// </summary>
+    public static class SafePathConfigurationExtensions
+    {
+        /// <summary>
+        /// Configures <see cref="SafePathConfiguration"/> 
+        /// </summary>
+        /// <param name="environment"><see cref="INancyEnvironment"/> that should be configured.</param>
+        /// <param name="paths">Safe paths.</param>
+        public static void Paths(this INancyEnvironment environment, IEnumerable<string> paths = null)
```
khellang: `SafePaths`? Why is `paths` optional? Does it make sense to call this method without them?

src/Nancy/SafePathConfiguration.cs
```diff
@@ -0,0 +1,34 @@
...
+        public SafePathConfiguration(IEnumerable<string> paths)
+        {
+            this.Paths = paths ?? Default.Paths;
+        }
+
+        /// <summary>
+        /// Gets or sets the paths.
+        /// </summary>
+        /// <value>The paths.</value>
+        public IEnumerable<string> Paths{ get; set; }
```
khellang: Nit: space before `{ get; set; }`

src/Nancy/SafePathConfiguration.cs
```diff
@@ -0,0 +1,34 @@
...
+    }
+
```
khellang: :fire:

khellang:
I'm not really sure why this requires its own configuration... :confused:

src/Nancy/RootPathApplicationStartup.cs
jchannon: It needs to be injected I think

src/Nancy/SafePathConfigurationExtensions.cs
jchannon: just mirroring what xmlsettings did

jchannon:
Just taking it from the list here https://github.com/NancyFx/Nancy/issues/2000

src/Nancy/RootPathApplicationStartup.cs
thecodejunkie: You can take ctor dependency on `INancyEnvironment`

jchannon:
Waiting on #2186 for the environment to be on the context before this can be finished

src/Nancy/SafePathConfigurationExtensions.cs
jchannon: I also can't see how this is used, same for Views in `ViewConfigurationExtensions`

jchannon:
no longer WIP, go nuts

src/Nancy.Tests/Unit/ResponseExtensionsFixture.cs
```diff
@@ -48,10 +47,11 @@ public void Should_use_filename_and_content_type_for_attachments_from_file_respo
             var assemblyPath =
                 Path.GetDirectoryName(this.GetType().Assembly.Location);
 
-            GenericFileResponse.SafePaths = new List<string> {assemblyPath};
+            var environment = new DefaultNancyEnvironment();
+            environment.AddValue(new SafePathConfiguration(new[]{ assemblyPath }));
```
khellang: Isn't there an extension method for this?

khellang: Preferably taking a `params` arg...

src/Nancy.Tests/Unit/Responses/GenericFileResponseFixture.cs
```diff
@@ -11,13 +12,19 @@ public class GenericFileResponseFixture
     {
         private readonly string imagePath;
         private const string imageContentType = "image/png";
+        private readonly NancyContext context;
+        private readonly INancyEnvironment envrionment;
+        private readonly SafePathConfiguration configuration;
```
khellang: You're not using these (`environment` and `configuration`= anywhere...

src/Nancy.Tests/Unit/StaticContentConventionBuilderFixture.cs
```diff
@@ -20,12 +22,15 @@ public class StaticContentConventionBuilderFixture
...
+        private readonly SafePathConfiguration configuration;
 
         public StaticContentConventionBuilderFixture()
         {
             this.directory = Environment.CurrentDirectory;
-
-            GenericFileResponse.SafePaths.Add(this.directory);
+            this.configuration = new SafePathConfiguration(new[]{ this.directory });
+            this.envrionment = new DefaultNancyEnvironment();
+            this.envrionment.AddValue(this.configuration);
```
khellang: Same... Extension method.

src/Nancy.Tests/Unit/Responses/GenericFileResponseFixture.cs
```diff
@@ -11,13 +12,19 @@ public class GenericFileResponseFixture
...
         public GenericFileResponseFixture()
         {
             var assemblyPath =
                 Path.GetDirectoryName(this.GetType().Assembly.Location);
-
-            GenericFileResponse.SafePaths = new List<string> {assemblyPath};
+            this.context = new NancyContext();
+            this.configuration = new SafePathConfiguration(new[]{ assemblyPath });
+            this.envrionment = new DefaultNancyEnvironment();
+            this.envrionment.AddValue(this.configuration);
```
khellang: Extension?

src/Nancy/DefaultSafePathConfigurationProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        /// <summary>
+        /// Gets the default configuration instance to register in the <see cref="INancyEnvironment"/> .
+        /// </summary>
+        /// <returns>The configuration instance.</returns>
+        /// <remarks>Will return <see cref="SafePathConfiguration.Default"/>.</remarks>
+        public override SafePathConfiguration GetDefaultConfiguration()
+        {
+            return new SafePathConfiguration(new []{this.rootPathProvider.GetRootPath()});
```
khellang: Nit: spaces?

src/Nancy/SafePathConfiguration.cs
```diff
@@ -0,0 +1,26 @@
...
+        public SafePathConfiguration(IEnumerable<string> paths)
+        {
+            this.Paths = paths;
+        }
+
+        /// <summary>
+        /// Gets or sets the paths.
+        /// </summary>
+        /// <value>The paths.</value>
+        public IEnumerable<string> Paths { get; set; }
```
khellang: Can this be immutable?

thecodejunkie: Let me rephrase that for you.. "this *needs* to be immutable" ktnxbai :)

src/Nancy/SafePathConfigurationExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace Nancy
+{
+    using System.Collections.Generic;
+    using Nancy.Configuration;
+
+    /// <summary>
+    /// Contains <see cref="SafePathConfiguration"/> configuration for <see cref="INancyEnvironment"/>.
```
thecodejunkie: You're missing "extensions" in this description. Look at the other ones to see what I mean

src/Nancy.Tests/Unit/ResponseExtensionsFixture.cs
jchannon: there is an extension method but whats the diff in reality? will still be same number LOC. a params arg of what?

khellang: > there is an extension method but whats the diff in reality?

It looks better and shows how you're "supposed" to use it? :smile:

> a params arg of what?

`string[]`?

jchannon: Whats the benefit of `params string[]` over `IEnumerable<string>`

khellang: You can do `environment.SafePaths(assemblyPath)` instead of `environment.SafePaths(new[] { assemblyPath })` :smile:

jchannon:
updated per comments - only issue i see is this https://github.com/NancyFx/Nancy/pull/2184/files#diff-8d35d2dce4ba30c6c4950c03202edef0R281 but not sure what else can be done

khellang:
@jchannon Can't you just pass in the environment to the method?

src/Nancy/SafePathConfigurationExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// Contains <see cref="SafePathConfiguration"/> configuration extensions for <see cref="INancyEnvironment"/>.  
+    /// </summary>
+    public static class SafePathConfigurationExtensions
+    {
+        /// <summary>
+        /// Configures <see cref="SafePathConfiguration"/> 
+        /// </summary>
+        /// <param name="environment"><see cref="INancyEnvironment"/> that should be configured.</param>
+        /// <param name="paths">Safe paths.</param>
+        public static void Paths(this INancyEnvironment environment, params string[] paths)
```
khellang: This method name still makes no sense. "Paths" what?

jchannon: Dunno. Same as Views for view engines

jchannon:
Pass the Env to the method via a Theory?

src/Nancy/SafePathConfigurationExtensions.cs
khellang: OK, we have a bad name somewhere, let's use that as a precedent for new APIs :wink:

khellang:
@jchannon Yes, you've created a field, `envrionment`, in the class. Pass that to the method?

jchannon:
Renamed Paths but can't use Theory as it needs static data. I could use `this.environment.AddSafePaths` but that overwrites the SafePathConfiguration already in the environment and then blows up as the key is already there.  Problem with making SafePathConfiguration:Paths immutable is we can't append them

jchannon:
Or in AddSafePaths we check to see if a SafePathConfiguration has been added and if so remove it and add a new one with the incoming path but not sure thats a great idea?

jchannon:
Question comes down to whether you should be allowed to append the environment's safe paths once you have already initialised them

thecodejunkie:
A couple of remarks

1. The config stuff is designed so have immutable classes, exactly because you don't want stuff to modify, the config bits, all over the place. A single point of configuration
1. The whole `SafePathConfiguration` sounds just a bit too specific to be its own config entity. This might be *a* config property of config for static content in general. Even though we're not migrating (yet) other parts of static content, to the new config API, it does not mean we cannot put this property in a `StaticContentConfiguration` class
1. If we do 2 then the extension method would be called `StaticContent`

@khellang If you/we thing that `Views` isn't a good name for the view engine related config method, then *now* is the time for us to change it. Suggestions are more than welcome

jchannon:
So you'd have `StaticContentConfiguration` with a property called
`SafePathConfiguration` which is of type `SafePathConfiguration` and you
could access `Paths`

Would the env extension not be StaticContent.AddSafePaths and in time you'd
have StaticContent.XXX?

thecodejunkie:
@jchannon no. There's no need for the inner config type. `SafePaths` (if that's what we want to call it) would be the only one on there anyway so there's no need to "namespace" stuff.

```c#
public class StaticContentConfiguration
{
   ... other bits and bobs

  public IReadOnlyCollection<string> SafePaths { get; private set; }

  ... other static content config properties once we migrate them at some point
}
```

The `INancyEnvironment` extension would probably be something like this

```c#
env.StaticContent(
   safePaths: new [] { ..... });
```

granted I've not given it a lot of though, but I see no point in locking ourselves in with `AddXXX` methods here. For all we know it might be more convenient, for users, to stick that list somewhere else and just pass it in or read them from the database or what not. Having to do a translation from some sort of list, into multiple `AddXXX` calls doesn't sound nice :P

Path of least resistance

jchannon:
@thecodejunkie done!

src/Nancy/StaticContentConfigurationExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// Contains <see cref="StaticContentConfiguration"/> configuration extensions for <see cref="INancyEnvironment"/>.  
+    /// </summary>
+    public static class StaticContentConfigurationExtensions
+    {
+        /// <summary>
+        /// Configures <see cref="StaticContentConfiguration"/> 
+        /// </summary>
+        /// <param name="environment"><see cref="INancyEnvironment"/> that should be configured.</param>
+        /// <param name="paths">Safe paths.</param>
+        public static void StaticContent(this INancyEnvironment environment, params string[] paths)
```
thecodejunkie: I think the `paths` parameter needs to be renamed to something more descriptive and explicit, perhaps ´safePaths` because that's what it is called everywhere else. I also think we need to update the XML comment, for the parameter, to be a lot more descriptive. The XML comment should let you understand what the parameters means without having to resort to external docks considering this is a config API

src/Nancy/StaticContentConfigurationExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            environment.AddValue(new StaticContentConfiguration(paths));
```
thecodejunkie: Look how values are inserted for the other config stuff, both the use of the named parameters and the formatting

src/Nancy/StaticContentConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// Initializes a new instance of the <see cref="StaticContentConfiguration"/> class.
+        /// </summary>
+        /// <param name="safePaths">StaticContent.</param>
+        public StaticContentConfiguration(IEnumerable<string> safePaths)
+        {
+            this.SafePaths = safePaths;
+        }
+
+        /// <summary>
+        /// Gets or sets the SafePaths.
```
thecodejunkie: This is an auto-generated XML comment if I've ever seen one! :D GhostDoc has no place in our code. More descriptive and explicit comment please :D

src/Nancy/StaticContentConfiguration.cs
```diff
@@ -0,0 +1,25 @@
+namespace Nancy
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Safe path configuration.
```
thecodejunkie: You forgot to fix this when you renamed the class ;)

src/Nancy/StaticContentConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public class StaticContentConfiguration
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="StaticContentConfiguration"/> class.
+        /// </summary>
+        /// <param name="safePaths">StaticContent.</param>
```
thecodejunkie: Umm :)

src/Nancy/StaticContentConfigurationExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+    /// <summary>
+    /// Contains <see cref="StaticContentConfiguration"/> configuration extensions for <see cref="INancyEnvironment"/>.  
+    /// </summary>
+    public static class StaticContentConfigurationExtensions
+    {
+        /// <summary>
+        /// Configures <see cref="StaticContentConfiguration"/> 
+        /// </summary>
+        /// <param name="environment"><see cref="INancyEnvironment"/> that should be configured.</param>
```
thecodejunkie: Check how this parameter has been commented in the other extension methods

src/Nancy/DefaultStaticContentConfigurationProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+            this.rootPathProvider = rootPathProvider;
+        }
+
+        /// <summary>
+        /// Gets the default configuration instance to register in the <see cref="INancyEnvironment"/> .
+        /// </summary>
+        /// <returns>The configuration instance.</returns>
+        public override StaticContentConfiguration GetDefaultConfiguration()
+        {
+            return new StaticContentConfiguration(new []{this.rootPathProvider.GetRootPath()});
```
thecodejunkie: Explicitly use a named parameter here please

src/Nancy/DefaultStaticContentConfigurationProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// Provides the default configuration for <see cref="StaticContentConfiguration"/>. 
+    /// </summary>
+    public class DefaultStaticContentConfigurationProvider : NancyDefaultConfigurationProvider<StaticContentConfiguration>
+    {
+        private readonly IRootPathProvider rootPathProvider;
+
+        /// <summary>
+        /// Creates and instance of DefaultStaticContentConfigurationProvider
+        /// </summary>
+        /// <param name="rootPathProvider">Use <see cref="IRootPathProvider"/> to get root path</param>
```
thecodejunkie: Questionable parameter comment. NIT :)

src/Nancy/DefaultStaticContentConfigurationProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// <summary>
+    /// Provides the default configuration for <see cref="StaticContentConfiguration"/>. 
+    /// </summary>
+    public class DefaultStaticContentConfigurationProvider : NancyDefaultConfigurationProvider<StaticContentConfiguration>
+    {
+        private readonly IRootPathProvider rootPathProvider;
+
+        /// <summary>
+        /// Creates and instance of DefaultStaticContentConfigurationProvider
```
thecodejunkie: `<see ...  />` tag around the class name

src/Nancy.Tests/Unit/StaticContentConventionBuilderFixture.cs
```diff
@@ -274,28 +278,30 @@ private Response GetStaticContentResponse(string virtualDirectory, string reques
 
             var rootFolder = root ?? this.directory;
 
-            GenericFileResponse.SafePaths.Add(rootFolder);
+            context.Environment = new DefaultNancyEnvironment();
+            context.Environment.StaticContent(rootFolder);
```
thecodejunkie: Explicitly use a named parameter here please

src/Nancy.Tests/Unit/StaticContentConventionBuilderFixture.cs
```diff
@@ -20,12 +22,13 @@ public class StaticContentConventionBuilderFixture
...
         private readonly string directory;
+        private readonly INancyEnvironment envrionment;
 
         public StaticContentConventionBuilderFixture()
         {
             this.directory = Environment.CurrentDirectory;
-
-            GenericFileResponse.SafePaths.Add(this.directory);
+            this.envrionment = new DefaultNancyEnvironment();
+            this.envrionment.StaticContent(this.directory);
```
thecodejunkie: Explicitly use a named parameter here please

src/Nancy.Tests/Unit/Responses/GenericFileResponseFixture.cs
```diff
@@ -11,13 +12,17 @@ public class GenericFileResponseFixture
...
         private const string imageContentType = "image/png";
+        private readonly NancyContext context;
 
         public GenericFileResponseFixture()
         {
             var assemblyPath =
                 Path.GetDirectoryName(this.GetType().Assembly.Location);
+            
+            var envrionment = new DefaultNancyEnvironment();
+            envrionment.StaticContent(assemblyPath);
```
thecodejunkie: Explicitly use a named parameter here please

src/Nancy.Tests/Unit/ResponseExtensionsFixture.cs
```diff
@@ -48,10 +47,11 @@ public void Should_use_filename_and_content_type_for_attachments_from_file_respo
             var assemblyPath =
                 Path.GetDirectoryName(this.GetType().Assembly.Location);
 
-            GenericFileResponse.SafePaths = new List<string> {assemblyPath};
+            var environment = new DefaultNancyEnvironment();
+            environment.StaticContent(assemblyPath);
```
thecodejunkie: Explicitly use a named parameter here please

src/Nancy/StaticContentConfigurationExtensions.cs
jchannon: Its exactly the same as ViewExtensions

src/Nancy/DefaultStaticContentConfigurationProvider.cs
jchannon: What?

thecodejunkie: `Creates and instance of <see cref="DefaultStaticContentConfigurationProvider" />`

src/Nancy/StaticContentConfigurationExtensions.cs
thecodejunkie: Then we need to fix that, there's a missing `An` at the start of that sentence =)

src/Nancy.Tests/Unit/StaticContentConventionBuilderFixture.cs
```diff
@@ -274,28 +278,30 @@ private Response GetStaticContentResponse(string virtualDirectory, string reques
...
             var context = new NancyContext
             {
                 Request = request,
                 Trace = new DefaultRequestTrace
                 {
-                    TraceLog = new DefaultTraceLog()
-                }
+                    TraceLog = new DefaultTraceLog(),
+                },
+                Environment = this.envrionment
```
thecodejunkie: `envrionment` you say? :-D

src/Nancy.Tests/Unit/Responses/GenericFileResponseFixture.cs
```diff
@@ -11,13 +12,17 @@ public class GenericFileResponseFixture
...
         private readonly string imagePath;
         private const string imageContentType = "image/png";
+        private readonly NancyContext context;
 
         public GenericFileResponseFixture()
         {
             var assemblyPath =
                 Path.GetDirectoryName(this.GetType().Assembly.Location);
+            
+            var envrionment = new DefaultNancyEnvironment();
```
thecodejunkie: Copy/paste for the win `envrionment` :-D

PR #2186, opened 2016-01-12
Started migrating routing configuration to the new configuration API

thecodejunkie:
Moved routing configurations over to the new configuration as part of #2000 

1. Added `RouteConfiguration`, `RouteConfigurationExtensions` and `DefaultRouteConfigurationProvider`
1. Renamed `EnableHeadRouting` to `ExplicitHeadRouting` when it was moved
1. Removed routing related settings from `StaticConfiguration`
1. Removed `HEAD` routing configuration check in `NancyModule.Head` (see discussion below)
1. Added the `INancyEnvironment` onto `NancyContext` :warning: 

### Old description

Currently what's left is to figure out the best approach to get the `INancyEnvironment` into `NancyModule` because it is currently used by the [Head](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/NancyModule.cs#L87) declaration to throw an exception if you try to declare `HEAD` route while it's disabled

I think our options are one of the following

1. Add it to `INancyModule` (feels icky)
2. Add it to `NancyContext` (feels less icky, and makes it accessible wherever the context is)
3. Remove the check & exception from `NancyModule`

I think option 3 is the most compelling because the [DefaultRouteResolve](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Routing/DefaultRouteResolver.cs#L163) also makes use of the setting to determine if it should translate `HEAD` requests into `GET` requests. Removing the check & exception would  only mean that you'd not get an exception (which is a bit weird anyway) if you try to declare `HEAD` requests while it's disabled.. it will just mean they'll never been able to be called until you toggle the setting

### DIFF
--- a/src/Nancy.Testing/StaticConfigurationContext.cs
+++ b/src/Nancy.Testing/StaticConfigurationContext.cs
@@ -19,8 +19,6 @@ namespace Nancy.Testing
         {
             this.existingConfiguration.CaseSensitive = StaticConfiguration.CaseSensitive;
             this.existingConfiguration.DisableErrorTraces = StaticConfiguration.DisableErrorTraces;
-            this.existingConfiguration.DisableMethodNotAllowedResponses = StaticConfiguration.DisableMethodNotAllowedResponses;
-            this.existingConfiguration.EnableHeadRouting = StaticConfiguration.EnableHeadRouting;
             this.existingConfiguration.EnableRequestTracing = StaticConfiguration.EnableRequestTracing;
             this.existingConfiguration.RequestQueryFormMultipartLimit = StaticConfiguration.RequestQueryFormMultipartLimit;
 
@@ -44,8 +42,6 @@ namespace Nancy.Testing
         {
             StaticConfiguration.CaseSensitive = values.CaseSensitive;
             StaticConfiguration.DisableErrorTraces = values.DisableErrorTraces;
-            StaticConfiguration.DisableMethodNotAllowedResponses = values.DisableMethodNotAllowedResponses;
-            StaticConfiguration.EnableHeadRouting = values.EnableHeadRouting;
             StaticConfiguration.EnableRequestTracing = values.EnableRequestTracing;
             StaticConfiguration.RequestQueryFormMultipartLimit = values.RequestQueryFormMultipartLimit;
         }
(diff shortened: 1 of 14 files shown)


jchannon:
what would you get for option 3? 405 response?

thecodejunkie:
@jchannon that would depend on this https://github.com/NancyFx/Nancy/pull/2186/files#diff-85f32a7553a78a9ff4b2905f7c21280cR30

phillip-haydon:
IMO this check should be done before it ever gets to the Module... 

I vote 3.

grumpydev:
I'd prefer option 2, feels like something that would benefit from being on the context anyway.

thecodejunkie:
@grumpydev I kind of thing 2 & 3 would be the way. I think throwing the exception is stupid. It means you can't temporarily toggle that feature on/off without making modification (even if that only means commenting out code) or it will throw exception at startup. I would vote 3 for this PR and then we create a new issue for 2 ?

grumpydev:
Only problem with removing the exception is that people could then add HEAD routes and they just wouldn't work with no indication as to why. I agree the exception should go, but I guess only when we fix the routing so it will give custom HEAD routes precedence over the default stuff, although I'm not sure why we don't already do that as that seems the "correct" thing to do.

thecodejunkie:
@grumpydev I'm trying to limit the scope so we can ship 2.0, not expand it.. unless you're offering to get a PR in for that behaviour? :D

grumpydev:
I'm just saying don't remove the exception yet until we've fixed it properly, so stick it in the context and log an issue to remove the exception/improve routing later :)

thecodejunkie:
@grumpydev you're just trying to avoid doing any work! ;)

grumpydev:
Avoid, no, postpone, yes :p

phillip-haydon:
Can someone explain to me why it belongs in the module or context when this AFAIK dictates if routing should occur?

grumpydev:
We are talking about the nancy environment object (and hence config in general) not this particular setting @phillip-haydon

grumpydev:
The setting itself should go away, but not yet.

phillip-haydon:
In that case option 2 and remove that setting when we can because it makes no sense

thecodejunkie:
No longer WIP. Updated description. Ready to be reviewed and pulled ping @NancyFx/most-valued-minions @NancyFx/nancy-core-contributors

phillip-haydon:
:+1: 

Looks much cleaner than the old statics everywhere.

thecodejunkie:
Added one more thing to the description (Step 5)

jchannon:
:shipit:

PR #2189, opened 2016-01-14
Update Customer.cs

uliian:
add a prompt


khellang:
What does this mean?

uliian:
Nancy implement IValidatableObject different ASP.NET
In ASP.NET if I yield return a ValidationResult object but don't specify member name,like this:
```C#
yield return new ValidationResult("test!");
```

the ModelState.IsValid is false!

but in Nancy,if I don't specify member name
ModelValidationResult.IsValid is true

so,I want annotate it

grumpydev:
Thanks for the info - so the correct way forward is either:

a) File a bug, and potentially a PR with a fix, detailing the differences and why the change needs to be made.

or

b) Update the documentation in wiki with information about the differences

Adding a code comment to one of our sample apps doesn't really help anyone so I'm going to close this for now.

PR #2190, opened 2016-01-15
Update ModelValidationResult.cs

uliian:
Fix bug:
if entity implement `IValidatableObject` but yield return has no assign `memberNames` like this:

```C#
public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
{
     if (this.Id > 100)
     {
          yield return new ValidationResult("The Id cannot be greater than 100");
     }
 }
```

the `ModelValidationResult.IsValid` is `true`.
I fix this bug

### DIFF
--- a/src/Nancy/Validation/ModelValidationResult.cs
+++ b/src/Nancy/Validation/ModelValidationResult.cs
@@ -86,6 +86,16 @@
 
                     output[name].Add(result);
                 }
+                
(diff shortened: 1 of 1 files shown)


thecodejunkie:
Could you please modify you uses of `""` to use `string.Empty` instead?

Thanks :)

uliian:
I've changed the code.It can do the merge？

PR #2201, opened 2016-01-25
Migrated tracing settings from StaticConfiguration to the new config API

thecodejunkie:
Migrated tracing related settings, from `StaticConfiguration` to the new config API.

There is no `TraceConfiguration.Default` because the default configuration rely on being able to tell if we're running in debug mode or not. The old static settings did the same, so the `DefaultTraceConfigurationProvider` class takes a dependency on `IRuntimeEnvironmentInformation` to gain access so `IsDebugMode`

The `INancyEnvironment` extension method for tracing is called `Tracing` and has two, mandatory, configurations; `Enabled` and `DisplayErrorTraces`. Both of these existed on `StaticConfiguration` but was renamed to be more aligned with the fact that they're not in a config entity.

### DIFF
--- a/src/Nancy.Testing.Tests/ContextExtensionsTests.cs
+++ b/src/Nancy.Testing.Tests/ContextExtensionsTests.cs
@@ -102,6 +102,10 @@ namespace Nancy.Testing.Tests
             // Given
             var environment = new DefaultNancyEnvironment();
             environment.AddValue(XmlConfiguration.Default);
+            environment.Tracing(
+                enabled: true,
+                displayErrorTraces: true);
+
             var response = new XmlResponse<Model>(new Model() { Dummy = "Data" }, new DefaultXmlSerializer(environment), environment);
             var context = new NancyContext() { Response = response };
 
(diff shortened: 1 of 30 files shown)


src/Nancy.Tests/Unit/DefaultSerializerFactoryFixture.cs
```diff
@@ -159,12 +159,16 @@ public void Should_return_default_serializer_if_no_other_match_could_be_made()
...
+            var environment =
                 new DefaultNancyEnvironment();
 
-            envionment.AddValue(JsonConfiguration.Default);
+            environment.Tracing(
+                enabled: true,
+                displayErrorTraces: true);
 
-            return envionment;
+            environment.AddValue(JsonConfiguration.Default);
```
jchannon: use `environment.Json())` which uses the extension method

src/Nancy.Testing.Tests/ContextExtensionsTests.cs
```diff
@@ -129,12 +133,16 @@ public void Should_fail_to_return_xml_body_on_non_xml_response()
...
-            var envionment =
+            var environment =
                 new DefaultNancyEnvironment();
 
-            envionment.AddValue(JsonConfiguration.Default);
+            environment.Tracing(
+                enabled: true,
+                displayErrorTraces: true);
+
+            environment.AddValue(JsonConfiguration.Default);
```
jchannon: `environment.Json()`

src/Nancy/TraceConfigurationExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    /// <summary>
+    /// Contains <see cref="TraceConfiguration"/> configuration extensions for <see cref="INancyEnvironment"/>.
+    /// </summary>
+    public static class TraceConfigurationExtensions
+    {
+        /// <summary>
+        /// Configures <see cref="TraceConfiguration"/>.
+        /// </summary>
+        /// <param name="environment"><see cref="INancyEnvironment"/> that should be configured.</param>
```
jchannon: Needs `An`

PR #2202, opened 2016-01-26
demo for conventions

hwqdt:
it is necessary to add a demo to tell users how to use Convention


thecodejunkie:
Hey, thanks for sending this in but, unfortunately, we will not be pulling this in. The reason is that we're not adding and entire project (35 files) simple to show this line of code

`nancyConventions.StaticContentsConventions.Add(StaticContentConventionBuilder.AddDirectory("assets");`

We have quite extensive documentation for [Managing Static Content](https://github.com/NancyFx/Nancy/wiki/Managing-static-content) in our wiki.

Thanks!

PR #2206, opened 2016-01-31
Implemented ITypeCatalog and IAssemblyCatalog

thecodejunkie:
This is meant to be a pre-requisite for the instance based scanning (#1846) change that we want to replace the `AppDomainAssemblyTypeScanner`

- Adds `ITypeCatalog` and the ´DefaultTypeCatalog`. This default implementation uses a cache internally because once Nancy is booted up, it should be treated as immutable anyway.
- Adds `IAssemblyCatalog` and the `AppDomainAssemblyCatalog`. This implementation mimics a lot of the assembly functionality in `AppDomainAssemblyTypeScanner` by means of ensuring that Nancy referencing assemblies are loaded and it looks the bin paths for it.
- Replaces `ScanMode` with a strategy based approach. The strategy is defined by the `TypeResolveStrategy` delegate and there are implementations, in `TypeResolveStrategies` that represent the `ScanMode` members. This is a better approach because you can add strategies if you need them in your Nancy application.
- Includes `ITypeCatalogExtension` which adds a couple of `ITypeCatalog.GetTypesAssignableTo` overloads which uses `TypeResolveStrategies.All`

## Discussion point

The `AppDomainAssemblyCatalog` only scans over 

- Nancy assembly,
- Any Nancy referencing assemblies
- Ignored `Nancy.Testing` assembly

But has no means of letting someone bring more assemblies to its attention. `AppDomainAssemblyTypeScanner` had a bunch of `AddAssembliesToScan` and `LoadAssemblies` method to allow you to do this. However I think this design is bad and needs a bit of rethinking. In my opinion, these should be no runtime modification of this catalog, because most parts of Nancy is unable to react to the changes and make use of the new types anyway.

So I thought about adding something like

```c#
public virtual IEnumerable<string> AddtionalAssembliesToScan()
{
   return Enumerable.Empty<string>();
}
```

To the `AppDomainAssemblyTypeScanner` implementation. So if you wanted to include more assemblies, you would simply derive a new class, override `AddtionalAssembliesToScan` (naming suggestions are welcome!) and tell Nancy to use this assembly catalog instead. 

I thought about adding it as a ctor overload instead, but it just gets ugly quick. Supporting both is not something I would like to do.

Thoughts?

### DIFF
--- a/src/Nancy/AppDomainAssemblyCatalog.cs
+++ b/src/Nancy/AppDomainAssemblyCatalog.cs
@@ -1,7 +1,6 @@
 namespace Nancy
 {
     using System;
-    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
@@ -11,31 +10,21 @@ namespace Nancy
     /// Default implementation of the <see cref="IAssemblyCatalog"/> interface, based on
     /// retrieving <see cref="Assembly"/> information from <see cref="AppDomain.CurrentDomain"/>.
     /// </summary>
-    public class AppDomainAssemblyCatalog : IAssemblyCatalog
+    public class AppDomainAssemblyCatalog : AssemblyCatalogBase
     {
-        private readonly Lazy<IReadOnlyCollection<Assembly>> assemblies = new Lazy<IReadOnlyCollection<Assembly>>(GetAssembliesInAppDomain);
-        private readonly ConcurrentDictionary<AssemblyResolveStrategy, IReadOnlyCollection<Assembly>> cache;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="AppDomainAssemblyCatalog"/> class.
         /// </summary>
         public AppDomainAssemblyCatalog()
         {
-            this.cache = new ConcurrentDictionary<AssemblyResolveStrategy, IReadOnlyCollection<Assembly>>();
-
-            EnsureNancyReferencingAssembliesAreLoaded();
+            this.EnsureNancyReferencingAssembliesAreLoaded();
         }
 
         /// <summary>
-        /// Gets all <see cref="Assembly"/> instances in the catalog.
+        /// Get all available <see cref="Assembly"/> instances.
         /// </summary>
         /// <returns>An <see cref="IReadOnlyCollection{T}"/> of <see cref="Assembly"/> instances.</returns>
-        public IReadOnlyCollection<Assembly> GetAssemblies(AssemblyResolveStrategy strategy)
-        {
-            return this.cache.GetOrAdd(strategy, s => this.assemblies.Value.Where(s.Invoke).ToArray());
-        }
-
-        private static IReadOnlyCollection<Assembly> GetAssembliesInAppDomain()
+        protected override IReadOnlyCollection<Assembly> GetAvailableAssemblies()
         {
             return AppDomain.CurrentDomain
                 .GetAssemblies()
@@ -44,9 +33,9 @@ namespace Nancy
                 .ToArray();
         }
 
-        private static void EnsureNancyReferencingAssembliesAreLoaded()
+        private void EnsureNancyReferencingAssembliesAreLoaded()
         {
-            var assemblyPaths = GetAssembliesInAppDomain()
+            var assemblyPaths = this.GetAvailableAssemblies()
                 .Select(assembly => assembly.Location)
                 .ToArray();
 
--- a/src/Nancy/AppDomainAssemblyCatalog.cs
+++ b/src/Nancy/AppDomainAssemblyCatalog.cs
@@ -1,0 +1,145 @@
+namespace Nancy
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Collections.ObjectModel;
+    using System.IO;
+    using System.Linq;
+    using System.Reflection;
+
+    /// <summary>
+    /// Default implementation of the <see cref="IAssemblyCatalog"/> interface, based on
+    /// retrieving <see cref="Assembly"/> information from <see cref="AppDomain.CurrentDomain"/>.
+    /// </summary>
+    public class AppDomainAssemblyCatalog : IAssemblyCatalog
+    {
+        private readonly Lazy<IReadOnlyCollection<Assembly>> assemblies = new Lazy<IReadOnlyCollection<Assembly>>(GetAllAssemblies);
+
+        /// <summary>
+        /// Gets all <see cref="Assembly"/> instances in the catalog.
+        /// </summary>
+        /// <returns>An <see cref="IReadOnlyCollection{T}"/> of <see cref="Assembly"/> instances.</returns>
+        public IReadOnlyCollection<Assembly> GetAssemblies()
+        {
+            return this.assemblies.Value;
+        }
+
+        private static IReadOnlyCollection<Assembly> GetAllAssemblies()
+        {
+            EnsureNancyReferencingAssembliesAreLoaded();
+
+            return GetAssembliesInAppDomain();
+        }
+
+        private static IReadOnlyCollection<Assembly> GetAssembliesInAppDomain()
+        {
+            var assemblies = AppDomain.CurrentDomain
+                                      .GetAssemblies()
+                                      .Where(IsNancyReferencingAssembly)
+                                      .Where(assembly => !assembly.IsDynamic)
+                                      .Where(assembly => !assembly.ReflectionOnly)
+                                      .ToArray();
+
+            return new ReadOnlyCollection<Assembly>(assemblies);
+        }
+
+        private static bool IsNancyReferencingAssembly(Assembly assembly)
+        {
+            if (assembly.Equals(typeof(INancyEngine).Assembly))
+            {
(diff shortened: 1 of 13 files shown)


jchannon:
Add in what youve done with the ability to add extra assemblies due to the shit .net linker

thecodejunkie:
@jchannon I really did try to make sense of what you said, but I failed :D

jchannon:
get rid of AppDomainAssemblyTypeScanner but make AddtionalAssembliesToScan  visible somewhere

thecodejunkie:
@jchannon oh you're saying that the virtual member is a good approach? =D

grumpydev:
My thinking around this was to have strategies for assembly loading / discovery *and* scanning of the types, so no virtual property needed, then we ship some defaults in the box, including the "things referencing Nancy" one.

thecodejunkie:
@grumpydev so break out the `IsNancyReferencingAssembly` method into a strategy, keep `EnsureNancyReferencingAssembliesAreLoaded` in the class and perhaps add some per-strategy caching ?

src/Nancy/AppDomainAssemblyCatalog.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                return true;
+            }
+
+            if (assembly.GetName().Name.Equals("Nancy.Testing", StringComparison.OrdinalIgnoreCase))
+            {
+                return false;
+            }
+
+            if (assembly.GetReferencedAssemblies().Any(reference => reference.Name.StartsWith("Nancy", StringComparison.OrdinalIgnoreCase)))
```
khellang: `Equals`?

src/Nancy/DefaultTypeCatalog.cs
```diff
@@ -0,0 +1,59 @@
...
+            return this.cache.GetOrAdd(type, t => this.GetTypesAssignableTo(type)).Where(strategy.Invoke).ToArray();
+        }
+
+        /// <summary>
+        /// Gets all types that are assignable to the provided <typeparamref name="TType"/>.
+        /// </summary>
+        /// <typeparam name="TType">The <see cref="Type"/> that returned types should be assignable to.</typeparam>
+        /// <param name="strategy">A <see cref="TypeResolveStrategy"/> that should be used then retrieving types.</param>
+        /// <returns>An <see cref="IReadOnlyCollection{T}"/> of <see cref="Type"/> instances.</returns>
+        public IReadOnlyCollection<Type> GetTypesAssignableTo<TType>(TypeResolveStrategy strategy)
```
khellang: Extension method?

src/Nancy/AppDomainAssemblyCatalog.cs
thecodejunkie: Yep :+1:

src/Nancy/DefaultTypeCatalog.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            return this.GetTypesAssignableTo(typeof(TType), strategy);
+        }
+
+        private IReadOnlyCollection<Type> GetTypesAssignableTo(Type type)
+        {
+            return this.assemblyCatalog
+                            .GetAssemblies()
```
khellang: Let's kill the leading whitespace here...

src/Nancy/AppDomainAssemblyCatalog.cs
```diff
@@ -0,0 +1,145 @@
...
+            EnsureNancyReferencingAssembliesAreLoaded();
+
+            return GetAssembliesInAppDomain();
+        }
+
+        private static IReadOnlyCollection<Assembly> GetAssembliesInAppDomain()
+        {
+            var assemblies = AppDomain.CurrentDomain
+                                      .GetAssemblies()
+                                      .Where(IsNancyReferencingAssembly)
```
khellang: Let's kill the leading whitespace here...

src/Nancy/ITypeCatalog.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <returns>An <see cref="IReadOnlyCollection{T}"/> of <see cref="Type"/> instances.</returns>
+        IReadOnlyCollection<Type> GetTypesAssignableTo(Type type, TypeResolveStrategy strategy);
+
+        /// <summary>
+        /// Gets all types that are assignable to the provided <typeparamref name="TType"/>.
+        /// </summary>
+        /// <typeparam name="TType">The <see cref="Type"/> that returned types should be assignable to.</typeparam>
+        /// <param name="strategy">A <see cref="TypeResolveStrategy"/> that should be used then retrieving types.</param>
+        /// <returns>An <see cref="IReadOnlyCollection{T}"/> of <see cref="Type"/> instances.</returns>
+        IReadOnlyCollection<Type> GetTypesAssignableTo<TType>(TypeResolveStrategy strategy);
```
khellang: Extension method?

src/Nancy/ITypeCatalogExtensions.cs
```diff
@@ -0,0 +1,33 @@
+namespace Nancy
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Contains extension methods for <see cref="ITypeCatalog"/> implementations.
+    /// </summary>
+    public static class ITypeCatalogExtensions
```
khellang: Remove the `I` prefix?

src/Nancy/TypeResolveStrategies.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            return type.Assembly.Equals(typeof(INancyEngine).Assembly);
+        };
+
+        /// <summary>
+        /// Resolve types that are located in the Nancy namespace.
+        /// </summary>
+        public static readonly TypeResolveStrategy OnlyNancyNamespace = type =>
+        {
+            return type.Assembly.GetName().Name.StartsWith("Nancy", StringComparison.OrdinalIgnoreCase);
```
khellang: Is this correct? Shouldn't this check the `Namespace` property instead?

src/Nancy/AppDomainAssemblyCatalog.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        private static Assembly SafeLoadReflectionOnlyAssembly(string assemblyPath)
+        {
+            try
+            {
+                return Assembly.ReflectionOnlyLoadFrom(assemblyPath);
+            }
+            catch (BadImageFormatException)
+            {
+                //the assembly maybe it's not managed code
```
khellang: Nit: `might not be a managed assembly`?

thecodejunkie:
Had a quick chat with @grumpydev and we came up with the idea that it might be a good idea to create a base class for `IAssemblyCatalog` that takes care of caching and strategy application. It would have a virtual `GetAvailableAssemblies` method on it that you would implement. For .NET Framework it would get them from the app domain, while on Core it would use the `LibraryManager`

thecodejunkie:
Implemented `AssemblyCatalogBase` and updated `AppDomainAssemblyCatalog` to inherit from it

src/Nancy/AppDomainAssemblyCatalog.cs
```diff
@@ -0,0 +1,116 @@
...
+            catch
+            {
+            }
+        }
+
+        private static Assembly SafeLoadReflectionOnlyAssembly(string assemblyPath)
+        {
+            try
+            {
+                return Assembly.ReflectionOnlyLoadFrom(assemblyPath);
```
Yantrio: If you are loading this in reflection only mode to check if it's loadable and for no other reason, is it a good idea to load them into a seperate app domain then unload that entire app domain once it's determined that the dlls are okay, just to make sure that you aren't keeping things in the app domain that don't need to be there?

thecodejunkie: They're loaded as reflection only so they won't stick around in the app domain unless we explicitly decide to load them in after the checks have passed. Setting up another app domain wouldn't get rid of the load into the current domain + we would have to take the cost of putting up a new app domain and then pulling it back down again

khellang: It's loaded in reflection context because it's used to check whether the assembly references Nancy.

src/Nancy/AssemblyCatalogBase.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// <summary>
+    /// Base class implementation of <see cref="IAssemblyCatalog"/> which adds per-strategy caching.
+    /// </summary>
+    public abstract class AssemblyCatalogBase : IAssemblyCatalog
+    {
+        private readonly Lazy<IReadOnlyCollection<Assembly>> assemblies;
+        private readonly ConcurrentDictionary<AssemblyResolveStrategy, IReadOnlyCollection<Assembly>> cache;
+
+
```
khellang: :fire:

src/Nancy/AppDomainAssemblyCatalog.cs
Yantrio: Okay, that makes sense :+1:

PR #2212, opened 2016-02-02
Added new solution with Nancy building for net451

khellang:


Yantrio:
What's the plan with the shared assemblyinfo?
Looks like this could be a solution https://github.com/aspnet/Home/wiki/Project.json-file#shared-files

global.json
```diff
@@ -0,0 +1,6 @@
+{
+    "projects": [ "src", "test", "samples" ],
+    "sdk": {
+        "version": "1.0.0-rc1-final"
```
Yantrio: How come you chose final and not update1?

khellang: Cause it's a long time ago and they're basically the same :smile:

khellang:
> What's the plan with the shared assemblyinfo?

What for? Versioning is handled by the build, so there's no need for patching assembly info. We can just use something like `DNX_BUILD_VERSION` (or the dotnet equivalent)

All the other info is duplicated in project.json for now, because there's no support for pulling up author info etc. in global.json, yet.

The SharedAssemblyInfo.cs is still around because this will just be an addition to the csproj based system, not a replacement yet. There's still loads to do before this is a working alternative.

thecodejunkie:
@Yantrio our `SharedAssemblyInfo.cs` is managed by our build script and really  never manually edited. We issue `rake nancy_build[<version_number_goes_here>]` and it updates it for us and does other magic :D

thecodejunkie:
So this is just to get our current `v2`-target running on both the old and the new project format. It'll be a branching point :D

Yantrio:
Okay, everything else LGTM :+1:

khellang:
Alright. That's it for me :wink: Good look with the CoreCLR porting :grin:

thecodejunkie:
image

PR #2217, opened 2016-02-02
Added type and assembly catalogs on the bootstrapper

thecodejunkie:
Added `AssemblyCatalog` and `TypeCatalog`, as `protected virtual` properties on the `NancyBootstrapperBase<T>` class + made sure they are also registered in the container. This is going to make it possible for us to take dependencies on `IAssemblyCatalog` and `ITypeCatalog` in the code and replace the use of `AppDomainAssemblyTypeScanner`

PR #2222, opened 2016-02-03
Replaced use of ADATS with ITypeCatalog inside the bootstrapperbase

thecodejunkie:
Changed `NancyBootstrappeBase<T>` to make use of the new `AssemblyCatalog` and `TypeCatalog` properties that were added by #2217. This makes the main class free of all `AppDomainAssemblyTypeScan` uses.

### DIFF
--- a/src/Nancy/Bootstrapper/NancyBootstrapperBase.cs
+++ b/src/Nancy/Bootstrapper/NancyBootstrapperBase.cs
@@ -133,8 +133,8 @@
                 return
                     this.modules
                     ??
-                    (this.modules = AppDomainAssemblyTypeScanner
-                                        .TypesOf<INancyModule>(ScanMode.ExcludeNancy)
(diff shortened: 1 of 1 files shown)


src/Nancy/Bootstrapper/NancyBootstrapperBase.cs
```diff
@@ -187,7 +187,7 @@ protected virtual IEnumerable<Type> BodyDeserializers
         /// </summary>
         protected virtual IEnumerable<Type> ApplicationStartupTasks
         {
-            get { return AppDomainAssemblyTypeScanner.TypesOf<IApplicationStartup>(); }
+            get { return this.typeCatalog.GetTypesAssignableTo<IApplicationStartup>(); }
```
jchannon: use `TypeCatalog`

Issue #2221, opened 2016-02-03
Migrate uses of AppDomainAssemblyTypeScanner to the new catalogs


thecodejunkie:
With the introduction of the new `IAssemblyCatalog` and `ITypeCatalog` (see #2217) we need to migrate the code to use these new types instead of the old (static) `AppDomainAssemblyTypeScanner`. We can use #1846 as a sort of reference as it was a spike for a similar approach. 

## Nancy
- [x] [NancyBootstrappeBase](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Bootstrapper/NancyBootstrapperBase.cs) - Used to resolve all sorts of types that are going to be put into the container at startup (@TheCodeJunkie #2222)
- [x] [NancyInternalConfiguration](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Bootstrapper/NancyInternalConfiguration.cs) - This is the main consumer, the problem here is that is it being used from the static property [Default](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Bootstrapper/NancyInternalConfiguration.cs#L32). The class is used from the [InternalConfiguration](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Bootstrapper/NancyBootstrapperBase.cs#L83) property on the NancyBoostrapperBase (@TheCodeJunkie #2225)
- [x] [Registrations](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Bootstrapper/Registrations.cs) - It's used in a couple of helper methods. The `Registrations` instances are pulled from the container but the problem here is that we would have to enforce a ctor signature (something like `protected Registrations(ITypeCatalog catalog)`) and force all implementations to call it, so a major breaking change (@TheCodeJunkie #2240)
- [x] [NancyConventions](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Conventions/NancyConventions.cs) - Used to find the default conventions. Another issue is how we make use of the conventions. Currently we just create a new instance, of `NancyConventions` in [NancyBoostrapperBase.ctor](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Bootstrapper/NancyBootstrapperBase.cs#L72) (@TheCodeJunkie #2231)
[...]

PR #2223, opened 2016-02-03
Removed ADATS in razor localization demo to use IAssemblyCatalog

jchannon:
Removes ADATS from razor localisation project as shown in #2221

### DIFF
--- a/samples/Nancy.Demo.Razor.Localization/CustomResourceAssemblyProvider.cs
+++ b/samples/Nancy.Demo.Razor.Localization/CustomResourceAssemblyProvider.cs
@@ -1,10 +1,7 @@
 ﻿namespace Nancy.Demo.Razor.Localization
 {
     using System.Collections.Generic;
-    using System.Linq;
     using System.Reflection;
-
-    using Nancy.Bootstrapper;
 
     /// <summary>
     /// Use a custom <see cref="IResourceAssemblyProvider"/> because the default one ignores any
@@ -13,16 +10,17 @@
     /// </summary>
     public class CustomResourceAssemblyProvider : IResourceAssemblyProvider
     {
+        private readonly IAssemblyCatalog assemblyCatalog;
         private IEnumerable<Assembly> filteredAssemblies;
+
+        public CustomResourceAssemblyProvider(IAssemblyCatalog assemblyCatalog)
+        {
+            this.assemblyCatalog = assemblyCatalog;
+        }
 
         public IEnumerable<Assembly> GetAssembliesToScan()
         {
-            return (this.filteredAssemblies ?? (this.filteredAssemblies = GetFilteredAssemblies()));
-        }
-
-        private static IEnumerable<Assembly> GetFilteredAssemblies()
-        {
-            return AppDomainAssemblyTypeScanner.Assemblies.Where(x => !x.IsDynamic);
+            return (this.filteredAssemblies ?? (this.filteredAssemblies = this.assemblyCatalog.GetAssemblies()));
         }
     }
 }


samples/Nancy.Demo.Razor.Localization/CustomResourceAssemblyProvider.cs
```diff
@@ -13,16 +11,22 @@
...
         public IEnumerable<Assembly> GetAssembliesToScan()
         {
             return (this.filteredAssemblies ?? (this.filteredAssemblies = GetFilteredAssemblies()));
         }
 
-        private static IEnumerable<Assembly> GetFilteredAssemblies()
+        private IEnumerable<Assembly> GetFilteredAssemblies()
         {
-            return AppDomainAssemblyTypeScanner.Assemblies.Where(x => !x.IsDynamic);
+            return this.assemblyCatalog.GetAssemblies(AssemblyResolveStrategies.All).Where(x => !x.IsDynamic);
```
khellang: Is `AssemblyResolveStrategies.All` needed here?

thecodejunkie: There should be an extension method that defaults to that

jchannon:
Updated to use extension method and squashed it just for @khellang

samples/Nancy.Demo.Razor.Localization/CustomResourceAssemblyProvider.cs
```diff
@@ -13,16 +11,22 @@
     /// </summary>
     public class CustomResourceAssemblyProvider : IResourceAssemblyProvider
     {
+        private readonly IAssemblyCatalog assemblyCatalog;
         private IEnumerable<Assembly> filteredAssemblies;
 
+        public CustomResourceAssemblyProvider(IAssemblyCatalog assemblyCatalog)
```
thecodejunkie: Missing constructor XML comment (including the parameter)

samples/Nancy.Demo.Razor.Localization/CustomResourceAssemblyProvider.cs
```diff
@@ -13,16 +11,22 @@
...
         public IEnumerable<Assembly> GetAssembliesToScan()
         {
             return (this.filteredAssemblies ?? (this.filteredAssemblies = GetFilteredAssemblies()));
         }
 
-        private static IEnumerable<Assembly> GetFilteredAssemblies()
+        private IEnumerable<Assembly> GetFilteredAssemblies()
         {
-            return AppDomainAssemblyTypeScanner.Assemblies.Where(x => !x.IsDynamic);
+            return this.assemblyCatalog.GetAssemblies().Where(x => !x.IsDynamic);
```
khellang: Also, the `.Where(x => !x.IsDynamic)` call is redundant; https://github.com/NancyFx/Nancy/blob/fade252d9aa9c25fe190706fb7ae6b66b0da675c/src/Nancy/AppDomainAssemblyCatalog.cs#L30

samples/Nancy.Demo.Razor.Localization/CustomResourceAssemblyProvider.cs
```diff
@@ -13,16 +11,22 @@
...
         public IEnumerable<Assembly> GetAssembliesToScan()
         {
             return (this.filteredAssemblies ?? (this.filteredAssemblies = GetFilteredAssemblies()));
         }
 
-        private static IEnumerable<Assembly> GetFilteredAssemblies()
+        private IEnumerable<Assembly> GetFilteredAssemblies()
         {
-            return AppDomainAssemblyTypeScanner.Assemblies.Where(x => !x.IsDynamic);
+            return this.assemblyCatalog.GetAssemblies();
```
khellang: Aaaand now we're at a point where I don't see the point in having this method at all :stuck_out_tongue_closed_eyes:

jchannon:
I guess by default it does All, if you want to do something else you use your own?

samples/Nancy.Demo.Razor.Localization/CustomResourceAssemblyProvider.cs
jchannon: I guess by default it does `All`, if you want to do something else you use your own?

thecodejunkie: Should just move the call to `GetAssemblies` to the end of line 24 and kill the method

thecodejunkie:
In reality we could just pipe `this.assemblyCatalog.GetAssemblies()` straight through (it does caching internally per strategy) , but we can keep it like this because it makes no difference really

PR #2225, opened 2016-02-03
Removed ADATS dependency in NancyInternalConfiguration

thecodejunkie:
As part of #2221 

This one was a bit tricky as we could simply not just take a constructor dependency on `ITypeCatalog` in `NancyInternalConfiguration` because of the `Default` member and `WithOverrides`-method. I used the same approach that was taken in the #1846 spike. The `InternalConfiguration` property, on `NancyBoostrapperBase<T>` changed from plain `NancyInternalConfiguration` return type into `Func<ITypeCatalog, NancyInternalConfiguration>` so that we can inject the correct `ITypeCatalog` into it

This also meant that I had to update how the `WithOverrides` method was designed internally. It also meant that I had to rework `ConfigurableBootstrapper` worked. Previously I did modify a `NancyInternalConfiguration` instance, but it can not do that anymore because of the signature change on the bootstrapper. So instead the configurator build up a list of `Action<NancyInternalConfiguration>` which is then applied, one-by-one, on top of the `NancyInternalConfiguration` instance that the `ConfigurableBootstrapper` will end up using.

### DIFF
--- a/src/Nancy/Bootstrapper/NancyInternalConfiguration.cs
+++ b/src/Nancy/Bootstrapper/NancyInternalConfiguration.cs
@@ -3,7 +3,6 @@ namespace Nancy.Bootstrapper
     using System;
     using System.Collections.Generic;
     using System.Linq;
-
     using Nancy.Configuration;
     using Nancy.Culture;
     using Nancy.Diagnostics;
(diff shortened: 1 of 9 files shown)


khellang:
Needs a rebase

thecodejunkie:
Rebased

src/Nancy/Bootstrapper/NancyInternalConfiguration.cs
```diff
@@ -29,59 +28,59 @@ public sealed class NancyInternalConfiguration
...
-                    EnvironmentFactory = typeof(DefaultNancyEnvironmentFactory),
-                    EnvironmentConfigurator = typeof(DefaultNancyEnvironmentConfigurator),
-                    DefaultConfigurationProviders = AppDomainAssemblyTypeScanner.TypesOf<INancyDefaultConfigurationProvider>().ToList(),
-                    SerializerFactory = typeof(DefaultSerializerFactory),
-                    RuntimeEnvironmentInformation = typeof(DefaultRuntimeEnvironmentInformation),
+                    ViewLocator = typeof(DefaultViewLocator),
+                    ViewFactory = typeof(DefaultViewFactory),
+                    ViewResolver = typeof(DefaultViewResolver),
+                    ViewCache = typeof(DefaultViewCache),
+                    ViewLocationProvider = typeof(FileSystemViewLocationProvider),
```
damianh: Only thing I can that happened here is alphabetical sorting...

khellang:
YOLO! :grin:

PR #2239, opened 2016-02-04
Remove adats dependency in razor tests

thecodejunkie:
Part of #2221

### DIFF
--- a/src/Nancy.ViewEngines.Razor/ClrTypeResolver.cs
+++ b/src/Nancy.ViewEngines.Razor/ClrTypeResolver.cs
@@ -17,6 +17,8 @@
     internal abstract class ClrTypeResolver<TSymbolType, TSymbol>
         where TSymbol : SymbolBase<TSymbolType>
     {
+        private readonly IAssemblyCatalog assemblyCatalog;
+
         private readonly TSymbolType identifier;
         private readonly TSymbolType keyword;
         private readonly TSymbolType dot;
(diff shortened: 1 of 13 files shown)


src/Nancy.ViewEngines.Razor/ClrTypeResolver.cs
```diff
@@ -182,8 +186,12 @@ private Type ResolveTypeByName(string typeName)
...
             return Type.GetType(typeName)
                    ?? ResolvePrimitiveType(typeName)
-                   ?? AppDomainAssemblyTypeScanner.Types.FirstOrDefault(t => t.FullName == typeName)
-                   ?? AppDomainAssemblyTypeScanner.Types.FirstOrDefault(t => t.Name == typeName);
+                   ?? this.ResolveTypeFromAssemblyCatalog(typeName);
+        }
+
+        private Type ResolveTypeFromAssemblyCatalog(string typeName)
+        {
+            return this.assemblyCatalog.GetAssemblies().Select(assembly => assembly.GetType(typeName)).FirstOrDefault(type => type != null);
```
khellang: Will this work correctly? :confused:

thecodejunkie: It should :smile: 
Since we only get the qualified name (namespace + type) we cannot just use `Type.GetType` as it wants us to send in an assembly-qualified name. Iterating over the assemblies and using `assembly.GetType` does the same thing but without having to provide the assembly name.

Tests pass (we have razor tests that explicitly looks for a type in a satellite assembly) and I tried the razor routes in the aspnet host demo

PR #2241, opened 2016-02-05
Removed ADATS usage in the Razor engine

thecodejunkie:
Part of #2221


khellang:
Mmmmm... `this` :yum:

PR #2242, opened 2016-02-05
Removed ADATS usage in the ConfigurableBootstrapper

thecodejunkie:
Part of #2221 

The removed code should no longer be needed. It was put in place to resolve the case where the assembly under test was not loaded in during runtime, causing it to not be part of the `AppDomainAssemblyTypeScanner` cache. It could cause things such as modules not being discovered and test would fail because the route could not be hit. This should be resolved by the recent changed to the `IAssemblyCatalog` (see #2234).

I re-ran all tests after the code was removed and they all turned green

### DIFF
--- a/src/Nancy.Testing/ConfigurableBootstrapper.cs
+++ b/src/Nancy.Testing/ConfigurableBootstrapper.cs
@@ -70,11 +70,6 @@ namespace Nancy.Testing
             this.applicationStartupActions = new List<Action<TinyIoCContainer, IPipelines>>();
             this.requestStartupActions = new List<Action<TinyIoCContainer, IPipelines, NancyContext>>();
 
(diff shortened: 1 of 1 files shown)


src/Nancy.Testing/ConfigurableBootstrapper.cs
```diff
@@ -241,7 +197,7 @@ private IEnumerable<Type> Resolve<T>()
         /// <summary>
         /// Nancy internal configuration
         /// </summary>
-        protected override sealed NancyInternalConfiguration InternalConfiguration
+        protected sealed override NancyInternalConfiguration InternalConfiguration
```
khellang: Phew! :sweat_smile::sweat_drops:

PR #2243, opened 2016-02-05
Move Web.configs into MSBuild folders

khellang:
Moved Web.config into MSBuild folders


thecodejunkie:
Hold of on this.. we're investigating something else

Issue #2220, opened 2016-02-03
Get Nancy compiling on dotnet54/dnxcore50


Yantrio:
## Proposal
As part of the latest microsoft movement, it would be nice to get nancy working on the new dotnet environment. 
This was started with pull requests #2211 and #2212 to get things compiling using the project.json project system. Now we need to get some good work done to get this working on dotnet54.

### Approach
So far, there has been analysis completed on the nancy package to see what is/isn't available in the latest runtimes, and there are some areas that need replacing.
Due to the amount of work that needs doing, It's been suggested by @thecodejunkie that we try to get the project compiling first, then approach tests later on once we are happy and settled with the replacement code.

Most areas of the code should remain the same for the full CLR, and ifdefs should be used to swap out logic with comparable parts in the new environment.

### Why choose dotnet54 and not dnxcore50?
Dotnet54 is intended for use by microsoft by libraries and not the consuming applications, more information can be found here : aspnet/announcements#98

### Things that need tackling to get things compiling

- [x] Addition of dotnet54 to project.json file [@yantrio #2248]
- [x] Removal of SerializableAttribute (should be simple, but not sure on repercussions) [@yantrio #2255]
- [x] RNGCryptoServiceProvider -> RandomNumberGenerator.Create(), (this is currently what the Microsoft Identity packages use.) [@yantrio #2252]
- [x] BootstrapperException Ctor (SerializationInfo info, StreamingContext context), SerializationInfo isn't available, suitable replacement needed [@yantrio #2253]
- [x] DefaultObjectSerializer uses BinaryFormatter, it's not available, suitable replacement needed [@jchannon #2264]
- [x] Classes inheriting Stream need to use the suitable overrides, signatures have changed [@yantrio #2280 ]
- [x] Url class implements ICloneable, it's not available, is this an issue? [@yantrio #2251]
- [x] Most reflection needs tweaking/replacing. [@yantrio #2274]
- [x] CultureInfo caching (ie BuiltInCultureConventions class) needs replacing, suitable replacement needed [@jchannon #2258]
- [x] RijndaelEncryptionProvider makes use of RijndaelManaged class, not available. Replace or remove? (replace) [@yantrio #2254]
- [x] MemoryStream.GetBuffer is not available, suitable replacement needed [@yantrio #2249]
- [x] Uri.IsHexDigit/Uri.FromHex() is not available, suitable replacement needed. [@yantrio #2250]

### Things to note:
[...]

PR #2249, opened 2016-02-07
MemoryStream.GetBuffer -> MemoryStream.TryGetBuffer

Yantrio:
Part of #2220 

Replaced Usage of MemoryStream.GetBuffer to MemoryStream.TryGetBuffer

For now, I couldn't find a suitable exception to throw if TryGetBuffer fails, what would be the best approach for this? for now I have just returned an empty object however I think this could be a bit of an issue.

### DIFF
--- a/src/Nancy/DefaultObjectSerializer.cs
+++ b/src/Nancy/DefaultObjectSerializer.cs
@@ -1,5 +1,6 @@
 namespace Nancy
 {
+    using Extensions;
     using System;
     using System.IO;
     using System.Runtime.Serialization;
@@ -25,9 +26,8 @@ namespace Nancy
             {
                 formatter.Serialize(outputStream, sourceObject);
 
-                var outputBytes = outputStream.GetBuffer();
-
-                return Convert.ToBase64String(outputBytes, 0, (int)outputStream.Length);
+                var buffer = outputStream.GetBufferSegment();
+                return Convert.ToBase64String(buffer.Array, buffer.Offset, buffer.Count);
             }
         }
 
--- a/src/Nancy/Extensions/MemoryStreamExtensions.cs
+++ b/src/Nancy/Extensions/MemoryStreamExtensions.cs
@@ -1,0 +1,31 @@
+﻿namespace Nancy.Extensions
+{
+    using System;
+    using System.IO;
+
+    internal static class MemoryStreamExtensions
+    {
+        public static ArraySegment<byte> GetBufferSegment(this MemoryStream stream)
+        {
+#if DOTNET5_4
+            ArraySegment<byte> buffer;
+            if (stream.TryGetBuffer(out buffer))
+            {
+                return buffer;
+            }
+#endif
+            var bytes = stream.GetBytes();
+
+            return new ArraySegment<byte>(bytes, 0, bytes.Length);
+        }
+
+        private static byte[] GetBytes(this MemoryStream stream)
+        {
+#if DOTNET5_4
+            return stream.ToArray(); // This is all we have if TryGetBuffer fails in GetBufferSegment
+#else
+            return stream.GetBuffer();
+#endif
+        }
+    }
+}
--- a/src/Nancy/project.json
+++ b/src/Nancy/project.json
@@ -1,4 +1,4 @@
-﻿{
+{
     "version": "2.0.0-*",
 
     "description": "Nancy is a lightweight web framework for the .Net platform, inspired by Sinatra. Nancy aim at delivering a low ceremony approach to building light, fast web applications.",
@@ -7,12 +7,10 @@
     "projectUrl": "http://nancyfx.org",
     "licenseUrl": "https://github.com/NancyFx/Nancy/blob/master/license.txt",
     "iconUrl": "http://nancyfx.org/nancy-nuget.png",
-
     "resource": [
         "Diagnostics/Resources/**/*.*",
         "Diagnostics/Views/**/*.*"
     ],
-
     "frameworks": {
         "dotnet5.4": {
           "dependencies": {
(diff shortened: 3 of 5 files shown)


src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -25,8 +25,13 @@ public string Serialize(object sourceObject)
             {
                 formatter.Serialize(outputStream, sourceObject);
 
+#if DOTNET5_4
+                ArraySegment<byte> buffer;
+                var gotBuffer = outputStream.TryGetBuffer(out buffer);
+                var outputBytes = gotBuffer ? buffer.Array : new byte[] { };
```
davidfowl: This isn't right. You need to use the array segment length and offset. Better to always use an array segment:

```C#
var buffer = default(ArraySegment<byte>);

#if DOTNET5_4
    outputStream.TryGetBuffer(out buffer);
#else
    var outputBytes = outputStream.GetBuffer();
    buffer = new ArraySegment<byte>(outputBytes, 0, outputBytes.Length);
#endif
    return Convert.ToBase64String(buffer.Array, buffer.Offset, buffer.Count);
```

Something like that

Yantrio: Fixed as you suggested @davidfowl :)

src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -25,9 +25,15 @@ public string Serialize(object sourceObject)
             {
                 formatter.Serialize(outputStream, sourceObject);
 
-                var outputBytes = outputStream.GetBuffer();
+                var buffer = default(ArraySegment<byte>);
 
-                return Convert.ToBase64String(outputBytes, 0, (int)outputStream.Length);
+#if DOTNET5_4
```
khellang: Can we roll all of this into a `GetBufferSegment` extension method and reuse that in `HttpUtility` as well?

khellang: The method could then be something like

```csharp
internal static class MemoryStreamExtensions
{
    public static ArraySegment<byte> GetBufferSegment(this MemoryStream stream)
    {
#if DOTNET5_4
        ArraySegment<byte> buffer;
        if (stream.TryGetBuffer(out buffer))
        {
            return buffer;
        }
#endif

        var bytes = stream.GetBytes();

        return new ArraySegment<byte>(bytes, 0, bytes.Length);
    }

    private static byte[] GetBytes(this MemoryStream stream)
    {
#if DOTNET5_4
        return stream.ToArray();
#else
        return stream.GetBuffer();
#endif
    }
}
```

Yantrio: that sounds like a good idea :) i'll implement that in the morning.

khellang:
:watch:

Yantrio:
Added extension method as discussed with @khellang

src/Nancy/Extensions/MemoryStreamExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+    using System.IO;
+
+    internal static class MemoryStreamExtensions
+    {
+        public static ArraySegment<byte> GetBufferSegment(this MemoryStream stream)
+        {
+            var buffer = default(ArraySegment<byte>);
+
+#if DOTNET5_4
+            stream.TryGetBuffer(out buffer);
```
khellang: This falls over if `TryGetBuffer` fails...

Yantrio: I assume if TryGetBuffer fails, that .ToArray() would also fail. I don't see much of a way around this.

khellang: Don't ever *assume* anything.

```csharp
public class Program
{
    public static void Main()
    {
        var buffer = new[] { 'H', 'e', 'l', 'l', 'o' }.Select(x => (byte) x).ToArray();

        var stream = new MemoryStream(buffer);

        ArraySegment<byte> segment;
        if (!stream.TryGetBuffer(out segment))
        {
            Console.WriteLine("TryGetBuffer failed. Falling back to ToArray...");

            var bytes = stream.ToArray();

            segment = new ArraySegment<byte>(bytes, 0, bytes.Length);
        }

        if (segment != default(ArraySegment<byte>))
        {
            Console.WriteLine(new string(segment.Array.Select(x => (char) x).ToArray()));
            Console.WriteLine("Well look at that. It worked!");
        }
    }
}
```

Output:

> TryGetBuffer failed. Falling back to ToArray...
> Hello
> Well look at that. It worked!

khellang:
Build is failing... 

> error CS1061: 'System.IO.MemoryStream' does not contain a definition for 'GetBufferSegment' and no extension method 'GetBufferSegment' accepting a first argument of type 'System.IO.MemoryStream'

:watch:

Yantrio:
fixed the build issue, realised im not building csproj's before I pushed.

jchannon:
Thats what CI is for :)

Yantrio:
:watch:

jchannon:
LGTM

jchannon:
i think this needs a rebase

Yantrio:
rebased.

jchannon:
I wonder if we should `throw new InvalidOperationException ` if `TryGetBuffer` fails?

Yantrio:
if TryGetBuffer fails, it falls back to ToArray() on dotnet54 or .GetBuffer() on full clr.

jchannon:
I know, my point is the empty object as the end result

Yantrio:
I can't see where it would return an empty object.

src/Nancy/project.json
```diff
@@ -7,12 +7,10 @@
     "projectUrl": "http://nancyfx.org",
     "licenseUrl": "https://github.com/NancyFx/Nancy/blob/master/license.txt",
     "iconUrl": "http://nancyfx.org/nancy-nuget.png",
-
```
khellang: I liked these :disappointed: I think @thecodejunkie also did, since he added them...

jchannon:
If there was nothing in the stream but as i sayI think we're protected against that in the defaultobjectserilaizer already.

PR #2250, opened 2016-02-07
Implemented IsHexDigit/FromHex in HttpEncoder

Yantrio:
Part of #2220, 
used an IfDef instead of replacement incase ms decides to update something.

### DIFF
--- a/src/Nancy/Helpers/HttpEncoder.cs
+++ b/src/Nancy/Helpers/HttpEncoder.cs
@@ -1,7 +1,7 @@
 //
 // Authors:
 //   Patrik Torstensson (Patrik.Torstensson@labs2.com)
-//   Wictor Wil�n (decode/encode functions) (wictor@ibizkit.se)
+//   Wictor Wilén (decode/encode functions) (wictor@ibizkit.se)
 //   Tim Coleman (tim@timcoleman.com)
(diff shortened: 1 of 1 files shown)


davidfowl:
Why ifdef at all? Just use the custom implementation.

khellang:
Yeah, let's just keep the #ifdefs to an absolute minimum. If there's a bug in the implementation, at least it'll be consistent across platforms :wink:

khellang:
:watch:

Yantrio:
fixed.

src/Nancy/Helpers/HttpEncoder.cs
```diff
@@ -1,7 +1,7 @@
 //
 // Authors:
 //   Patrik Torstensson (Patrik.Torstensson@labs2.com)
-//   Wictor Wil�n (decode/encode functions) (wictor@ibizkit.se)
+//   Wictor Wil�n (decode/encode functions) (wictor@ibizkit.se)
```
khellang: Pfff. Who needs tooling? :stuck_out_tongue_closed_eyes:

Yantrio: i spent 1 morning with vs2015, and it did crap like this, im going back to atom.

PR #2251, opened 2016-02-07
Ifdefed out usage of ICloneable interface in Url class

Yantrio:
Part of #2220

We don't seem to be using the ICloneable method Clone() too often, but i've ifdefed it instead of replacement for now, if people want, we can remove it completely.

### DIFF
--- a/src/Nancy/Url.cs
+++ b/src/Nancy/Url.cs
@@ -122,17 +122,6 @@ namespace Nancy
                 .ToString();
         }
 
-#if !DOTNET5_4
-        /// <summary>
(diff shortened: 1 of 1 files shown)


khellang:
Let's just remove the #ifdef'ing and remove the interface altogether. I can't imagine someone taking an `ICloneable` argument or using it as a generic constraint. Let's just keep the method as is.

thecodejunkie:
Yeah I second that we just kill the interface. YAGNI and all that :D

Yantrio:
The interface has now been killed :)

PR #2253, opened 2016-02-07
ifdefed out usages of SerializationInfo BootstrapperException ctors

Yantrio:
part of #2220


khellang:
I have my doubts about whether it's worth keeping these ctors around at all, but anyway... Nice work :heart:

Issue #2257, opened 2016-02-08
Add Culture support to INancyRuntimeEnvironment


jchannon:
Currently CultureConventions are supported and cultures are validated with a cache gained from calling [CultureInfo.GetCultures](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Conventions/BuiltInCultureConventions.cs#L24)

The proposal is to replace that call by using the `INancyRuntimeEnvironment`.  Nancy will support `en-us` by default however APIs will be made available for the user to add their own supported cultures.

PR #2258, opened 2016-02-08
Moved allowed cultures to a configuration class

jchannon:
Addresses #2257


src/Nancy/Culture/DefaultCultureService.cs
```diff
@@ -31,7 +34,7 @@ public CultureInfo DetermineCurrentCulture(NancyContext context)
 
             foreach (var convention in this.cultureConventions)
             {
-                culture = convention(context);
+                culture = convention(context, this.configuration);
```
jchannon: Only thing I'm a bit unsure of is this. The `ICultureService` interface contract is `CultureInfo DetermineCurrentCulture(NancyContext context)` but I use a `CultureConfiguration` that is injected in and is required to determine current culture however I think it would hard to pass that configuration as a method arg as it is called from `DefaultNancyContextFactory` using `ICultureService`. There's a couple of options but I don't like any of them

src/Nancy/CultureConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Culture configuration
+    /// </summary>
+    public class CultureConfiguration
+    {
+        /// <summary>
+        /// A default instance of the <see cref="CultureConfiguration"/> class
+        /// </summary>
+        public static readonly CultureConfiguration Default = new CultureConfiguration(cultureNames: new[] { "en-us" });
```
Yantrio: why used the named parameter here when there is only 1?

src/Nancy/CultureConfigurationExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// Configures <see cref="CultureConfiguration"/>
+        /// </summary>
+        /// <param name="environment">An <see cref="INancyEnvironment"/> that should be configured.</param>
+        /// <param name="cultures">Cultures that the application can accept</param>
+        public static void Cultures(this INancyEnvironment environment, IEnumerable<string> cultures)
+        {
+            environment.AddValue(new CultureConfiguration(
+                cultureNames:cultures));
```
Yantrio: any reason to put this on a new line?

jchannon:
@yantrio because @thecodejunkie is a weirdo

jchannon:
@thecodejunkie what about deleting ICultureService and putting its method into CultureConfiguration ?

thecodejunkie:
@jchannon Been giving this some thought over the evening and perhaps this all should be moved to a `Globalization` config class/extension etc and have the parameter be called something along the lines of `supportedCultures` ? That way we can add more related stuff in there later on, `CultureConfiguration` and `Culture` extension is perhaps too specific in this case?

thecodejunkie:
I think the `ICultureService` should stay where it is

jchannon:
@thecodejunkie renamed, I'm still not sure about `ICultureService` but Ill bow down to your wiseness

thecodejunkie:
What's wrong with the service? The config is just that.. config values. it does nothing with them.. the `ICultureService` combines two things; the conventions and the cultures and returns something?

jchannon:
I think the service method probably needs the globalisation config passed
to the method?

thecodejunkie:
@jchannon you're going to have to paint a clearer picture for me here mate - I've not looked at this code in quite a while :)

I assume you mean something along the lines that the service needs to call the culture conventions and pass in the supported culture names to them?

src/Nancy/Conventions/BuiltInCultureConventions.cs
```diff
@@ -52,8 +41,9 @@ public static CultureInfo FormCulture(NancyContext context)
         /// Checks to see if the first argument in the Path can be used to make a CultureInfo.
         /// </summary>
         /// <param name="context">NancyContext</param>
+        /// <param name="configuration">Culture configuration that contains allowed cultures</param>
         /// <returns>CultureInfo if found in Path otherwise null</returns>
-        public static CultureInfo PathCulture(NancyContext context)
+        public static CultureInfo PathCulture(NancyContext context, GlobalizationConfiguration configuration)
```
thecodejunkie: In two minds about passing in the entire configuration class vs just the supported culture names. But maybe it's fine

src/Nancy/Culture/DefaultCultureService.cs
```diff
@@ -10,14 +10,17 @@
...
         private readonly CultureConventions cultureConventions;
+        private readonly GlobalizationConfiguration configuration;
 
         /// <summary>
         /// Creates a new instance of DefaultCultureService
         /// </summary>
         /// <param name="cultureConventions">CultureConventions to use for determining culture</param>
-        public DefaultCultureService(CultureConventions cultureConventions)
+        /// <param name="configuration">GlobalizationConfiguration containing allowed cultures</param>
+        public DefaultCultureService(CultureConventions cultureConventions, GlobalizationConfiguration configuration)
```
thecodejunkie: How can the service take a dependency on `GlobalizationConfiguration` ? That type isn't registered in the container in a reliable way. It would work for TinyIoC because of the auto-registration but for everything else it would fail. It would be better to take the dependency on `INancyEnvironment` and pull it out instead

src/Nancy/Conventions/BuiltInCultureConventions.cs
```diff
@@ -15,29 +15,18 @@
     public static class BuiltInCultureConventions
```
khellang: Why don't we make this an instance and pass in `GlobalizationConfiguration` instead of passing it to every method?

jchannon: Cause they are all used in a List<Func<

khellang: Exactly. Then we don't have to alter the `Func` :smile:

jchannon: @thecodejunkie thoughts?

thecodejunkie:
YOLO

PR #2098, opened 2015-11-01
[WIP] Upgrade to xunit 2.1

damianh:
- Needed for improved and better async support
 - Will carry forward into dnx world later

(Will Travis accept it ?? :)

PR #2259, opened 2016-02-09
Update testing tools

thecodejunkie:
This is mostly a mirror of what @damianh did in #2098 but on the latest folder structure + using the lastest fake it easy

Short summary

- Updated to Xunit 2.1
- Updated to Xunit.runner.visualstudio 2.1.0
- Updated to FakeItEasy 2.0.0-beta011
- Added `AssemblyInfo.cs` to all test projects to disable parallel test execution (for the time being) as there are some issues (like mutatable static state in the DotLiquid parser - not out engine)
- Updated build scripts
- Removed Xunit.extensions as it's no longer needed

Updating to Xunit 2.1 is a pre-requisite for us to run tests using project.json as it's the only version that has a dnx runner


test/Nancy.Tests/Unit/ModelBinding/DefaultBinderFixture.cs
```diff
@@ -97,7 +97,7 @@ public void Should_not_throw_if_body_deserializer_fails_and_IgnoreErrors_is_true
             var config = new BindingConfig { IgnoreErrors = true };
 
             // When, Then
-            Assert.DoesNotThrow(() => binder.Bind(context, this.GetType(), null, config));
+            Record.Exception(() => binder.Bind(context, this.GetType(), null, config)).ShouldBeNull();
```
khellang: Do we need `Record.Exception(/* ... */).ShouldBeNull()` here?

thecodejunkie: I guess not.. it it throws the test will fail anyway

PR #2264, opened 2016-02-09
Removed binaryformatter replaced with json serializer

jchannon:
address #2220

### DIFF
--- a/samples/Nancy.Demo.Hosting.Aspnet/DemoBootstrapper.cs
+++ b/samples/Nancy.Demo.Hosting.Aspnet/DemoBootstrapper.cs
@@ -66,7 +66,7 @@
 
             this.Conventions.StaticContentsConventions.Add(StaticContentConventionBuilder.AddDirectory("moo", "Content"));
 
-            CookieBasedSessions.Enable(pipelines);
+            CookieBasedSessions.Enable(pipelines, this.AssemblyCatalog);
 
             pipelines.AfterRequest += (ctx) =>
             {
--- a/src/Nancy.MSBuild/Nancy.csproj
+++ b/src/Nancy.MSBuild/Nancy.csproj
@@ -1376,6 +1376,12 @@
     <Compile Include="..\Nancy\Cryptography\AesEncryptionProvider.cs">
       <Link>Cryptography\AesEncryptionProvider.cs</Link>
     </Compile>
+    <Compile Include="..\Nancy\Json\SimpleJson.cs">
+      <Link>Json\SimpleJson.cs</Link>
+    </Compile>
+    <Compile Include="..\Nancy\Extensions\ObjectExtensions.cs">
+      <Link>Extensions\ObjectExtensions.cs</Link>
+    </Compile>
   </ItemGroup>
   <ItemGroup>
     <EmbeddedResource Include="..\Nancy\Diagnostics\Views\Dashboard.sshtml">
--- a/src/Nancy/DefaultObjectSerializer.cs
+++ b/src/Nancy/DefaultObjectSerializer.cs
@@ -3,11 +3,28 @@ namespace Nancy
     using Extensions;
     using System;
     using System.IO;
+    using System.Linq;
+    using System.Reflection;
     using System.Runtime.Serialization;
-    using System.Runtime.Serialization.Formatters.Binary;
+    using System.Text;
+    using Nancy.Extensions;
 
+    /// <summary>
+    /// Serializes/Deserializes objects for sessions
+    /// </summary>
     public class DefaultObjectSerializer : IObjectSerializer
     {
+        private readonly IAssemblyCatalog assemblyCatalog;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DefaultObjectSerializer"/> class.
+        /// </summary>
+        /// <param name="assemblyCatalog"></param>
+        public DefaultObjectSerializer(IAssemblyCatalog assemblyCatalog)
+        {
+            this.assemblyCatalog = assemblyCatalog;
+        }
+
         /// <summary>
         /// Serialize an object
         /// </summary>
@@ -17,18 +34,35 @@ namespace Nancy
         {
             if (sourceObject == null)
             {
-                return String.Empty;
+                return string.Empty;
             }
 
-            var formatter = new BinaryFormatter();
+            dynamic serializedObject = (sourceObject is string)
+                ? sourceObject
+                : this.AddTypeInformation(sourceObject);
 
-            using (var outputStream = new MemoryStream())
+            var json = SimpleJson.SerializeObject(serializedObject);
+            return Convert.ToBase64String(Encoding.UTF8.GetBytes(json));
+        }
+
+        private dynamic AddTypeInformation(object sourceObject)
+        {
+            var sourceType = this.assemblyCatalog.GetAssemblies().Select(assembly => assembly.GetType(sourceObject.GetType().FullName)).FirstOrDefault(type => type != null);
+            if (sourceType == null)
             {
-                formatter.Serialize(outputStream, sourceObject);
-
-                var buffer = outputStream.GetBufferSegment();
-                return Convert.ToBase64String(buffer.Array, buffer.Offset, buffer.Count);
+                throw new SerializationException("Unable to find type " + sourceObject.GetType() + " in its assembly to serialize");
             }
+
+            dynamic serializedObject = null;
+
+            var assemblyQualifiedName = sourceType.GetTypeInfo().AssemblyQualifiedName;
+            if (!string.IsNullOrWhiteSpace(assemblyQualifiedName))
+            {
+                serializedObject = sourceObject.ToDynamic();
+                serializedObject.TypeObject = assemblyQualifiedName;
+            }
+
+            return serializedObject;
         }
 
         /// <summary>
@@ -46,13 +80,15 @@ namespace Nancy
             try
             {
                 var inputBytes = Convert.FromBase64String(sourceString);
+                var json = Encoding.UTF8.GetString(inputBytes);
 
-                var formatter = new BinaryFormatter();
-
-                using (var inputStream = new MemoryStream(inputBytes, false))
+                if (ContainsTypeDescription(json))
                 {
-                    return formatter.Deserialize(inputStream);
+                    dynamic serializedObject = SimpleJson.DeserializeObject(json);
+                    var actual = SimpleJson.DeserializeObject(json, Type.GetType(serializedObject.TypeObject));
+                    return actual;
                 }
+                return SimpleJson.DeserializeObject(json);
             }
             catch (FormatException)
             {
@@ -62,10 +98,15 @@ namespace Nancy
             {
                 return null;
             }
-	    catch (IOException)
-	    {
-		return null;
-	    }
+            catch (IOException)
+            {
+                return null;
+            }
+        }
+
+        private static bool ContainsTypeDescription(string json)
+        {
+            return json.Contains("TypeObject");
         }
     }
 }
--- a/src/Nancy/Diagnostics/DiagnosticsHook.cs
+++ b/src/Nancy/Diagnostics/DiagnosticsHook.cs
@@ -32,7 +32,7 @@ namespace Nancy.Diagnostics
         /// Enables the diagnostics dashboard and will intercept all requests that are passed to
         /// the condigured paths.
         /// </summary>
-        public static void Enable(IPipelines pipelines, IEnumerable<IDiagnosticsProvider> providers, IRootPathProvider rootPathProvider, IRequestTracing requestTracing, NancyInternalConfiguration configuration, IModelBinderLocator modelBinderLocator, IEnumerable<IResponseProcessor> responseProcessors, IEnumerable<IRouteSegmentConstraint> routeSegmentConstraints, ICultureService cultureService, IRequestTraceFactory requestTraceFactory, IEnumerable<IRouteMetadataProvider> routeMetadataProviders, ITextResource textResource, INancyEnvironment environment, IRuntimeEnvironmentInformation runtimeEnvironmentInformation, ITypeCatalog typeCatalog)
+        public static void Enable(IPipelines pipelines, IEnumerable<IDiagnosticsProvider> providers, IRootPathProvider rootPathProvider, IRequestTracing requestTracing, NancyInternalConfiguration configuration, IModelBinderLocator modelBinderLocator, IEnumerable<IResponseProcessor> responseProcessors, IEnumerable<IRouteSegmentConstraint> routeSegmentConstraints, ICultureService cultureService, IRequestTraceFactory requestTraceFactory, IEnumerable<IRouteMetadataProvider> routeMetadataProviders, ITextResource textResource, INancyEnvironment environment, IRuntimeEnvironmentInformation runtimeEnvironmentInformation, ITypeCatalog typeCatalog, IAssemblyCatalog assemblyCatalog)
         {
             var diagnosticsConfiguration =
                 environment.GetValue<DiagnosticsConfiguration>();
@@ -57,7 +57,7 @@ namespace Nancy.Diagnostics
                 new RouteResolverTrie(new TrieNodeFactory(routeSegmentConstraints)),
                 environment);
 
-            var serializer = new DefaultObjectSerializer();
+            var serializer = new DefaultObjectSerializer(assemblyCatalog);
 
             pipelines.BeforeRequest.AddItemToStartOfPipeline(
                 new PipelineItem<Func<NancyContext, Response>>(
--- a/src/Nancy/Extensions/ObjectExtensions.cs
+++ b/src/Nancy/Extensions/ObjectExtensions.cs
@@ -1,0 +1,29 @@
+namespace Nancy.Extensions
+{
+    using System.Collections.Generic;
+    using System.Dynamic;
+    using System.Reflection;
+
+    /// <summary>
+    /// Contains extensions to <see cref="object"/> class.
+    /// </summary>
+    public static class ObjectExtensions
+    {
+        /// <summary>
+        /// Convert an object to a dynamic type
+        /// </summary>
+        /// <param name="value">An object to convert to dynamic</param>
+        /// <returns>Returns a dynamic version of the specified type</returns>
+        public static dynamic ToDynamic(this object value)
+        {
+            var expando = new ExpandoObject() as IDictionary<string, object>;
+
+            foreach (var property in value.GetType().GetTypeInfo().DeclaredProperties)
+            {
+                expando.Add(property.Name, property.GetValue(value));
+            }
+
+            return (ExpandoObject)expando;
+        }
+    }
+}
--- a/src/Nancy/Session/CookieBasedSessions.cs
+++ b/src/Nancy/Session/CookieBasedSessions.cs
@@ -90,12 +90,13 @@ namespace Nancy.Session
         /// </summary>
         /// <param name="pipelines">Application pipelines</param>
         /// <param name="cryptographyConfiguration">Cryptography configuration</param>
+        /// <param name="assemblyCatalog">Assembly catalog</param>
         /// <returns>Formatter selector for choosing a non-default serializer</returns>
-        public static IObjectSerializerSelector Enable(IPipelines pipelines, CryptographyConfiguration cryptographyConfiguration)
+        public static IObjectSerializerSelector Enable(IPipelines pipelines, CryptographyConfiguration cryptographyConfiguration, IAssemblyCatalog assemblyCatalog)
         {
             var cookieBasedSessionsConfiguration = new CookieBasedSessionsConfiguration(cryptographyConfiguration)
             {
-                Serializer = new DefaultObjectSerializer()
+                Serializer = new DefaultObjectSerializer(assemblyCatalog)
             };
             return Enable(pipelines, cookieBasedSessionsConfiguration);
         }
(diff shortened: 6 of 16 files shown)


src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -16,14 +16,14 @@ public string Serialize(object sourceObject)
...
+                return string.Empty;
             }
 
-            var formatter = new BinaryFormatter();
-
             using (var outputStream = new MemoryStream())
+            using (var writer = new StreamWriter(outputStream))
             {
-                formatter.Serialize(outputStream, sourceObject);
+                var jsonSerializer = new JsonSerializer(new JavaScriptSerializer());
```
khellang: Should we really new this up every time?

src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -47,11 +47,11 @@ public object Deserialize(string sourceString)
             {
                 var inputBytes = Convert.FromBase64String(sourceString);
 
-                var formatter = new BinaryFormatter();
-
                 using (var inputStream = new MemoryStream(inputBytes, false))
+                using (var streamreader = new StreamReader(inputStream))
                 {
-                    return formatter.Deserialize(inputStream);
+                    var deserialzier = new JsonDeserializer(new JavaScriptSerializer());
```
khellang: Same here...

src/Nancy/DefaultObjectSerializer.cs
khellang: Also, why haven't we #ifdef'ed this to use `BinaryFormatter` where available?

jchannon: Because JS will work for both I guess

src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -3,10 +3,21 @@ namespace Nancy
     using System;
     using System.IO;
     using System.Runtime.Serialization;
-    using System.Runtime.Serialization.Formatters.Binary;
+    using Nancy.Json;
 
     public class DefaultObjectSerializer : IObjectSerializer
     {
+        private readonly JsonSerializer jsonSerializer;
+        private readonly JsonDeserializer jsonDeserialzier;
```
khellang: Y NO STATIC?

src/Nancy.MSBuild/Nancy.csproj
```diff
@@ -1531,6 +1537,9 @@
       <Link>Diagnostics\Views\login.sshtml</Link>
     </EmbeddedResource>
   </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
```
khellang: I can't see this anywhere?

src/Nancy/Diagnostics/DiagnosticsHook.cs
```diff
@@ -57,7 +57,7 @@ public static void Enable(IPipelines pipelines, IEnumerable<IDiagnosticsProvider
                 new RouteResolverTrie(new TrieNodeFactory(routeSegmentConstraints)),
                 environment);
 
-            var serializer = new DefaultObjectSerializer();
+            var serializer = new DefaultObjectSerializer(new AppDomainAssemblyCatalog());
```
khellang: Should this be hardcoded?

src/Nancy/Session/CookieBasedSessions.cs
```diff
@@ -95,7 +95,7 @@ public static IObjectSerializerSelector Enable(IPipelines pipelines, Cryptograph
         {
             var cookieBasedSessionsConfiguration = new CookieBasedSessionsConfiguration(cryptographyConfiguration)
             {
-                Serializer = new DefaultObjectSerializer()
+                Serializer = new DefaultObjectSerializer(new AppDomainAssemblyCatalog())
```
khellang: Again, I'm not sure this should be hard coded. Also, what happens to caching here?

src/Nancy/Diagnostics/DiagnosticsHook.cs
jchannon: Don't see why not if the newing up of DefaultObjectSerializer is.

test/Nancy.Tests/Unit/Diagnostics/DiagnosticsHookFixture.cs
```diff
@@ -21,7 +23,10 @@ public class DiagnosticsHookFixture
         public DiagnosticsHookFixture()
         {
             this.cryptoConfig = CryptographyConfiguration.Default;
-            this.objectSerializer = new DefaultObjectSerializer();
+            var ass = Assembly.GetExecutingAssembly();
```
khellang: `ass`?

jchannon: you don't like ass?

test/Nancy.Tests/Unit/Sessions/DefaultSessionObjectFormatterFixture.cs
```diff
@@ -68,11 +72,16 @@ public void Should_return_null_when_deserializing_empty_string()
...
-            public int IntValue { get; private set; }
+            public int IntValue { get;  set; }
+
+            public bool BoolValue { get;  set; }
 
-            public bool BoolValue { get; private set; }
+            public string StringValue { get;  set; }
 
-            public string StringValue { get; private set; }
+            public Payload()
```
khellang: Redundant...

jchannon: no, needed for serialization

test/Nancy.Tests/Unit/Diagnostics/CustomInteractiveDiagnosticsFixture.cs
```diff
@@ -26,7 +28,10 @@ public class CustomInteractiveDiagnosticsHookFixture
         public CustomInteractiveDiagnosticsHookFixture()
         {
             this.cryptoConfig = CryptographyConfiguration.Default;
-            this.objectSerializer = new DefaultObjectSerializer();
+            var ass = Assembly.GetExecutingAssembly();
```
khellang: I don't think this'll work well on core.Let's use `typeof(CustomInteractiveDiagnosticsHookFixture).GetTypeInfo().Assembly` or something.

test/Nancy.Tests/Unit/Sessions/DefaultSessionObjectFormatterFixture.cs
khellang: An it wasn't before? This means that everyone needs to add this.

jchannon: Its a test class used for testing!

Yantrio: I think this brings the serialization in line with every other serializer, the need for a parameterless ctor is almost expected now-a-days with everything using reflection.

src/Nancy/Diagnostics/DiagnosticsHook.cs
khellang: Because this will have to be replaced on other platforms, like DNX. There's also a caching concern here.

test/Nancy.Tests/Unit/Sessions/DefaultSessionObjectFormatterFixture.cs
khellang: > Its a test class used for testing!. Serialization needs an empty constructor

Sure, in this case. But people use this in production, right? This should reflect production code. Which means if you had to add it, other people will probably have to add it as well. So this is a breaking change.

Yantrio: Sure, it's a breaking change, but i think it's an acceptable one.

khellang: > Sure, it's a breaking change, but i think it's an acceptable one.

I'm not arguing that. It's just that we're making breaking changes left and right without marking them as such, or without even realizing that we're breaking stuff.

src/Nancy/Extensions/ObjectExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// Convert an object to a dynamic type
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public static dynamic ToDynamic(this object value)
+        {
+            IDictionary<string, object> expando = new ExpandoObject();
+
+            foreach (PropertyDescriptor property in TypeDescriptor.GetProperties(value.GetType()))
+                expando.Add(property.Name, property.GetValue(value));
```
thecodejunkie: Missing `{ }`

src/Nancy/Extensions/ObjectExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <summary>
+        /// Convert an object to a dynamic type
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public static dynamic ToDynamic(this object value)
+        {
+            IDictionary<string, object> expando = new ExpandoObject();
+
+            foreach (PropertyDescriptor property in TypeDescriptor.GetProperties(value.GetType()))
```
thecodejunkie: Use `var`

src/Nancy/Extensions/ObjectExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+    public static class ObjectExtensions
+    {
+        /// <summary>
+        /// Convert an object to a dynamic type
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public static dynamic ToDynamic(this object value)
+        {
+            IDictionary<string, object> expando = new ExpandoObject();
```
thecodejunkie: Use `var`

src/Nancy/Extensions/ObjectExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+
+            foreach (PropertyDescriptor property in TypeDescriptor.GetProperties(value.GetType()))
+                expando.Add(property.Name, property.GetValue(value));
+
+            return (ExpandoObject)expando;
```
thecodejunkie: The cast might not be needed after changing to `var` above

src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -16,19 +33,33 @@ public string Serialize(object sourceObject)
...
+            else
+            {
+                var sourceType = this.assemblyCatalog.GetAssemblies().Select(assembly => assembly.GetType(sourceObject.GetType().FullName)).FirstOrDefault(type => type != null);
+                if (sourceType == null)
+                {
+                    throw new DllNotFoundException("Unable to find type " + sourceObject.GetType() + " in its assembly to serialize");
+                }
 
-                return Convert.ToBase64String(outputBytes, 0, (int)outputStream.Length);
+                var aqn = sourceType.GetTypeInfo().AssemblyQualifiedName;
```
thecodejunkie: Not a fan of `aqn`, variable names are free so no need to make them cryptic :smile:

src/Nancy/Extensions/ObjectExtensions.cs
phillip-haydon: Can't be var, has to be dynamic or explicit or he can't modify it. The cast on the return of dynamic assignment wouldn't be needed.

src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -16,19 +33,33 @@ public string Serialize(object sourceObject)
...
-
-                var outputBytes = outputStream.GetBuffer();
+                serializedObject = sourceObject;
+            }
+            else
+            {
+                var sourceType = this.assemblyCatalog.GetAssemblies().Select(assembly => assembly.GetType(sourceObject.GetType().FullName)).FirstOrDefault(type => type != null);
+                if (sourceType == null)
+                {
+                    throw new DllNotFoundException("Unable to find type " + sourceObject.GetType() + " in its assembly to serialize");
```
thecodejunkie: Not sure this exception is the best one. From a user point of view you want to know that it was trying to deserialize something (session state?) and was unable to find the type of the session state data. The exception here could be fine if it is caught higher up and made more user-friendly

src/Nancy/Session/CookieBasedSessions.cs
thecodejunkie: This is going to murder performance since it needs to scan the appdomain each time

src/Nancy/Session/CookieBasedSessions.cs
```diff
@@ -109,7 +109,7 @@ public static IObjectSerializerSelector Enable(IPipelines pipelines)
         {
             return Enable(pipelines, new CookieBasedSessionsConfiguration
             {
-                Serializer = new DefaultObjectSerializer()
+                Serializer = new DefaultObjectSerializer(new AppDomainAssemblyCatalog())
```
thecodejunkie: Same here

src/Nancy/Extensions/ObjectExtensions.cs
thecodejunkie: Good point

thecodejunkie:
So the one thing here is that this does not seem to handle object graphs. The `ObjectExtensions.ToDynamic` does not seem to handle if there are complex types on one of the properties? With the `BinaryFormatter` is not a problem, but here we only seem to handle 1-level of properties and they have to be primitive types

test/Nancy.Tests/Unit/Security/CsrfFixture.cs
```diff
@@ -34,8 +35,10 @@ public CsrfFixture()
             this.pipelines = new MockPipelines();
 
             this.cryptographyConfiguration = CryptographyConfiguration.Default;
-
-            this.objectSerializer = new DefaultObjectSerializer();
+            var fakeAssCatalog = A.Fake<IAssemblyCatalog>();
```
phillip-haydon: People might get offended by the word 'Ass'

thecodejunkie: :smile: This is the same as the `adq` variable, why shorten it? `fakeAssemblyCatalog` is waaaaay more descriptive. :+1: for changing this, even if it's not for the same reason @phillip-haydon pointed out ;)

src/Nancy/Extensions/ObjectExtensions.cs
jchannon: can't

src/Nancy/Extensions/ObjectExtensions.cs
jchannon: can't

jchannon:
@thecodejunkie seems to be ok or are you thinking more complex https://dotnetfiddle.net/0RDcN1

thecodejunkie:
I stand corrected :D

thecodejunkie:
Looking at AppVeyor, we should get #2274 in first and then rebase this?

samples/Nancy.Demo.Hosting.Aspnet/DemoBootstrapper.cs
```diff
@@ -66,7 +66,7 @@ protected override void ApplicationStartup(TinyIoCContainer container, IPipeline
 
             this.Conventions.StaticContentsConventions.Add(StaticContentConventionBuilder.AddDirectory("moo", "Content"));
 
-            CookieBasedSessions.Enable(pipelines);
+            CookieBasedSessions.Enable(pipelines, container.Resolve<IAssemblyCatalog>());
```
thecodejunkie: `this.AssemblyCatalog` should be all that is needed here, no?

src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -17,18 +34,33 @@ public string Serialize(object sourceObject)
...
             {
-                return String.Empty;
+                return string.Empty;
             }
 
-            var formatter = new BinaryFormatter();
+            dynamic serializedObject = null;
 
-            using (var outputStream = new MemoryStream())
+            //In the case of adding values to cookies it is a string else an object for sessions
```
thecodejunkie: I think what we should do here is

1. Kill the comment
2. Break out the `else`-block into a method
3. Change it to

```
serializedObject = (sourceObject is string)
   ? sourceObject
   : this.WhatEverTheMethodIsCalled(....)
```

src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -46,13 +78,15 @@ public object Deserialize(string sourceString)
             try
             {
                 var inputBytes = Convert.FromBase64String(sourceString);
-
-                var formatter = new BinaryFormatter();
-
-                using (var inputStream = new MemoryStream(inputBytes, false))
+                var json = Encoding.UTF8.GetString(inputBytes);
+                //When deserializing objects it will have a TypeObject property
```
thecodejunkie: Kill the comment

src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -46,13 +78,15 @@ public object Deserialize(string sourceString)
             try
             {
                 var inputBytes = Convert.FromBase64String(sourceString);
-
-                var formatter = new BinaryFormatter();
-
-                using (var inputStream = new MemoryStream(inputBytes, false))
+                var json = Encoding.UTF8.GetString(inputBytes);
+                //When deserializing objects it will have a TypeObject property
+                if (json.Contains("TypeObject"))
```
thecodejunkie: Move this to a helper method with a descriptive name, something like "ContainsTypeDecription(json)` or similar

src/Nancy/Extensions/ObjectExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Contains extensions to <see cref="object"/> class.
+    /// </summary>
+    public static class ObjectExtensions
+    {
+        /// <summary>
+        /// Convert an object to a dynamic type
+        /// </summary>
+        /// <param name="value"></param>
```
thecodejunkie: You almost committed a new class without any review comments... almost ;)

src/Nancy/Extensions/ObjectExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <returns></returns>
```
thecodejunkie: This one as well :D

src/Nancy/project.json
```diff
@@ -1,8 +1,8 @@
-{
+﻿{
     "version": "2.0.0-*",
 
     "description": "Nancy is a lightweight web framework for the .Net platform, inspired by Sinatra. Nancy aim at delivering a low ceremony approach to building light, fast web applications.",
-    "authors": [ "Andreas Håkansson, Steven Robbins and contributors" ],
+    "authors": [ "Andreas H√•kansson, Steven Robbins and contributors" ],
```
thecodejunkie: umm :)

src/Nancy/Extensions/ObjectExtensions.cs
jchannon: huh?

src/Nancy/DefaultObjectSerializer.cs
```diff
@@ -17,18 +34,35 @@ public string Serialize(object sourceObject)
...
+                : this.AddTypeInformation(sourceObject);
+
+            var json = SimpleJson.SerializeObject(serializedObject);
+            return Convert.ToBase64String(Encoding.UTF8.GetBytes(json));
+        }
 
-            using (var outputStream = new MemoryStream())
+        private dynamic AddTypeInformation(object sourceObject)
+        {
+            var sourceType = this.assemblyCatalog.GetAssemblies().Select(assembly => assembly.GetType(sourceObject.GetType().FullName)).FirstOrDefault(type => type != null);
```
thecodejunkie: @jchannon this can be replaced by `sourceObject.GetType().AssemblyQualifiedName` can't it?

jchannon: we already do that a few lines below, this is finding the type in the list of assemblies to get the assembly qualified name as the object being serialized might be in a project just with pocos for example

khellang: > this is finding the type in the list of assemblies to get the assembly qualified name

But why do you need to find it in a list of assemblies when you got the type right there? It's even used to get the `FullName` of the type...

jchannon: I think maybe AssemblyQualifiedName is empty if its a class in a separate poco lib or something.

khellang: You *think*? :stuck_out_tongue_closed_eyes:

jchannon: I think thats some sarcasm I'm not getting :smile: We need to find the type in the assemblies to get the AQN

khellang: You can't *think*, you need to *know*.

The only way `AssemblyQualifiedName` returns `null` is if it's a generic type parameter. From MSDN:

> The assembly-qualified name of the Type, which includes the name of the assembly from which the Type was loaded, or null if the current instance represents a generic type parameter.

image

jchannon: Well if you hadn't insisted on one commit message per PR then I might have
something that jogs my memory why I added it.

Remove it, run the tests and if all is ok then consider me told

PR #2269, opened 2016-02-09
Replaced SchemeDelimiter, UriSchemeHttp, UriSchemeHttps

jchannon:
Replaced SchemeDelimiter, UriSchemeHttp, UriSchemeHttps with string literals as suggested by portability analyser


khellang:
7 failing tests... :watch:

PR #2272, opened 2016-02-10
Converted Nancy.Tests to project.json

thecodejunkie:
Converted `Nancy.Tests` to `project.json`

- Includes a couple of `ifdef` to work around the fact that `Assembly.Location` will always be null on `dnx`
- Moved out the `InternalsVisibleTo` declarations, from `SharedAssemblyInfo.cs` and added them to `InternalsVisibleTo.cs` files in the projects that needed them.


.gitignore
```diff
@@ -1,4 +1,4 @@
-*.[Cc]ache
+m*.[Cc]ache
```
khellang: ??

thecodejunkie: wow that must've been me typing while having the wrong window in focus :smile:

test/Nancy.Tests/Unit/ResponseExtensionsFixture.cs
```diff
@@ -177,5 +174,23 @@ public void Should_set_status_code_int()
...
+            return Path.GetDirectoryName(this.GetType().Assembly.Location);
+#endif
+        }
+
+        private string GetFilePath()
+        {
+#if DNX
+            return Path.Combine("Resources", "test.txt");
+#else
+            return Path.GetFileName(this.GetType().Assembly.Location);
```
khellang: Will this yield the same as above (in `DNX`)?

khellang:
LGTM! Nice work @thecodejunkie! :heart:

test/Nancy.Tests/Unit/ResponseExtensionsFixture.cs
thecodejunkie: We can't check the same file in `DNX`. On `DNX` we look for a file that's in `Nancy.Tests/Resources` and for non-`DNX` we just use the assembly name `Nancy.Tests.dll` because we end up in a temp folder

Yantrio:
LGTM :shipit:

jchannon:
Seg fault on Travis?

PR #2274, opened 2016-02-10
Getting reflection working on dotnet54 and net451

Yantrio:
part of #2220

### DIFF
--- a/src/Nancy/Bootstrapper/FavIconApplicationStartup.cs
+++ b/src/Nancy/Bootstrapper/FavIconApplicationStartup.cs
@@ -6,6 +6,7 @@
     using System.Linq;
     using Nancy.Configuration;
     using System.Reflection;
+    
     /// <summary>
     /// Application startup task that attempts to locate a favicon. The startup will first scan all
     /// folders in the path defined by the provided <see cref="IRootPathProvider"/> and if it cannot
--- a/src/Nancy/Bootstrapper/FavIconApplicationStartup.cs
+++ b/src/Nancy/Bootstrapper/FavIconApplicationStartup.cs
@@ -5,7 +5,7 @@
     using System.IO;
     using System.Linq;
     using Nancy.Configuration;
-
+    using System.Reflection;
     /// <summary>
     /// Application startup task that attempts to locate a favicon. The startup will first scan all
     /// folders in the path defined by the provided <see cref="IRootPathProvider"/> and if it cannot
@@ -49,7 +49,8 @@
         private static byte[] ExtractDefaultIcon()
         {
             var resourceStream =
-                typeof(INancyEngine).Assembly.GetManifestResourceStream("Nancy.favicon.ico");
+                typeof(INancyEngine).GetTypeInfo().Assembly.GetManifestResourceStream("Nancy.favicon.ico");
+
 
             if (resourceStream == null)
             {
--- a/src/Nancy/DefaultNancyBootstrapper.cs
+++ b/src/Nancy/DefaultNancyBootstrapper.cs
@@ -277,9 +277,16 @@ namespace Nancy
         /// <param name="container">Container instance</param>
         private static void AutoRegister(TinyIoCContainer container, IEnumerable<Func<Assembly, bool>> ignoredAssemblies)
         {
-            var assembly = typeof(NancyEngine).Assembly;
+            var assembly = typeof(NancyEngine).GetTypeInfo().Assembly;
 
+#if !DNX
             container.AutoRegister(AppDomain.CurrentDomain.GetAssemblies().Where(a => !ignoredAssemblies.Any(ia => ia(a))), DuplicateImplementationActions.RegisterMultiple, t => t.Assembly != assembly);
+#else
+            var libraryManager =
+                Microsoft.Extensions.PlatformAbstractions.PlatformServices.Default.LibraryManager;
+            var thisAssemblyName = typeof(DefaultNancyBootstrapper).GetTypeInfo().Assembly.GetName().Name;
+            var referencing = libraryManager.GetReferencingLibraries(thisAssemblyName);
+#endif
         }
     }
 }
--- a/src/Nancy/DefaultNancyBootstrapper.cs
+++ b/src/Nancy/DefaultNancyBootstrapper.cs
@@ -8,7 +8,7 @@ namespace Nancy
     using Nancy.Configuration;
     using Nancy.Diagnostics;
     using Nancy.TinyIoc;
-
+    using Extensions;
     /// <summary>
     /// TinyIoC bootstrapper - registers default route resolver and registers itself as
     /// INancyModuleCatalog for resolving modules but behavior can be overridden if required.
@@ -275,18 +275,11 @@ namespace Nancy
         /// Executes auto registration with the given container.
         /// </summary>
         /// <param name="container">Container instance</param>
-        private static void AutoRegister(TinyIoCContainer container, IEnumerable<Func<Assembly, bool>> ignoredAssemblies)
+        private void AutoRegister(TinyIoCContainer container, IEnumerable<Func<Assembly, bool>> ignoredAssemblies)
         {
             var assembly = typeof(NancyEngine).GetTypeInfo().Assembly;
 
-#if !DNX
-            container.AutoRegister(AppDomain.CurrentDomain.GetAssemblies().Where(a => !ignoredAssemblies.Any(ia => ia(a))), DuplicateImplementationActions.RegisterMultiple, t => t.Assembly != assembly);
-#else
-            var libraryManager =
-                Microsoft.Extensions.PlatformAbstractions.PlatformServices.Default.LibraryManager;
-            var thisAssemblyName = typeof(DefaultNancyBootstrapper).GetTypeInfo().Assembly.GetName().Name;
-            var referencing = libraryManager.GetReferencingLibraries(thisAssemblyName);
-#endif
+            container.AutoRegister(this.AssemblyCatalog.GetAssemblies().Where(a => !ignoredAssemblies.Any(ia => ia(a))), DuplicateImplementationActions.RegisterMultiple, t => t.GetAssembly() != assembly);
         }
     }
 }
--- a/src/Nancy/DefaultRuntimeEnvironmentInformation.cs
+++ b/src/Nancy/DefaultRuntimeEnvironmentInformation.cs
@@ -3,6 +3,7 @@ namespace Nancy
     using System;
     using System.Diagnostics;
     using System.Linq;
+    using System.Reflection;
 
     /// <summary>
     /// Default implementation of the <see cref="IRuntimeEnvironmentInformation"/> interface.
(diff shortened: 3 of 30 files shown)


thecodejunkie:
I was thinking about this before. Most of these seems to be related to `Assembly` vs. `Assembly()` so perhaps it would be better to get our own extension method, something like `GetAssembly()` and hide all that `ifdef` logic in a single place

Same applies to any other such common pattern

Yantrio:
That sounds like a good idea to me, I should've thought of that but im a bit too tired. I'll get it done tomorrow. 

I'll wait for the :watch: from @khellang

khellang:
What @thecodejunkie said. Isn't `Assembly()` already an extension method?

:watch:

Yantrio:
Ah yeah, looks like Assembly() is an extension method already, I'll start work on moving to using that.

khellang:
The name `Assembly()` is pretty bad, IMO. Just sayin' :wink:

Yantrio:
it's also inside TinyIoc so im going to add another method to TypeExtensions `GetAssembly(this Type t)` like @thecodejunkie suggested.

Yantrio:
or another option is to just use `type.GetTypeInfo().Assembly` everywhere instead.

Yantrio:
no longer [WIP] , feel free to review :D

khellang:
History is borked, as usual :wink:

:watch:

Yantrio:
fixed.

src/Nancy/Bootstrapper/FavIconApplicationStartup.cs
```diff
@@ -5,7 +5,7 @@
     using System.IO;
     using System.Linq;
     using Nancy.Configuration;
-
+    using System.Reflection;
     /// <summary>
```
thecodejunkie: Need a blank line between the `using`-statements and the XML comment

Yantrio: fixed :)

src/Nancy/DefaultNancyBootstrapper.cs
```diff
@@ -277,9 +277,16 @@ protected override sealed INancyModule GetModule(TinyIoCContainer container, Typ
...
-            var assembly = typeof(NancyEngine).Assembly;
+            var assembly = typeof(NancyEngine).GetTypeInfo().Assembly;
 
+#if !DNX
             container.AutoRegister(AppDomain.CurrentDomain.GetAssemblies().Where(a => !ignoredAssemblies.Any(ia => ia(a))), DuplicateImplementationActions.RegisterMultiple, t => t.Assembly != assembly);
+#else
+            var libraryManager =
+                Microsoft.Extensions.PlatformAbstractions.PlatformServices.Default.LibraryManager;
+            var thisAssemblyName = typeof(DefaultNancyBootstrapper).GetTypeInfo().Assembly.GetName().Name;
+            var referencing = libraryManager.GetReferencingLibraries(thisAssemblyName);
```
thecodejunkie: This does not seem to be used? I.e it's not feature equivalent of what is in the `#if !DNX`-section.

All of this should probably just change from using the `AppDomain` to using the `IAssemblyCatalog` which is now on the `NancyBootstrapperBase` so you should just be able to use `this.AssemblyCatalog.GetAssemblies().Where(....)`

src/Nancy/DefaultRuntimeEnvironmentInformation.cs
```diff
@@ -35,10 +36,20 @@ private static bool GetDebugMode(ITypeCatalog typeCatalog)
...
+#if DOTNET5_4
+                    .Select(x => x.GetTypeInfo().Assembly.GetCustomAttributes(typeof(DebuggableAttribute)))
+                    .Where(x => x.Any());
+
+                    return assembliesInDebug.Any(d => ((DebuggableAttribute)d.First()).IsJITTrackingEnabled);
+#else
                     .Select(x => x.Assembly.GetCustomAttributes(typeof(DebuggableAttribute), true))
                     .Where(x => x.Length != 0);
+                    
+                    return assembliesInDebug.Any(d => ((DebuggableAttribute)d[0]).IsJITTrackingEnabled);
```
thecodejunkie: Has the return of `GetCustomAttributes` changed from? Looks like this could be made much more cross-platform if it's only the `GetTypeInfo()` that is required on `DOTNET5_4` since it works on both platforms?

src/Nancy/DefaultNancyBootstrapper.cs
Yantrio: this method is static so it doesn't have access to `this.AssemblyCatalog`

src/Nancy/Diagnostics/InteractiveDiagnostics.cs
```diff
@@ -175,7 +175,11 @@ private static string GetDescriptionFromProperty(IDiagnosticsProvider diagnostic
 
         private static string GetDescriptionFromAttribute(IDiagnosticsProvider diagnosticsProvider, MethodInfo methodInfo)
         {
+#if DOTNET5_4
+            var attribute = (DescriptionAttribute)methodInfo.GetCustomAttribute(typeof(DescriptionAttribute));
+#else
             var attribute = (DescriptionAttribute)Attribute.GetCustomAttribute(methodInfo, typeof(DescriptionAttribute));
```
thecodejunkie: Can't this also be unified? There is a `GetCustomAttribute` method on `MethodInfo` on full-framework, if I am not mistaken?

src/Nancy/Extensions/TypeExtensions.cs
```diff
@@ -3,12 +3,23 @@
...
     /// </summary>
     public static class TypeExtensions
     {
+
+        /// <summary>
+        /// returns the assembly that the type belongs to
+        /// </summary>
+        /// <param name="source"></param>
+        /// <returns> The assembly that contains the type </returns>
+        public static Assembly Assembly(this Type source)
```
thecodejunkie: Perhaps `GetAssembly()` would be a more appropriate name?

src/Nancy/Helpers/ReflectionUtils.cs
```diff
@@ -164,19 +175,19 @@ public static Type GetMemberUnderlyingType(MemberInfo member)
...
-
-            PropertyInfo propertyInfo = member as PropertyInfo;
-
-            if (propertyInfo != null)
-                return IsIndexedProperty(propertyInfo);
-            else
-                return false;
-        }
-
+        public static bool IsIndexedProperty(MemberInfo member)
```
thecodejunkie: This method needs curly-braces around single-line if/else blocks

src/Nancy/Helpers/ReflectionUtils.cs
```diff
@@ -164,19 +175,19 @@ public static Type GetMemberUnderlyingType(MemberInfo member)
...
+        {
+            if (member == null)
+                throw new ArgumentNullException("member");
+
+            PropertyInfo propertyInfo = member as PropertyInfo;
```
thecodejunkie: Use `var` instead of explicit `PropertyInfo`

src/Nancy/Helpers/ReflectionUtils.cs
```diff
@@ -184,59 +195,59 @@ public static bool IsIndexedProperty(MemberInfo member)
...
-            if (property == null)
-                throw new ArgumentNullException("property");
-
-            return (property.GetIndexParameters().Length > 0);
-        }
-
+        public static bool IsIndexedProperty(PropertyInfo property)
+        {
+            if (property == null)
+                throw new ArgumentNullException("property");
```
thecodejunkie: Missing curly-braces

src/Nancy/DefaultRuntimeEnvironmentInformation.cs
Yantrio: I had an issue here where the method sig for GetCustomAttributes had changed, we can no longer check for inherited attributes by passing true as the second arg,

This section of code also worried me because there is no way to get the `IsJITTrackingEnabled` yet, and I fear it may need to be scrapped :(.

src/Nancy/ResponseExtensions.cs
```diff
@@ -182,7 +182,8 @@ public static Response WithStatusCode(this Response response, int statusCode)
         private static Tuple<string, string> GetTuple(object header)
```
thecodejunkie: Unrelated to the PR, but brownie points if you get rid of the weird indentations in this method :heart_eyes:

src/Nancy/Diagnostics/InteractiveDiagnostics.cs
Yantrio: yes this can be unified. doing it now :)

src/Nancy/project.json
```diff
@@ -14,7 +14,6 @@
     "dependencies": {
         "Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc1-final"
     },
-
```
thecodejunkie: Y u kill this? :smile:

src/Nancy/DefaultNancyBootstrapper.cs
thecodejunkie: It's a private method, we can un-static it if needed

src/Nancy/Helpers/ReflectionUtils.cs
Yantrio: ill apply this to the whole file.

src/Nancy/DefaultRuntimeEnvironmentInformation.cs
thecodejunkie: Not sure if this is the correct way, but using `PlatformServices.Default.Application.Configuration` we can get a string telling us the active configuration at runtime, which by default will be `Debug` when you run a debug build. Perhaps these is a more "solid" approach which would let us know if you run a custom configuration in debug-mode though?

src/Nancy/project.json
Yantrio: added it back, atom beautify keeps removing it, sorry. (turned it off for now)

src/Nancy/ResponseExtensions.cs
Yantrio: done :D

src/Nancy/Helpers/ReflectionUtils.cs
Yantrio: fixed.

src/Nancy/Helpers/ReflectionUtils.cs
Yantrio: done.

src/Nancy/Extensions/TypeExtensions.cs
Yantrio: renamed

src/Nancy/DefaultNancyBootstrapper.cs
Yantrio: I've now made this unstatic and implemented it using this.AssemblyCatalog.GetAssemblies(), seems to work nicely, will push soon.

src/Nancy/Helpers/ReflectionUtils.cs
thecodejunkie: Whops, failed to notice the file.. Yeah leave it as is then

thecodejunkie: Oh you changed it already :P

Yantrio: I can change it back if you want, it's up to you :)

thecodejunkie: Na let's leave it =)

PR #2277, opened 2016-02-12
Coreclr convert more test projects

thecodejunkie:
Converted all but 3 test projects to project.json + made some code changes to make tests pass on both environments

The projects that have not yet been converted are

- Nancy.ViewEngines.Spark
- Nancy.ViewEngines.Razor
- Nancy.Tests.Functional

All of these are related to the fact that assemblies will not only exist in memory so they cannot be loaded from the file system, by the engines, to be used to resolve types during view compilation.

- Spark - I've talked to Robert Greyling about this and Spark does not currently support that you pass it `Assembly` instances that can be used during compilation. He's asked us to raise an issue on the Spark repo

- Razor - This is similar to Spark. In Razor we use `CSharpCodeProvider` and it can only take a list of assembly locations that will be loaded, from disk, by the compiler at runtime. Again with in-memory assemblies there is not much we can do except move away from `CSharpCodeProvider`

The reason that `Nancy.Tests.Functional` has not been converted was there there are a bunch of fixtures that rely on Razor and hence they fail and will continue to do so until we've resolved the problems with the Razor engine.


src/Nancy.Hosting.Self/FileSystemRootPathProvider.cs
```diff
@@ -1,17 +1,37 @@
...
+        }
+
+        private static string ExtractRootPath()
+        {
+            var assembly =
+                Assembly.GetEntryAssembly() ?? Assembly.GetExecutingAssembly();
+
+            var location = string.Empty;
+
+#if DNX
```
khellang: I do worry about all of these...

src/Nancy/DefaultRootPathProvider.cs
```diff
@@ -13,7 +13,11 @@ public class DefaultRootPathProvider : IRootPathProvider
         /// <returns>A <see cref="string"/> containing the path of the root folder.</returns>
         public string GetRootPath()
         {
+#if DNX
+            return Microsoft.Extensions.PlatformAbstractions.PlatformServices.Default.Application.ApplicationBasePath;
```
khellang: Law of Demeter, what? :scream:

PR #2286, opened 2016-02-19
Await call to Process in NancyHost

jchannon:
pinging @damianh :smile:


damianh:
No, don't do this, it will serialize (sequential-ize) handling of requests.

jchannon:
:confounded:

grumpydev:
The way it is currently setup though you have an unobserved task which may randomly blow up the appdomain?

damianh:
@grumpydev not on .net 4.5 unless you...

```
<configuration> 
    <runtime> 
        <ThrowUnobservedTaskExceptions enabled="true"/> 
    </runtime> 
</configuration>
```

There is also some good exception handling coverage. https://github.com/NancyFx/Nancy/blob/master/src/Nancy.Hosting.Self/NancyHost.cs#L412-L430

Though this https://github.com/NancyFx/Nancy/blob/master/src/Nancy.Hosting.Self/NancyHost.cs#L429 could be wrapped in a try-catch as it reaches out to external code.

grumpydev:
Still feels wrong to me having a method marked as async and returning a Task but ignoring it :/

damianh:
Yeah a `Task.Run(() => Process(context));` would be more aesthetic.

PR #2290, opened 2016-02-19
Don't use Thread.CurrentThread.CurrentCulture for coreclr

jchannon:
Thread.CurrentThread.CurrentCulture does not exist for CoreCLR. 

Looking at how Microsoft [do it](https://github.com/aspnet/Localization/blob/dev/src/Microsoft.AspNetCore.Localization/RequestLocalizationMiddleware.cs#L129-L135) I've made the containing changes.

### DIFF
--- a/src/Nancy/Configuration/ConfigurationException.cs
+++ b/src/Nancy/Configuration/ConfigurationException.cs
@@ -1,0 +1,20 @@
+namespace Nancy
+{
+    using System;
+    using Nancy.Configuration;
+
+    /// <summary>
+    /// An exception related to an invalid configuration created within <see cref="INancyEnvironment"/>
+    /// </summary>
+    public class ConfigurationException : Exception
+    {
+        /// <summary>
+        /// Create an instance of <see cref="ConfigurationException"/>
+        /// </summary>
+        /// <param name="message">A message to be passed into the exception</param>
+        public ConfigurationException(string message) : base(message)
+        {
+            
+        }
+    }
+}
--- a/src/Nancy/Conventions/BuiltInCultureConventions.cs
+++ b/src/Nancy/Conventions/BuiltInCultureConventions.cs
@@ -133,7 +133,12 @@
         /// <returns>CultureInfo from CurrentThread</returns>
         public static CultureInfo ThreadCulture(NancyContext context, GlobalizationConfiguration configuration)
         {
+#if DOTNET5_4
+            CultureInfo.CurrentCulture = new CultureInfo(configuration.SupportedCultureNames.First());
+            return CultureInfo.CurrentCulture;
+#else
             return Thread.CurrentThread.CurrentCulture;
+#endif
         }
 
         /// <summary>
--- a/src/Nancy/Conventions/BuiltInCultureConventions.cs
+++ b/src/Nancy/Conventions/BuiltInCultureConventions.cs
@@ -130,7 +130,7 @@
         /// <param name="context">NancyContext</param>
         /// <param name="configuration">Culture configuration that contains allowed cultures</param>
         /// <returns></returns>
-        public static CultureInfo ConfigurationCulture(NancyContext context, GlobalizationConfiguration configuration)
+        public static CultureInfo GlobalizationConfigurationCulture(NancyContext context, GlobalizationConfiguration configuration)
         {
             if (configuration.DefaultCulture != null)
             {
--- a/src/Nancy/Conventions/BuiltInCultureConventions.cs
+++ b/src/Nancy/Conventions/BuiltInCultureConventions.cs
@@ -124,6 +124,26 @@
 
             return null;
         }
+        /// <summary>
+        /// Checks to see if a default culture has been set on <see cref="GlobalizationConfiguration"/>
+        /// </summary>
+        /// <param name="context">NancyContext</param>
+        /// <param name="configuration">Culture configuration that contains allowed cultures</param>
+        /// <returns></returns>
+        public static CultureInfo ConfigurationCulture(NancyContext context, GlobalizationConfiguration configuration)
+        {
+            if (configuration.DefaultCulture != null)
+            {
+                if (!IsValidCultureInfoName(configuration.DefaultCulture, configuration))
+                {
+                    return null;
+                }
+
+                return new CultureInfo(configuration.DefaultCulture);
+            }
+
+            return null;
+        }
 
         /// <summary>
         /// Uses the Thread.CurrentThread.CurrentCulture
@@ -134,7 +154,6 @@
         public static CultureInfo ThreadCulture(NancyContext context, GlobalizationConfiguration configuration)
         {
 #if DOTNET5_4
-            CultureInfo.CurrentCulture = new CultureInfo(configuration.SupportedCultureNames.First());
             return CultureInfo.CurrentCulture;
 #else
             return Thread.CurrentThread.CurrentCulture;
--- a/src/Nancy/Conventions/BuiltInCultureConventions.cs
+++ b/src/Nancy/Conventions/BuiltInCultureConventions.cs
@@ -153,11 +153,7 @@
         /// <returns>CultureInfo from CurrentThread</returns>
         public static CultureInfo ThreadCulture(NancyContext context, GlobalizationConfiguration configuration)
         {
-#if DOTNET5_4
             return CultureInfo.CurrentCulture;
-#else
-            return Thread.CurrentThread.CurrentCulture;
-#endif
         }
 
         /// <summary>
--- a/src/Nancy/Conventions/BuiltInCultureConventions.cs
+++ b/src/Nancy/Conventions/BuiltInCultureConventions.cs
@@ -129,7 +129,7 @@
         /// </summary>
         /// <param name="context">NancyContext</param>
         /// <param name="configuration">Culture configuration that contains allowed cultures</param>
-        /// <returns></returns>
+        /// <returns>CultureInfo if found in <see cref="GlobalizationConfiguration"/> Default Culture else null</returns>
         public static CultureInfo GlobalizationConfigurationCulture(NancyContext context, GlobalizationConfiguration configuration)
         {
             if (configuration.DefaultCulture != null)
--- a/src/Nancy/Conventions/DefaultCultureConventions.cs
+++ b/src/Nancy/Conventions/DefaultCultureConventions.cs
@@ -47,7 +47,7 @@
                 BuiltInCultureConventions.HeaderCulture,
                 BuiltInCultureConventions.SessionCulture,
                 BuiltInCultureConventions.CookieCulture,
-                BuiltInCultureConventions.ConfigurationCulture,
+                BuiltInCultureConventions.GlobalizationConfigurationCulture,
                 BuiltInCultureConventions.ThreadCulture
             };
         }
--- a/src/Nancy/Conventions/DefaultCultureConventions.cs
+++ b/src/Nancy/Conventions/DefaultCultureConventions.cs
@@ -4,6 +4,9 @@
     using System.Collections.Generic;
     using System.Globalization;
 
+    /// <summary>
+    /// Default implementation of <see cref="IConvention"/>
+    /// </summary>
     public class DefaultCultureConventions : IConvention
     {
         /// <summary>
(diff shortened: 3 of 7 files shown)


src/Nancy/Conventions/BuiltInCultureConventions.cs
```diff
@@ -133,7 +133,12 @@ public static CultureInfo CookieCulture(NancyContext context, GlobalizationConfi
         /// <returns>CultureInfo from CurrentThread</returns>
         public static CultureInfo ThreadCulture(NancyContext context, GlobalizationConfiguration configuration)
         {
+#if DOTNET5_4
+            CultureInfo.CurrentCulture = new CultureInfo(configuration.SupportedCultureNames.First());
```
thecodejunkie: Is this even the same thing as `Thread.CurrentThread.CurrentCulture` maybe it is the `ThreadCulture` that should not be supported on `dotnet5_4` at all?

jchannon: Here's the docs for it

CurrentCulture

Gets or sets the CultureInfo
<http://dotnet.github.io/api/System.Globalization.CultureInfo.html#System_Globalization_CultureInfo>
object
that represents the culture used by the current thread.


Don't forget to click the link to MS way of doing it

jchannon:
Just ran this in a console app and the results were `en-US, en-GB (my default)`

            CultureInfo.CurrentCulture = new CultureInfo("en-us");
            Console.WriteLine(CultureInfo.CurrentCulture);
            
            var t = Task.Run(async() =>{
                Console.WriteLine(CultureInfo.CurrentCulture);
            });
            
            Task.WaitAll(t);
            Console.ReadKey();

src/Nancy/Conventions/BuiltInCultureConventions.cs
thecodejunkie: What I meant it, is setting the `CultureInfo.CurrentCulture`, to the first `SupportedCultureNames`, and then returning it the same as returning `Thread.CurrentThread.CurrentCulture`?

Before we just returned `Thread.CurrentThread.CurrentCulture` which was set by something else based on some criterion, now we explicitly set it to the first configured one'

I know you can change it using

```xml
<globalization culture="en-US" />
```

in your `web.config`, which would mimic using `SupportedCultureNames.First()`, but if the `globalization` element then did it not pick it up form the OS? That wouldn't be a supported case for us anymore, with this code, and a major breaking change?

jchannon:
Have removed call to  `return Thread.CurrentThread.CurrentCulture;` and replaced with `return CultureInfo.CurrentCulture;` as the latter calls the former under the hood anyway, plus this means no need for an `ifdef` for coreclr

src/Nancy/GlobalizationConfiguration.cs
```diff
@@ -15,15 +15,22 @@ public class GlobalizationConfiguration
...
 
         /// <summary>
         /// A set of supported cultures
         /// </summary>
         public IEnumerable<string> SupportedCultureNames { get; private set; }
+
+        /// <summary>
+        /// The default culture for the application
+        /// </summary>
+        public string DefaultCulture { get; set; }
```
thecodejunkie: Should have a private setter

src/Nancy/GlobalizationConfigurationExtensions.cs
```diff
@@ -13,10 +13,12 @@ public static class GlobalizationConfigurationExtensions
...
         /// <param name="environment">An <see cref="INancyEnvironment"/> that should be configured.</param>
         /// <param name="supportedCultureNames">Cultures that the application can accept</param>
-        public static void Cultures(this INancyEnvironment environment, IEnumerable<string> supportedCultureNames)
+        /// <param name="defaultCulture">Used to set a default culture for the application <remarks>If not specified the first supported culture is used</remarks></param>
+        public static void Cultures(this INancyEnvironment environment, IEnumerable<string> supportedCultureNames, string defaultCulture = null)
         {
             environment.AddValue(new GlobalizationConfiguration(
-                supportedCultureNames: supportedCultureNames));
+                supportedCultureNames: supportedCultureNames,
+                defaultCulture:defaultCulture));
```
thecodejunkie: Missing a space after the `:`

src/Nancy/GlobalizationConfigurationExtensions.cs
```diff
@@ -13,10 +13,12 @@ public static class GlobalizationConfigurationExtensions
         /// </summary>
         /// <param name="environment">An <see cref="INancyEnvironment"/> that should be configured.</param>
         /// <param name="supportedCultureNames">Cultures that the application can accept</param>
-        public static void Cultures(this INancyEnvironment environment, IEnumerable<string> supportedCultureNames)
+        /// <param name="defaultCulture">Used to set a default culture for the application <remarks>If not specified the first supported culture is used</remarks></param>
+        public static void Cultures(this INancyEnvironment environment, IEnumerable<string> supportedCultureNames, string defaultCulture = null)
         {
             environment.AddValue(new GlobalizationConfiguration(
```
thecodejunkie: If no `defaultCulture` is provided, then we should _attempt_ to use `supportedCultureNames.First()`. Emphasis on _attempt_ because we still need to be sure we don't get a `NullRefererenceException` and so on. 

1. There is a remark that seem to suggest this is the case but I can't see the code that actually performs this task
2. The remark should not be inside the `<param name=..` but rather on the line below

src/Nancy/Conventions/BuiltInCultureConventions.cs
```diff
@@ -124,6 +124,26 @@ public static CultureInfo CookieCulture(NancyContext context, GlobalizationConfi
 
             return null;
         }
+        /// <summary>
+        /// Checks to see if a default culture has been set on <see cref="GlobalizationConfiguration"/>
+        /// </summary>
+        /// <param name="context">NancyContext</param>
+        /// <param name="configuration">Culture configuration that contains allowed cultures</param>
+        /// <returns></returns>
+        public static CultureInfo ConfigurationCulture(NancyContext context, GlobalizationConfiguration configuration)
```
thecodejunkie: Maybe call this `GlobalizationCulture` or `GlobalizationConfigurationCulture` so it is more descriptive of what it does

src/Nancy/GlobalizationConfiguration.cs
```diff
@@ -15,15 +15,22 @@ public class GlobalizationConfiguration
         /// <summary>
         /// Initializes a new instance of the <see cref="GlobalizationConfiguration"/> class
         /// </summary>
-        /// <param name="supportedCultureNames"></param>
-        public GlobalizationConfiguration(IEnumerable<string> supportedCultureNames)
+        /// <param name="supportedCultureNames">An array of supported cultures</param>
+        /// <param name="defaultCulture">The default culture of the application</param>
```
thecodejunkie: I think we need to point out that the culture you provide here, needs to exist in the `supportedCultureNames` collection and handle it appropriately when it does not. Maybe we need to introduce a general purpose `ConfigurationException` for these types of things?

src/Nancy/GlobalizationConfigurationExtensions.cs
jchannon: Actually not sure we need to use `First` as it will fall back to the Thread convention

thecodejunkie: Again this can't be the case. You cannot have a supported list of cultures be `en-US` and `sv-SE` and have it fallback to `en-GB` because that happens to be the OS culture. In the context of your application, one of the configured cultures _has_ to be able to be treated as the _default_ one.

The `ThreadCulture` should be the absolute last fallback and I can only ever see that happen if you configure `env.Cultures(null)` or `env.Cultures(Enumerable.Empty<string>())`

src/Nancy/GlobalizationConfiguration.cs
jchannon: Currently if the list is not populated but the default is or the default isn't in the list, the convention  will fail and return null and therefore will fallback to the thread convention. We can add checks in this class and throw an exception if you want though

thecodejunkie: My thoughts were that his is just not for the culture conventions, but the configuration in general. If you say you support a + b and then say c should be the default, that's an invalid configuration if you ask me. This means that you can't reliably check `DefaultCulture` from anywhere without also validating it against `SupportedCultureNames`.

src/Nancy/GlobalizationConfigurationExtensions.cs
```diff
@@ -13,10 +14,18 @@ public static class GlobalizationConfigurationExtensions
...
-        public static void Cultures(this INancyEnvironment environment, IEnumerable<string> supportedCultureNames)
+        /// <param name="defaultCulture">Used to set a default culture for the application</param>
+        /// <remarks>If defaultCulture not specified the first supported culture is used</remarks>
+        public static void Cultures(this INancyEnvironment environment, IEnumerable<string> supportedCultureNames, string defaultCulture = null)
         {
+            if (supportedCultureNames != null && supportedCultureNames.Any() && defaultCulture == null)
+            {
+                defaultCulture = supportedCultureNames.FirstOrDefault();
+            }
+
```
thecodejunkie: This is where we should make sure that `defaultCulture` actually is in the list of `supportedCultureNames`

src/Nancy/GlobalizationConfiguration.cs
jchannon: That's exactly what is done to determine the Context.CurrentCulture. You wouldn't be using the configuration default culture anyway I hope, you should be using the Context.CurrentCulture

src/Nancy/GlobalizationConfigurationExtensions.cs
jchannon: If we're going to do this then we'd need the check in the extension class and in the configuration class because the user could use either class

src/Nancy/GlobalizationConfiguration.cs
thecodejunkie: You seem to associate the Culture settings only being used with the Culture Conventions? There is nothing stopping someone from grabbing the `GlobalizationConfiguration` and using `SupportedCultureNames` and/or `DefaultCulture` anywhere in their code? It makes sense to make sure the latter is included in the former.

src/Nancy/Configuration/ConfigurationException.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    /// <summary>
+    /// An exception related to an invalid configuration created within <see cref="INancyEnvironment"/>
+    /// </summary>
+    public class ConfigurationException : Exception
+    {
+        /// <summary>
+        /// Create an instance of <see cref="ConfigurationException"/>
+        /// </summary>
+        /// <param name="message"></param>
```
thecodejunkie: image

jchannon: Want the configuration object passed?

thecodejunkie: I want the XML comment not to be blank :D

src/Nancy/Conventions/BuiltInCultureConventions.cs
```diff
@@ -124,6 +124,26 @@ public static CultureInfo CookieCulture(NancyContext context, GlobalizationConfi
 
             return null;
         }
+        /// <summary>
+        /// Checks to see if a default culture has been set on <see cref="GlobalizationConfiguration"/>
+        /// </summary>
+        /// <param name="context">NancyContext</param>
+        /// <param name="configuration">Culture configuration that contains allowed cultures</param>
+        /// <returns></returns>
```
thecodejunkie: image

src/Nancy/Conventions/DefaultCultureConventions.cs
```diff
@@ -4,6 +4,9 @@
     using System.Collections.Generic;
     using System.Globalization;
 
+    /// <summary>
+    /// Default implementation of <see cref="IConvention"/>
```
thecodejunkie: Brownie points :)

src/Nancy/Conventions/BuiltInCultureConventions.cs
Yantrio: github isn't a place for jokes @thecodejunkie ..

thecodejunkie:
A rebase and we'll try and get this in tomorrow morning? :heart:

thecodejunkie:
Is this superseded by #2310 ?

jchannon:
Yup

PR #2291, opened 2016-02-19
Implemented RazorAssemblyProvider for the Razor engine

thecodejunkie:
This pull-request is a pre-requisite for an upcoming change to the `IAssemblyCatalog` implementations. They are going to be updated to _only_ contain Nancy & Nancy-referencing assemblies. However the Roslyn engine needs access to a wider selection of assemblies in order to be able to compile views, enter the `RazorAssemblyProvider`. The class is an aggregate of several assembly sources

- `IAssemblyCatalog.GetAssemblies()` which will make sure Razor has access to the Nancy assemblies
- `IRazorConfiguration.GetAssemblyNames()` which will be used to tell the engine about additional assemblies it should be interested in
- A [built in list of assemblies](https://github.com/thecodejunkie/Nancy/blob/b1b4df176a760a0798800676e75c4bf7b0a672f9/src/Nancy.ViewEngines.Razor/RazorAssemblyProvider.cs#L16) that are either essential for Roslyn to function or there to just be nice to include for view authors as default

Most of the pull-request consists of making sure that the `RazorAssemblyProvider` is passed around instead of an `IAssemblyCatalog`. The `RazorAssemblyProvider` is essentially used for two things

- Provide assembly information to Roslyn
- Used by the Razor engine to look up a model type that may exist in a satellite assembly without a Nancy-reference (hence not in the `IAssemblyCatalog`)


src/Nancy.ViewEngines.Razor/RazorAssemblyProvider.cs
```diff
@@ -0,0 +1,94 @@
+namespace Nancy.ViewEngines.Razor
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Reflection;
+
+    /// <summary>
+    /// Provides access to the <see cref="Assembly"/> that are availab to Razor.
```
Yantrio: typo here. should be available

thecodejunkie:
This needs reviewing and pulled asap so I can rebase #2292, which is the final obstacle for us to implement an `IAssemblyCatalog` for `DNX` and remove the remaining compilation errors when targeting `dotnet5_4` from the Nancy project

Yantrio:
LGTM, let's ship it!

src/Nancy.ViewEngines.Razor/RazorAssemblyProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+            return loadedAssemblies;
+        }
+
+        private IEnumerable<Assembly> GetDefaultAssemblies()
+        {
+            var loadedAssemblies = new HashSet<Assembly>();
+
+            foreach (var assemblyDefinition in this.defaultAssemblyDefinitions)
+            {
+                var assembly = AppDomain.CurrentDomain.GetAssemblies().SingleOrDefault(x => x.GetName().Name.Equals(assemblyDefinition));
```
jchannon: need a ordinal.ignorecase comparison here?

thecodejunkie: Technically it shouldn't be needed because they're cased correctly, but it couldn't hurt to safe-guard us :+1:

PR #2292, opened 2016-02-19
Updated AppDomainAssemblyCatalog implementation

thecodejunkie:
- Will now only load Nancy referencing assemblies
- Will use a second application domain to inspect assemblies

Removed the various `AssemblyResolveStrategies` and the `AssemblyCatalogBase` classes

For the time being this is [WIP] and needs to be rebased after #2291 has been merged

Once this is in, I will rebase my `ILibraryManager`-based implementation for core and send a separate pull-request


jchannon:
Plan is to have another implementation of IAssemblyCatalog for CoreCLR?

thecodejunkie:
Yep, that's an incoming pull-request. It has a pre-requisite on this one because this one also removes the `AssemblyResolveStrategy` (from the project) which was used on `IAssenmblyCatalog.GetAssemblies`. I have an `ILibraryManager` based implementation ready, that I will rebase once this is in

src/Nancy/AppDomainAssemblyCatalog.cs
```diff
@@ -10,126 +10,105 @@ namespace Nancy
...
-            this.isAppDomainPrepouplated = true;
+            var assemblies = GetLoadedNancyReferencingAssemblies();
 
-            var assemblies = this.GetAvailableAssemblies();
+            var loaded = LoadNancyReferencingAssemblies(assemblies);
 
-            var assemblyPaths = assemblies
-                .Select(assembly => assembly.Location)
-                .ToArray();
+            return assemblies.Union(loaded).ToArray();
```
jchannon: Any difference in using `Concat` ?

thecodejunkie: `Union` would remove duplicates

PR #2293, opened 2016-02-19
Removed GetRawCertData error by passing X509Certificate2 

jchannon:
Currently we pass a `byte[]` into `Request` to construct a `ClientCertificate`, in ASPNet Host and SelfHost the incoming request already gives us a `X509Certificate2` so we can pass that through.

The OWIN spec says that `ssl.ClientCertificate` is type `X509Certificate` however we can pull it out of the environment dictionary and cast that to a `X509Certificate2`. This is also how MS [does it](https://github.com/aspnet/HttpAbstractions/blob/5e7b30c04b89441df4b60c0bc49a3b05c3a642db/src/Microsoft.AspNetCore.Owin/OwinFeatureCollection.cs#L248) (their `Prop` method is the same as our `Get` method when retrieving items out the dictionary to a `Type`.

### DIFF
--- a/src/Nancy/Owin/NancyMiddleware.cs
+++ b/src/Nancy/Owin/NancyMiddleware.cs
@@ -68,12 +68,12 @@
                         var owinRequestProtocol = Get<string>(environment, "owin.RequestProtocol");
                         var owinCallCancelled = Get<CancellationToken>(environment, "owin.CallCancelled");
                         var owinRequestHost = GetHeader(owinRequestHeaders, "Host") ?? Dns.GetHostName();
-                        var owinUser = GetUser(environment); 
+                        var owinUser = GetUser(environment);
 
                         X509Certificate2 certificate = null;
                         if (options.EnableClientCertificates)
                         {
-                            var clientCertificate = Get<X509Certificate2>(environment, "ssl.ClientCertificate");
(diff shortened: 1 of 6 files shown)


src/Nancy/Owin/NancyMiddleware.cs
```diff
@@ -70,11 +70,11 @@ public static MidFunc UseNancy(NancyOptions options = null)
                         var owinRequestHost = GetHeader(owinRequestHeaders, "Host") ?? Dns.GetHostName();
                         var owinUser = GetUser(environment); 
 
-                        byte[] certificate = null;
+                        X509Certificate2 certificate = null;
                         if (options.EnableClientCertificates)
                         {
-                            var clientCertificate = Get<X509Certificate>(environment, "ssl.ClientCertificate");
-                            certificate = (clientCertificate == null) ? null : clientCertificate.GetRawCertData();
+                            var clientCertificate = Get<X509Certificate2>(environment, "ssl.ClientCertificate");
```
thecodejunkie: This is guaranteed to work because `X509Certificate2` inherits from `X509Certificate` ?

jchannon: Assume so

thecodejunkie: Something a bit more reassuring than assumptions would be nice. Have we verified that the code change works at runtime, i.e do we have any tests that actually executes these code paths?

jchannon: See new test

PR #2296, opened 2016-02-22
Added DNX support in the NancyBootstrapperLocator

thecodejunkie:
The following changes were made

- Now supports DNX through the `ILibraryManager`
- Updated so that only Nancy-referencing assemblies are returned (for both platforms)

Once this pull-request is in, and the one for a DNX compatible `IAssemblyCatalog`, then the `ILibraryManager` code, for getting Nancy-referencing assemblies, can be refactored out to a helper and used in both places. Right now there is a bit of code duplication going on

### DIFF
--- a/src/Nancy/Bootstrapper/NancyBootstrapperLocator.cs
+++ b/src/Nancy/Bootstrapper/NancyBootstrapperLocator.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Reflection;
     using Nancy.Extensions;
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


src/Nancy/Bootstrapper/NancyBootstrapperLocator.cs
```diff
@@ -52,6 +54,64 @@ private static IReadOnlyCollection<Type> GetAvailableBootstrapperTypes()
...
+                if (assembly != null)
+                {
+                    results.Add(assembly);
+                }
+            }
+
+            return results.ToArray();
+#endif
+
+            return AppDomain.CurrentDomain.GetAssemblies().Where(IsNancyReferencing);
```
Yantrio: should this be inside an `#else` instead of just outside the #if?

Yantrio:
LGTM :shipit:

PR #2297, opened 2016-02-23
Fixed NullReferenceException if Start is not called before Dispose.

SonnevilleJ:
Fixed an issue where NancyHost.Dispose() would throw an exception if Start() was not previously called.


thecodejunkie:
Thank you :yum: :

PR #2301, opened 2016-02-23
Updated DiagnosticsConfiguration with Enabled

thecodejunkie:
The following modifications where made

- new member to control if diagnostics should be enabled or not (will defaults to false)
- updated the diagnostics hook to not use `IRuntimeEnvironmentInformation.IsDebug` to determine if it should display a help page or not
- updated help page text as it was no longer relevant since we migrated to the configuration API

It makes sense to have the `Enabled` property on it, because then the user application can set this value depending on things like an environment variable etc.

Most of the changed files are either test fixtures or boostrappers in sample projects. The main file is the `DiagnosticsHook` class

### DIFF
--- a/src/Nancy/Diagnostics/DiagnosticsConfigurationExtensions.cs
+++ b/src/Nancy/Diagnostics/DiagnosticsConfigurationExtensions.cs
@@ -12,6 +12,28 @@
         /// Configures diagnostics.
         /// </summary>
         /// <param name="environment"><see cref="INancyEnvironment"/> that should be configured.</param>
+        /// <param name="password">Password used to secure the dashboard.</param>
+        /// <param name="path">Relative path of the dashboard.</param>
+        /// <param name="cookieName">Name of the cookie to store diagnostics information.</param>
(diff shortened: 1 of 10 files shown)


src/Nancy/Diagnostics/DiagnosticsConfigurationExtensions.cs
```diff
@@ -12,19 +12,21 @@ public static class DiagnosticsConfigurationExtensions
...
-        public static void Diagnostics(this INancyEnvironment environment, string password, string path = null, string cookieName = null, int slidingTimeout = 15, CryptographyConfiguration cryptographyConfiguration = null)
+        public static void Diagnostics(this INancyEnvironment environment, bool enabled, string password, string path = null, string cookieName = null, int slidingTimeout = 15, CryptographyConfiguration cryptographyConfiguration = null)
         {
             environment.AddValue(new DiagnosticsConfiguration(
-                password,
-                path,
-                cookieName,
-                slidingTimeout,
-                cryptographyConfiguration));
+                enabled: enabled,
```
khellang: To much smurf? :stuck_out_tongue_winking_eye:

PR #2307, opened 2016-02-23
Removed use of IAssemblyCatalog in DefaultObjectSerializer

thecodejunkie:
I removed the use of `IAssemblyCatalog` to resolve the `AssemblyQualifiedName` of a `Type`. All tests passed after the change. I also tidied up the code a bit after it was removed.


khellang:
This is really nice :heart: Too bad it wasn't caught in the first review :stuck_out_tongue_closed_eyes:

thecodejunkie:
That's all on me and not @jchannon I lead him down that path thinking it was for another setup :disappointed:

PR #2308, opened 2016-02-23
Removed use of Assembly.EscapedCodeBase

thecodejunkie:
This was used quite a lot, in the Razor engine, before we moved it over to Roslyn. The only assembly we now need to load into the `CSharpCodeProvider` is `Microsoft.CSharp` and since that is in the GAC we can just use `Asembly.Location`

### DIFF
--- a/src/Nancy.ViewEngines.Razor/CSharp/CSharpRazorViewRenderer.cs
+++ b/src/Nancy.ViewEngines.Razor/CSharp/CSharpRazorViewRenderer.cs
@@ -4,10 +4,7 @@
     using System.CodeDom.Compiler;
     using System.Collections.Generic;
     using System.Web.Razor;
-    using Microsoft.CSharp;
     using Microsoft.CSharp.RuntimeBinder;
-
-    using Nancy.Extensions;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Nancy.ViewEngines.Razor/CSharp/CSharpRazorViewRenderer.cs
```diff
@@ -50,7 +47,7 @@ public CSharpRazorViewRenderer(IAssemblyCatalog assemblyCatalog)
         {
             this.Assemblies = new List<string>
             {
-                typeof(Binder).GetAssemblyPath()
+                typeof(Binder).Assembly.Location
```
Yantrio: should use GetAssembly() or GetTypeInfo().Assembly

thecodejunkie: Hmm `GetAssembly` is only available on the `coreclr` branch =/

Yantrio: When we rebase the coreclr branch, we will just deal with it then :)

PR #2310, opened 2016-02-24
Created new culture convention for default culture

jchannon:
This adds a new globalization culture that adds a DefaultCulture property.  If one is not provided the first one in the list of supported cultures is chosen.

If a user passes in an empty list of cultures it will fall back to the ThreadCultureConvention which now uses `CultureInfo.CurrentCulture` instead of `Thread.CurrentThread.CurrentCulture` which makes it CoreCLR compatible

### DIFF
--- a/src/Nancy/GlobalizationConfiguration.cs
+++ b/src/Nancy/GlobalizationConfiguration.cs
@@ -15,15 +15,22 @@
         /// <summary>
         /// Initializes a new instance of the <see cref="GlobalizationConfiguration"/> class
         /// </summary>
-        /// <param name="supportedCultureNames"></param>
-        public GlobalizationConfiguration(IEnumerable<string> supportedCultureNames)
+        /// <param name="supportedCultureNames">An array of supported cultures</param>
+        /// <param name="defaultCulture">The default culture of the application</param>
+        public GlobalizationConfiguration(IEnumerable<string> supportedCultureNames, string defaultCulture = null)
         {
             this.SupportedCultureNames = supportedCultureNames;
+            this.DefaultCulture = defaultCulture;
         }
 
         /// <summary>
         /// A set of supported cultures
         /// </summary>
         public IEnumerable<string> SupportedCultureNames { get; private set; }
+
+        /// <summary>
+        /// The default culture for the application
+        /// </summary>
+        public string DefaultCulture { get; set; }
     }
 }
--- a/src/Nancy/GlobalizationConfiguration.cs
+++ b/src/Nancy/GlobalizationConfiguration.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Globalization;
     using System.Linq;
 
     /// <summary>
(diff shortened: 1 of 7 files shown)


src/Nancy/GlobalizationConfiguration.cs
```diff
@@ -10,20 +12,37 @@ public class GlobalizationConfiguration
...
         /// <summary>
         /// Initializes a new instance of the <see cref="GlobalizationConfiguration"/> class
         /// </summary>
-        /// <param name="supportedCultureNames"></param>
-        public GlobalizationConfiguration(IEnumerable<string> supportedCultureNames)
+        /// <param name="supportedCultureNames">An array of supported cultures</param>
+        /// <param name="defaultCulture">The default culture of the application</param>
+        public GlobalizationConfiguration(IEnumerable<string> supportedCultureNames, string defaultCulture = null)
         {
+            if (supportedCultureNames != null && supportedCultureNames.Any() && defaultCulture == null)
```
thecodejunkie: Can't the call to `Any()` be removed here? Just seem to add overhead since we do `FirstOfDefault` below

src/Nancy/GlobalizationConfiguration.cs
```diff
@@ -10,20 +12,37 @@ public class GlobalizationConfiguration
...
+        public static readonly GlobalizationConfiguration Default = new GlobalizationConfiguration(supportedCultureNames: new[] { "en-US" }, defaultCulture: "en-US");
 
         /// <summary>
         /// Initializes a new instance of the <see cref="GlobalizationConfiguration"/> class
         /// </summary>
-        /// <param name="supportedCultureNames"></param>
-        public GlobalizationConfiguration(IEnumerable<string> supportedCultureNames)
+        /// <param name="supportedCultureNames">An array of supported cultures</param>
+        /// <param name="defaultCulture">The default culture of the application</param>
+        public GlobalizationConfiguration(IEnumerable<string> supportedCultureNames, string defaultCulture = null)
```
thecodejunkie: The more I think about it, shouldn't passing in an empty `supportedCultureNames` be considered an invalid configuration? And if we do that, then the `ThreadCulture` can probably be nuked?

Perhaps the behavior should be that the *default* (that is set by `DefaultGlobalizationConfigurationProvider`) should actually be `CultureInfo.CurrentCulture` so unless you explicitly call `env.Cultures` then we use the OS culture. 

However *if* you actually do call `env.Cultures` then you `explicitly` care about the supported cultures, hence passing in null/empty list should be a `ConfigurationException`

The same behavior for `defaultCulture` should still remain as it is implemented below (optional, ensure it's one of the supported, use the first supported as the fallback default)

Does that not create a much more solid story?

jchannon: Sounds good

thecodejunkie: Sorry for creating additional work for you :) I just think this is one of those things that need a couple of iterations before we distill something really great. I think the last set of edits will put us at the sweet spot

thecodejunkie:
:heart:

PR #2311, opened 2016-02-25
Added kestrel demo project

jchannon:


samples/Nancy.Demo.Hosting.Kestrel/HomeModule.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public HomeModule()
+        {
+            Get["/"] = (o, token) => Task.FromResult<dynamic>("Hello from Nancy running on CoreCLR");
+
+            Get["/conneg/{name}"] = (parameters, token) => Task.FromResult<dynamic>(new Person() { Name = parameters.name });
+        }
+    }
+
+    public class Person
```
Yantrio: just because it's a sample, doesn't mean you should put multiple classes in the same file. tut tut tut.

Yantrio:
LGTM :shipit:

samples/Nancy.Demo.Hosting.Kestrel/HomeModule.cs
phillip-haydon: If the Person is a response object then it should be nested in the HomeModule, for name spacing, in a real world app, because that response object shouldn't be shared among other responses.

PR #2314, opened 2016-02-25
Consolidated the dictionary check on GetMatches (Routing)

phillip-haydon:


jchannon:
Looking on phone but don't we still need the IsNull check?

phillip-haydon:
The check was only checking the existence of the key in the dictionary, using a `TryGetValue` will return false if the key doesn't exist or throws. Currently it checks the dictionary twice, `TryGetValue` will result in a single check. (perf)

jchannon:
LGTM :shipit:

src/Nancy/Routing/Trie/RouteResolverTrie.cs
```diff
@@ -62,19 +62,15 @@ public void BuildTrie(IRouteCache cache)
         /// <returns>An array of <see cref="MatchResult"/> elements</returns>
         public MatchResult[] GetMatches(string method, string path, NancyContext context)
         {
-            if (string.IsNullOrEmpty(path))
```
thecodejunkie: I think @jchannon is correct. Since you remove the null-check on `path` it can explode below on line 72 ?

src/Nancy/Routing/Trie/RouteResolverTrie.cs
```diff
@@ -62,19 +62,15 @@ public void BuildTrie(IRouteCache cache)
...
-            // TODO -concurrent if allowing updates?
-            if (!this.routeTries.ContainsKey(method))
+            if (!this.routeTries.TryGetValue(method, out result))
             {
                 return MatchResult.NoMatches;
             }
 
-            return this.routeTries[method].GetMatches(path.Split(splitSeparators, StringSplitOptions.RemoveEmptyEntries), context)
-                                          .ToArray();
+            return result.GetMatches(path.Split(splitSeparators, StringSplitOptions.RemoveEmptyEntries), context)
```
thecodejunkie: i.e you can still end up trying to do a `path.Split` when `path` is null

jchannon: Yeah I rock! :)

phillip-haydon: It's not possible to have a null path, so its an necessary check.

jchannon: Because it gets passed from GetOptions? What calls that? Could that pass a null to it?

phillip-haydon: HTTP Spec states a path must start with "/" is be empty. If a web server does not conform to this then the request is invalid. 

If you ABSOLUTELY want me to put the check in there. I can. It's not expensive to do the null check. It's just unnecessary.

thecodejunkie: Leave it out
On Feb 27, 2016 1:37 PM, "Phillip Haydon" <notifications@github.com> wrote:
>
> In src/Nancy/Routing/Trie/RouteResolverTrie.cs:
>
> >              {
> >                  return MatchResult.NoMatches;
> >              }
> >
> > -            return
this.routeTries[method].GetMatches(path.Split(splitSeparators,
StringSplitOptions.RemoveEmptyEntries), context)
> > -                                          .ToArray();
> > +            return result.GetMatches(path.Split(splitSeparators,
StringSplitOptions.RemoveEmptyEntries), context)
>
> HTTP Spec states a path must start with "/" is be empty. If a web server
does not conform to this then the request is invalid.
>
> If you ABSOLUTELY want me to put the check in there. I can. It's not
expensive to do the null check. It's just unnecessary.

PR #2316, opened 2016-02-26
nancy forms auth on coreclr

jchannon:


khellang:
Where's the xproj/sln changes?

jchannon:
there isnt any

khellang:
Ah, now I see. You just added `dotnet5.4`. It would be nice if you kept the diff cleaner so it would be easier to see what changed :smile:

jchannon:
i used ctrl+K+D to format the json according to VS

khellang:
Yeah, please keep formatting changes separate -- at least until things been reviewed. Then you can squash. This isn't easy to review:

image

jchannon:
You're a big boy, I'm sure you can manager :wink: 

Try adding a `?w=1` that should do the trick

phillip-haydon:
4 space tabbing please. Fix this!

jchannon:
How do you change this in VS? I'm using their defaults

phillip-haydon:
vs

jchannon:
Cool, just found it.

Wonder if we could enforce that in our R# dotsettings file? Although I imagine when you ctrl+k+d, R# won't intercept that

jchannon:
Reformatted

PR #2325, opened 2016-02-29
Swapped out CsQuery with AngleSharp

thecodejunkie:
CsQuery has not been maintained for quite a while not, which means that there's no CoreCLR support on the horizon. However, AngleSharp will be releasing a CoreCLR compatible version very soon.

This change is _mostly_ compatible with what we had before, because of our `* Wrapper` classes. However it does not support all the same CSS selectors since CsQuery was a jQuery "port" so it supported the jQuery specific selectors, while AngleSharp follows the CSS specifications

### DIFF
--- a/src/Nancy.Testing/DocumentWrapper.cs
+++ b/src/Nancy.Testing/DocumentWrapper.cs
@@ -1,17 +1,17 @@
 ﻿namespace Nancy.Testing
 {
     using System.Collections.Generic;
+    using System.IO;
     using System.Linq;
-    using System.Text;
-
-    using CsQuery;
+    using AngleSharp.Dom.Html;
+    using AngleSharp.Parser.Html;
 
     /// <summary>
     /// A basic wrapper around CsQuery
     /// </summary>
     public class DocumentWrapper
     {
-        private readonly CQ document;
+        private readonly IHtmlDocument document;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DocumentWrapper"/> class.
@@ -19,9 +19,11 @@
         /// <param name="buffer">The document represented as a byte array.</param>
         public DocumentWrapper(IEnumerable<byte> buffer)
         {
-            var utf8String = Encoding.UTF8.GetString(buffer.ToArray());
-
-            this.document = CQ.Create(utf8String);
+            var parser = new HtmlParser();
+            using (var stream = new MemoryStream(buffer.ToArray()))
+            {
+                this.document = parser.Parse(stream);
+            }
         }
 
         /// <summary>
@@ -31,7 +33,7 @@
         /// <returns>A <see cref="QueryWrapper"/> instance.</returns>
         public QueryWrapper this[string selector]
         {
-            get { return this.document.Select(selector); }
+            get { return new QueryWrapper(this.document.QuerySelectorAll(selector).ToArray()); }
         }
     }
 }
--- a/test/Nancy.Testing.Tests/BrowserFixture.cs
+++ b/test/Nancy.Testing.Tests/BrowserFixture.cs
@@ -6,17 +6,17 @@ namespace Nancy.Testing.Tests
     using System.Linq;
     using System.Security.Cryptography.X509Certificates;
     using System.Text;
+    using System.Collections.ObjectModel;
+    using System.Threading.Tasks;
     using Nancy.Extensions;
     using Nancy.Helpers;
     using Nancy.Session;
     using Nancy.Tests;
-    using Xunit;
-    using FakeItEasy;
     using Nancy.Authentication.Forms;
-    using System.Collections.ObjectModel;
-    using System.Threading.Tasks;
     using Nancy.Configuration;
     using Nancy.Tests.xUnitExtensions;
+    using Xunit;
+    using FakeItEasy;
 
     public class BrowserFixture
     {
--- a/test/Nancy.Testing.Tests/QueryWrapperTests.cs
+++ b/test/Nancy.Testing.Tests/QueryWrapperTests.cs
@@ -1,33 +1,38 @@
 namespace Nancy.Testing.Tests
 {
     using System.Linq;
-
-    using CsQuery;
-    using CsQuery.Implementation;
-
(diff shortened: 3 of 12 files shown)


thecodejunkie:
Ping @FlorianRappl mind having a quick look at the changes? Thanks :)

test/Nancy.Testing.Tests/BrowserFixture.cs
```diff
@@ -6,17 +6,17 @@ namespace Nancy.Testing.Tests
...
     using Nancy.Session;
     using Nancy.Tests;
-    using Xunit;
-    using FakeItEasy;
     using Nancy.Authentication.Forms;
-    using System.Collections.ObjectModel;
-    using System.Threading.Tasks;
     using Nancy.Configuration;
     using Nancy.Tests.xUnitExtensions;
+    using Xunit;
```
Yantrio: should these be in alphabetical order?

thecodejunkie: It's not something we've enforced before

test/Nancy.Testing.Tests/QueryWrapperTests.cs
```diff
@@ -35,12 +40,11 @@ public void Should_filter_selection_when_second_indexer_invoked()
...
             var document =
-                CQ.Create(
-                    @"<html><head></head><body><table><tr><td></td><td></td></tr><tr><td></td><td></td></tr></body></html>");
+                this.parser.Parse(@"<html><head></head><body><table><tr><td></td><td></td></tr><tr><td></td><td></td></tr></body></html>");
 
-            var wrapper = new QueryWrapper(document);
+            var wrapper = new QueryWrapper(new[] { document.DocumentElement });
 
-            var row = wrapper["tr:first"];
+            var row = wrapper["tr:first-child"];
```
Yantrio: How did this test pass in the firs place? `:first` will only return 1 element where `:first-child` should return 2 in this case.

Maybe we should put in an issue with CsQuery to state that `:first` was returning multiple elements?

test/Nancy.Testing.Tests/QueryWrapperTests.cs
```diff
@@ -35,12 +40,11 @@ public void Should_filter_selection_when_second_indexer_invoked()
         {
             // Given
             var document =
-                CQ.Create(
-                    @"<html><head></head><body><table><tr><td></td><td></td></tr><tr><td></td><td></td></tr></body></html>");
+                this.parser.Parse(@"<html><head></head><body><table><tr><td></td><td></td></tr><tr><td></td><td></td></tr></body></html>");
```
Yantrio: for some reason this html is missing the `</table>` tag.

src/Nancy.Testing/AssertExtensions.cs
```diff
@@ -93,7 +93,7 @@ public static AndConnector<QueryWrapper> ShouldBeOfClass(this QueryWrapper query
         /// </summary>
         public static AndConnector<NodeWrapper> ShouldContain(this NodeWrapper node, string contents, StringComparison comparisonType = StringComparison.Ordinal)
         {
-            Asserts.Contains(contents, node.InnerText, comparisonType);
+            Asserts.Contains(contents, node.InnerHtml, comparisonType);
```
FlorianRappl: Is `InnerText` in CsQuery the text representation or the HTML representation?

In AngleSharp `InnerHtml` is `innerHTML` from the W3C specification, i.e., things like `<` (in text) will be represented as `&lt;`, whereas `TextContent` would represent the text only (leaving out the nodes), thus representing, e.g., `<`, literally.

According to the summary I believe `TextContent` would be more appropriate.

src/Nancy.Testing/DocumentWrapper.cs
```diff
@@ -19,9 +19,9 @@ public class DocumentWrapper
         /// <param name="buffer">The document represented as a byte array.</param>
         public DocumentWrapper(IEnumerable<byte> buffer)
         {
-            var utf8String = Encoding.UTF8.GetString(buffer.ToArray());
-
-            this.document = CQ.Create(utf8String);
+            var parser = new HtmlParser();
+            var stream = new MemoryStream(buffer.ToArray());
```
FlorianRappl: I would place a using over these lines. Shouldn't matter much though as the stream will not keep an external handle alive.

test/Nancy.Testing.Tests/QueryWrapperTests.cs
Yantrio: infact, after running under jsfiddle, it looks like this shouldn't've ever worked, :first doesn't return anything, but :first-child does.[jsfiddle test](https://jsfiddle.net/hjac6Lyn/)

FlorianRappl:
Hi @thecodejunkie, this seems alright. I would, however, place `TextContent` instead of `InnerHtml` (for replacing `InnerText`).

test/Nancy.Testing.Tests/QueryWrapperTests.cs
thecodejunkie: ¯\_(ツ)_/¯ CsQuery is no longer maintained

Yantrio: meh, yolo, we can ignore it then :)

thecodejunkie:
@FlorianRappl thanks! Will sort it out right away :+1:

thecodejunkie:
Fixed feedback. Removed WIP status

grumpydev:
I knew abstracting CsQuery was a good idea :D

thecodejunkie:
@grumpydev I think the abstraction was put in place back during the HTML Agility Pack days? So even older :P

grumpydev:
Yeah, I think you're right, it's certainly earned its place that abstraction :p

thecodejunkie:
@grumpydev we need an `ABSTRACTION PROVEN` stamp :D

thecodejunkie:
Woohoo, all CI builds are green :)

grumpydev:
RIP CsQuery :)

PR #2326, opened 2016-02-29
Removed FakeRootPathProvider and the single use of it

thecodejunkie:
Removed the `FakeRootPathProvider` from `Nancy.Tests`. It was really only used by `Nancy.Demo.Authentication.Forms.TestingDemo` and even then it wasn't using the actual implementation since it was explicitly setting it's own rootpath. Killing it because `Assembly.GetEntryAssembly` is only available from rc2 and we weren't really using it


thecodejunkie:
Added `Breaking Change` to this since I've removed a public type. Super simple to replace if anyone needs it

samples/Nancy.Demo.Authentication.Forms.TestingDemo/TestRootPathProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+        public string GetRootPath()
+        {
+            var assemblyFilePath =
+                new Uri(typeof(FormsAuthBootstrapper).Assembly.CodeBase).LocalPath;
+
+            var assemblyPath =
+                Path.GetDirectoryName(assemblyFilePath);
+
+            var rootPath =
+                PathHelper.GetParent(assemblyPath, 3);
```
khellang: Woah, this is quite the hack :stuck_out_tongue_closed_eyes:

thecodejunkie: @khellang :stuck_out_tongue: it's the code from the old `FakeRootPathProvider`, just copied it over to kill the class. I think we should retire the `Nancy.Demo.Authentication.Forms.TestingDemo` sample and move it over to the Forms Authentication section of the wiki. The demo basically just shows how you can unit tests forms authentication code

PR #2327, opened 2016-03-01
Converted Nancy.Tests.Functional to project.json

thecodejunkie:
Bringing back our functional tests on `DNX net451` now that Razor has been re-written using Roslyn. This is a pre-requisite for bringing them onto `dotnet5_4` on the `coreclr`-branch

### DIFF
--- a/test/Nancy.Tests.Functional/project.json
+++ b/test/Nancy.Tests.Functional/project.json
@@ -1,0 +1,25 @@
+{
+    "dependencies": {
+        "Nancy": { "target": "project" },
+        "Nancy.Testing": { "target": "project" },
+        "Nancy.ViewEngines.Razor": { "target": "project" },
+        "xunit": "2.1.0",
+        "xunit.runner.dnx": "2.1.0-*"
+    },
+
+    "commands": {
+        "test": "xunit.runner.dnx"
+    },
+
+    "compilationOptions": {
+        "define": [ "DNX" ]
+    },
+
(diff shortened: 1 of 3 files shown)


test/Nancy.Tests.Functional/project.json
```diff
@@ -0,0 +1,25 @@
...
+    "commands": {
+        "test": "xunit.runner.dnx"
+    },
+
+    "compilationOptions": {
+        "define": [ "DNX" ]
+    },
+
+    "compileFiles": [
+        "../Nancy.Tests/xUnitExtensions/RecordAsync.cs"
```
khellang: Just a sanity (copy-paste) check; is this used?

thecodejunkie: Yep, we have a couple of tests in `Nancy.Tests.Functional` that make use of those extension methods

khellang: Also, what's the diff between `compileFiles` and `compile`? :stuck_out_tongue_closed_eyes:

khellang: OK, from the documentation it seems it'll glob `compile` and `compileBuiltIn` (both allows wildcards), exclude `compileExclude` and then add the individual `compileFiles`. So there should be no difference, other than `compile` allows globbing, and `compileFiles` doesn't.

thecodejunkie:
All checks :white_check_mark:

PR #2333, opened 2016-03-02
Made Nancy.Tests.Functional run on on the CoreCLR

thecodejunkie:
Had to ifdef a couple of tests that are using Razor

### DIFF
--- a/src/Nancy/Extensions/TypeExtensions.cs
+++ b/src/Nancy/Extensions/TypeExtensions.cs
@@ -11,6 +11,31 @@
     public static class TypeExtensions
     {
         /// <summary>
+        /// Creates an instance of <paramref name="type"/> and cast it to <typeparamref name="T"/>.
+        /// </summary>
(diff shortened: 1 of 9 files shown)


src/Nancy/Extensions/TypeExtensions.cs
```diff
@@ -178,5 +203,46 @@ public static TypeCode GetTypeCode(this Type type)
...
+        {
+            var typeInfo = type.GetTypeInfo();
+
+            var constructors = typeInfo.DeclaredConstructors;
+
+            foreach (var constructor in constructors)
+            {
+                var parameters = constructor.GetParameters();
+
+                if (parameters.Length > 0)
```
Yantrio: could use .Any()

thecodejunkie: @Yantrio a `.Lenght` should be `O(1)` since it is not a calculated property, where as `.Count()` would be slower

PR #2340, opened 2016-03-08
Added NuGet restore task

thecodejunkie:
We're no longer putting the `packages` folder in the repo so we need to make sure that the `rake` script can build from a clean repo


jchannon:
LGTM

PR #2354, opened 2016-03-14
Performance tweaks for the RequestHeaders class

thecodejunkie:
I made some modifications to the `RequestHeaders.cs` class, more specifically to the headers that can have quality values assigned and calls the `GetWeightedValues` method internally. The value parsing was re-written to be a forward-reading algorithm, with reduced string manipulation (splits etc). This is the first pass and there are probably more improvements that can be performed so reviews and recommendations are more than welcome

Also removed from reflection + `Activator.CreateInstance` code that were used to get default fallback values based on a `Type` which in certain circumstances included creating a generic type on the fly.

Before
image

After
image

All tests pass, though there might be some unintentional side-effects for how the values are parsed that perhaps we do not have test coverage for

### DIFF
--- a/src/Nancy/RequestHeaders.cs
+++ b/src/Nancy/RequestHeaders.cs
@@ -33,7 +33,7 @@ namespace Nancy
         /// <value>An <see cref="IEnumerable{T}"/> that contains the header values if they are available; otherwise it will be empty.</value>
         public IEnumerable<Tuple<string, decimal>> Accept
         {
-            get { return this.GetWeightedValues("Accept").ToList(); }
+            get { return this.GetWeightedValues("Accept"); }
             set { this.SetHeaderValues("Accept", value, GetWeightedValuesAsStrings); }
         }
 
@@ -73,7 +73,7 @@ namespace Nancy
         /// <value>A <see cref="string"/> containing the header value if it is available; otherwise <see cref="string.Empty"/>.</value>
         public string Authorization
         {
-            get { return this.GetValue("Authorization", x => x.First()); }
+            get { return this.GetValue("Authorization", x => x.First(), string.Empty); }
             set { this.SetHeaderValues("Authorization", value, x => new[] { x }); }
         }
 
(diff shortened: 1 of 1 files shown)


src/Nancy/RequestHeaders.cs
```diff
@@ -112,27 +112,27 @@ public string Connection
         /// <value>The length of the contents if it is available; otherwise 0.</value>
         public long ContentLength
         {
-            get { return this.GetValue("Content-Length", x => Convert.ToInt64(x.First())); }
+            get { return this.GetValue("Content-Length", x => Convert.ToInt64(x.First()), 0); }
```
khellang: Shouldn't this really be a `long?`, and return `null` if the header is absent?

thecodejunkie: Quite possibly, but that's a change that would be unrelated to this pull-request

src/Nancy/RequestHeaders.cs
```diff
@@ -372,17 +428,19 @@ private static IEnumerable<INancyCookie> GetNancyCookies(IEnumerable<string> coo
 
         private IEnumerable<string> GetValue(string name)
         {
-            return this.GetValue(name, x => x);
+            return this.GetValue(name, x => x, new string[] {});
```
khellang: `Array.Empty<string>`? Or is that a 4.6+ API? Maybe we should have an `EmptyArrayCache<T>` class, similar to [MVC's TaskCache](https://github.com/aspnet/Mvc/blob/4f709bdbfdad5150ca941808ecbe3a15b9cc9089/src/Microsoft.AspNetCore.Mvc.Core/Internal/TaskCache.cs)? Either that, or we simply cache an empty string array in this class.

thecodejunkie: Yeah that's `4.6+`. I initially had it as `Enumerable.Empty<string>()` and that performed worse, so I decided to just K.I.S.S and stick with the primitive

khellang: For `Enumerable.Empty` you'd have to allocate an enumerator for each time as well, so that makes sense. I think we should at least cache this instance, if we don't want a generic empty array cache.

jchannon:
Tests pass & perf increased. :shipit:

PR #2356, opened 2016-03-14
Added ReflectionTypeLoadException to SafeGetExportedTypes method

jchannon:
This exception seems to be thrown on Mono but not on Windows, I tracked it down to a 3rd party dll that `GetExportedTypes` was being called on.

### DIFF
--- a/src/Nancy/Extensions/AssemblyExtensions.cs
+++ b/src/Nancy/Extensions/AssemblyExtensions.cs
@@ -31,8 +31,13 @@ namespace Nancy.Extensions
             {
                 types = new Type[] { };
             }
-            catch (FileLoadException) {
+            catch (FileLoadException)
+            {
                 // probably assembly version conflict
+                types = new Type[] { };
+            }
+            catch (ReflectionTypeLoadException)
+            {
                 types = new Type[] { };
             }
             return types;


khellang:
This has me worried. Why would this suddenly start throwing `ReflectionTypeLoadException`s? I don't think we'll ever call `SafeGetExportedTypes` on a reflection-only loaded assembly? I think this might be a bug in the new `AppDomainAssemblyCatalog`/`DefaultTypeCatalog`... Earlier, ADATS used to [filter out reflection-only assemblies](https://github.com/NancyFx/Nancy/blob/v1.4.3/src/Nancy/Bootstrapper/AppDomainAssemblyTypeScanner.cs#L208), but I'm not sure that the new setup does the same.

khellang:
I see there's a check [here](https://github.com/NancyFx/Nancy/blob/59fc06c44c49288d51adff6f601a92dd9b6624f2/src/Nancy/AppDomainAssemblyCatalog.cs#L42), but I think I'd rather see that check in the type catalog.

jchannon:
The exception I see is called from DefaultTypeCatalog here https://github.com/NancyFx/Nancy/blob/master/src/Nancy/DefaultTypeCatalog.cs#L35, notice https://github.com/NancyFx/Nancy/blob/master/src/Nancy/DefaultTypeCatalog.cs#L42 which is where it calls `SafeGetExportedTypes`

thecodejunkie:
@jchannon then stuff you showed me before I had to leave, had almost narrowed it down. We need to continue going down that rabbit whole until we actually have an explanation as to "why now?". Like I said, to me it is worrying that it started happen on the `2.0` work and slapping another `catch`-clause on there might be as good as slapping a bandied on a broken leg. Semantically nothing is supposed to have changed between `1.x` and `2.0` at this level, so if we are seeing a new problem emerge on an existing code base then we need to understand why.

In the screenshot it said `System.IO.FileNotFoundException: Could not load file or assembly or one of its dependencies. File na...` and then it was cut-off. What does the rest say? Why would we all of a sudden be experiencing problems with `antlr.runtime`? I assume this is not a new dependency that you introduced at the same time you upgraded to latest master?

This pull-request _might_ be the proper solution, but until we know _why_ then we should not pull this in

thecodejunkie:
@jchannon so this can be closed now?

PR #2357, opened 2016-03-14
Cache empty arrays

khellang:
Quick structural search & replace with R# :smile:

### DIFF
--- a/src/Nancy/ArrayCache.cs
+++ b/src/Nancy/ArrayCache.cs
@@ -1,0 +1,22 @@
+﻿namespace Nancy
+{
+    /// <summary>
+    /// A cache for empty arrays.
+    /// </summary>
+    public class ArrayCache
+    {
+        /// <summary>
+        /// Gets a cached, empty array of the specified type.
+        /// </summary>
+        /// <typeparam name="T">the type of array to get.</typeparam>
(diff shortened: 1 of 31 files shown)


jchannon:
What does this solve?

khellang:
> What does this solve?

It "solves" a bunch of allocations :smile:

src/Nancy/ArrayCache.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Nancy
+{
+    public class ArrayCache
```
thecodejunkie: You know better than to try and sneak undocumented classes below my radar :stuck_out_tongue_closed_eyes:

src/Nancy/ArrayCache.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public class ArrayCache
+    {
+        public static T[] Empty<T>()
+        {
+            return EmptyArray<T>.Value;
+        }
+    }
+
+    internal static class EmptyArray<T>
```
thecodejunkie: Maybe stick this as a nested class of `ArrayCache` ?

thecodejunkie:
Neat little trick :+1:

PR #2361, opened 2016-03-15
Optimizaiton of DefaultModelBinderLocator

thecodejunkie:
Minor tweak while chasing bigger dragons. The `DefaultModelBinderLocator` figured out which model binder is should use for a specific model type. This is part of a 100k requests sample where model binding is involved

- Removed use of LINQ in the resolving

Before
image

After
image

### DIFF
--- a/src/Nancy/ModelBinding/DefaultModelBinderLocator.cs
+++ b/src/Nancy/ModelBinding/DefaultModelBinderLocator.cs
@@ -12,7 +12,7 @@ namespace Nancy.ModelBinding
         /// <summary>
         /// Available model binders
         /// </summary>
-        private readonly IEnumerable<IModelBinder> binders;
+        private readonly IReadOnlyCollection<IModelBinder> binders;
 
         /// <summary>
         /// Default model binder to fall back on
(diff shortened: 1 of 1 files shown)


src/Nancy/ModelBinding/DefaultModelBinderLocator.cs
```diff
@@ -27,7 +27,11 @@ public class DefaultModelBinderLocator : IModelBinderLocator
         public DefaultModelBinderLocator(IEnumerable<IModelBinder> binders, IBinder fallbackBinder)
         {
             this.fallbackBinder = fallbackBinder;
-            this.binders = binders;
+
+            if (binders != null)
+            {
+                this.binders = binders.ToArray();
```
phillip-haydon: This is the same as my comment on the other issue, but I believe you can do:

`this.binders = new ReadOnlyCollection(binders);`

This means it will work on the original collection and not copy it to a new one.

thecodejunkie: We can't wrap it because it's `IEnumerable<T>` which means we need to materialize the collection first using either `ToList` or `ToArray`.

phillip-haydon: As discussed in chat, can't cast the value :(

PR #2363, opened 2016-03-15
Optimization of CheckStatusCodeHandler in NancyEngine

thecodejunkie:
Minor tweak while chasing bigger dragons. The `CheckStatusCodeHandler` runs on every request to see if it should be taking any special action based on the returned status code. This is part of a 100k requests sample

- Removed LINQ magic

This loops over the `IStatusCodeHandler` once and mimics of a `FirstOrDefault` by only assigning values if nothing has previously been written. If both the `default` and `custom` handlers has been assigned, then it will exist the loop early

Before
image

After
image

### DIFF
--- a/src/Nancy/NancyEngine.cs
+++ b/src/Nancy/NancyEngine.cs
@@ -27,7 +27,7 @@
         private readonly IRequestDispatcher dispatcher;
         private readonly INancyContextFactory contextFactory;
         private readonly IRequestTracing requestTracing;
-        private readonly IEnumerable<IStatusCodeHandler> statusCodeHandlers;
+        private readonly IReadOnlyCollection<IStatusCodeHandler> statusCodeHandlers;
         private readonly IStaticContentProvider staticContentProvider;
         private readonly IResponseNegotiator negotiator;
         private readonly CancellationTokenSource engineDisposedCts;
@@ -83,7 +83,7 @@
 
             this.dispatcher = dispatcher;
             this.contextFactory = contextFactory;
-            this.statusCodeHandlers = statusCodeHandlers;
+            this.statusCodeHandlers = statusCodeHandlers.ToArray();
             this.requestTracing = requestTracing;
             this.staticContentProvider = staticContentProvider;
             this.negotiator = negotiator;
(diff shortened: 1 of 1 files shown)


src/Nancy/NancyEngine.cs
```diff
@@ -217,17 +217,42 @@ private void CheckStatusCodeHandler(NancyContext context)
...
+                    {
+                        break;
+                    }
+                }
+            }
 
-            var customHandler = handlers
-                .FirstOrDefault(x => !(x is DefaultStatusCodeHandler));
+            //var handlers = this.statusCodeHandlers
+            //    .Where(x => x.HandlesStatusCode(context.Response.StatusCode, context))
```
khellang: :scream_cat:

src/Nancy/NancyEngine.cs
```diff
@@ -217,17 +217,42 @@ private void CheckStatusCodeHandler(NancyContext context)
...
 
-            var defaultHandler = handlers
-                .FirstOrDefault(x => x is DefaultStatusCodeHandler);
+            foreach (var statusCodeHandler in this.statusCodeHandlers)
+            {
+                if (statusCodeHandler.HandlesStatusCode(context.Response.StatusCode, context))
+                {
+                    if (defaultHandler == null && (statusCodeHandler is DefaultStatusCodeHandler))
+                    {
+                        defaultHandler = statusCodeHandler;
```
khellang: couldn't you just `continue` here?

src/Nancy/NancyEngine.cs
```diff
@@ -217,17 +217,42 @@ private void CheckStatusCodeHandler(NancyContext context)
...
+                    }
+
+                    if (customHandler == null && !(statusCodeHandler is DefaultStatusCodeHandler))
+                    {
+                        customHandler = statusCodeHandler;
```
khellang: same here. it's not possible that a handler can be both custom and default, right?

src/Nancy/NancyEngine.cs
thecodejunkie: Whoops :astonished:

src/Nancy/NancyEngine.cs
```diff
@@ -83,7 +83,7 @@ public class NancyEngine : INancyEngine
 
             this.dispatcher = dispatcher;
             this.contextFactory = contextFactory;
-            this.statusCodeHandlers = statusCodeHandlers;
+            this.statusCodeHandlers = statusCodeHandlers.ToArray();
```
phillip-haydon: Calling `ToArray()` will copy the list passed in, can't we just make the argument array to begin with and avoid the copy?

src/Nancy/NancyEngine.cs
```diff
@@ -217,17 +217,34 @@ private void CheckStatusCodeHandler(NancyContext context)
...
+            IStatusCodeHandler customHandler = null;
 
-            var defaultHandler = handlers
-                .FirstOrDefault(x => x is DefaultStatusCodeHandler);
-
-            var customHandler = handlers
-                .FirstOrDefault(x => !(x is DefaultStatusCodeHandler));
+            foreach (var statusCodeHandler in this.statusCodeHandlers)
+            {
+                if (statusCodeHandler.HandlesStatusCode(context.Response.StatusCode, context))
```
phillip-haydon: Can we invert the if so we don't have such nested code. 

```
if (!...) 
{
continue;
}

// stuffs thats inside the if
```

PR #2365, opened 2016-03-15
Optimization of ModuleExtensions.GetModuleName

thecodejunkie:
Minor tweak while chasing bigger dragons. The `ModuleExtensions.GetModuleName` simply extracts the name of a module, by attempting to strip out the `Module`-suffix. This is part of a 100k requests sample

- Removed use of a Regex in favor of some string manipulation

Before
image

After
image

### DIFF
--- a/src/Nancy/Extensions/ModuleExtensions.cs
+++ b/src/Nancy/Extensions/ModuleExtensions.cs
@@ -1,9 +1,6 @@
 namespace Nancy.Extensions
 {
     using System;
-    using System.Diagnostics;
-    using System.Text.RegularExpressions;
-
     using Nancy.ErrorHandling;
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


src/Nancy/Extensions/ModuleExtensions.cs
```diff
@@ -12,31 +9,22 @@ namespace Nancy.Extensions
...
         /// <returns>A string containing the name of the parameter.</returns>
-        /// <exception cref="FormatException"></exception>
         public static string GetModuleName(this INancyModule module)
         {
             var typeName = module.GetType().Name;
-            var nameMatch =
-                ModuleNameExpression.Match(typeName);
 
-            if (nameMatch.Success)
+            var offset = typeName.LastIndexOf("Module", StringComparison.Ordinal);
```
jchannon: Should this be OrdinalIgnoreCase to be consistent with code elsewhere

thecodejunkie: Would be a breaking change since the regex was not case sensitive. Decided to leave it as `Ordinal` as the goal of the change was performance improvements, not behavior change

PR #2369, opened 2016-03-17
RFC 5899 Compliant Link Header (master)

asbjornu:
This PR adds a few `HttpLink` classes to make the `Link` HTTP header being produced by `DefaultResponseNegotiator` a bit more [RFC 5899](https://tools.ietf.org/html/rfc5988) compliant. The parsing algorithm is not a full ABNF implementation, but I believe it's good enough to claim RFC 5899 compliancy.

To explain why this PR isn't just a simple switch of the `rel` and addition of `type` parameters in `DefaultResponseNegotiator.CreateLinkHeader()`, I believe that when an HTTP server framework adds a header, it's the framework's responsibility to ensure that the header is properly implemented according to the RFC in which it is defined and not leave this up to users of the framework.

Additionally, the `DefaultResponseNegotiator.AddLinkHeader()` method didn't check if a `Link` header already existed, effectively replacing any existing `Link` headers with its own. This is the particular problem that made me start looking into the issue to begin with.

This PR attempts to fix all this and while it might seem massive, I feel this is what's required.

### DIFF
--- a/src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
+++ b/src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
@@ -359,7 +359,7 @@
             var baseUrl = string.Concat(requestUrl.BasePath, "/", fileName);
 
             var links = linkProcessors
-                .Select(lp => string.Format("<{0}.{1}>; rel=\"{2}\"", baseUrl, lp.Key, lp.Value));
+                .Select(lp => string.Format("<{0}.{1}>; rel=\"alternate\"; type=\"{2}\"", baseUrl, lp.Key, lp.Value));
 
             return string.Join(",", links);
         }
--- a/src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
+++ b/src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
@@ -233,10 +233,9 @@
         /// <param name="negotiationContext">The negotiation context.</param>
         /// <param name="context">The context.</param>
         /// <returns>A <see cref="Response"/>.</returns>
-        private static Response CreateResponse(
-            IList<CompatibleHeader> compatibleHeaders,
-            NegotiationContext negotiationContext,
-            NancyContext context)
+        private Response CreateResponse(IList<CompatibleHeader> compatibleHeaders,
+                                        NegotiationContext negotiationContext,
+                                        NancyContext context)
         {
             var response = NegotiateResponse(compatibleHeaders, negotiationContext, context);
 
@@ -250,7 +249,7 @@
 
             response.WithHeader("Vary", "Accept");
 
-            AddLinkHeader(compatibleHeaders, response, context.Request.Url);
+            this.AddLinkHeader(compatibleHeaders, response, context.Request.Url);
             SetStatusCode(negotiationContext, response);
             SetReasonPhrase(negotiationContext, response);
             AddCookies(negotiationContext, response);
@@ -353,7 +352,7 @@
         /// <param name="linkProcessors">The link processors.</param>
         /// <param name="existingLinkHeader">The existing Link HTTP Header.</param>
         /// <returns>The link header.</returns>
-        private static string CreateLinkHeader(Url requestUrl, IEnumerable<KeyValuePair<string, MediaRange>> linkProcessors, string existingLinkHeader)
+        protected virtual string CreateLinkHeader(Url requestUrl, IEnumerable<KeyValuePair<string, MediaRange>> linkProcessors, string existingLinkHeader)
         {
             var fileName = Path.GetFileNameWithoutExtension(requestUrl.Path);
             var baseUrl = string.Concat(requestUrl.BasePath, "/", fileName);
--- a/src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
+++ b/src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
@@ -310,15 +310,14 @@
         /// <param name="compatibleHeaders">The compatible headers.</param>
         /// <param name="response">The response.</param>
         /// <param name="requestUrl">The request URL.</param>
-        private static void AddLinkHeader(
-            IEnumerable<CompatibleHeader> compatibleHeaders,
-            Response response,
-            Url requestUrl)
+        private void AddLinkHeader(IEnumerable<CompatibleHeader> compatibleHeaders, Response response, Url requestUrl)
         {
             var linkProcessors = GetLinkProcessors(compatibleHeaders, response.ContentType);
             if (linkProcessors.Any())
             {
-                response.Headers["Link"] = CreateLinkHeader(requestUrl, linkProcessors);
+                string existingLinkHeader;
+                response.Headers.TryGetValue("Link", out existingLinkHeader);
+                response.Headers["Link"] = this.CreateLinkHeader(requestUrl, linkProcessors, existingLinkHeader);
             }
         }
 
@@ -352,8 +351,9 @@
         /// </summary>
         /// <param name="requestUrl">The request URL.</param>
         /// <param name="linkProcessors">The link processors.</param>
+        /// <param name="existingLinkHeader">The existing Link HTTP Header.</param>
         /// <returns>The link header.</returns>
-        private static string CreateLinkHeader(Url requestUrl, IEnumerable<KeyValuePair<string, MediaRange>> linkProcessors)
+        private static string CreateLinkHeader(Url requestUrl, IEnumerable<KeyValuePair<string, MediaRange>> linkProcessors, string existingLinkHeader)
         {
             var fileName = Path.GetFileNameWithoutExtension(requestUrl.Path);
             var baseUrl = string.Concat(requestUrl.BasePath, "/", fileName);
@@ -361,7 +361,12 @@
             var links = linkProcessors
                 .Select(lp => string.Format("<{0}.{1}>; rel=\"alternate\"; type=\"{2}\"", baseUrl, lp.Key, lp.Value));
 
-            return string.Join(",", links);
+            if (!string.IsNullOrEmpty(existingLinkHeader))
+            {
+                links = links.Concat(new[] { existingLinkHeader });
+            }
+
+            return string.Join(", ", links);
         }
 
         /// <summary>
--- a/test/Nancy.Tests.Functional/Tests/JsonLdProcessor.cs
+++ b/test/Nancy.Tests.Functional/Tests/JsonLdProcessor.cs
@@ -1,0 +1,41 @@
+﻿namespace Nancy.Tests.Functional.Tests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Nancy.Responses.Negotiation;
+
+    public class JsonLdProcessor : IResponseProcessor
+    {
+        private readonly ISerializer serializer;
+
+        public JsonLdProcessor(IEnumerable<ISerializer> serializers)
+        {
+            this.serializer = serializers.FirstOrDefault(x => x.CanSerialize("application/json"));
+        }
+
+        public IEnumerable<Tuple<string, MediaRange>> ExtensionMappings
+        {
+            get { return new[] { new Tuple<string, MediaRange>("json", new MediaRange("application/json")) }; }
+        }
+
+        public ProcessorMatch CanProcess(MediaRange requestedMediaRange, dynamic model, NancyContext context)
+        {
+            return new ProcessorMatch
+            {
+                ModelResult = MatchResult.DontCare,
+                RequestedContentTypeResult = MatchResult.DontCare
+            };
+        }
+
+        public Response Process(MediaRange requestedMediaRange, dynamic model, NancyContext context)
+        {
+            return new Response
+            {
+                ContentType = "application/json",
+                Contents = stream => this.serializer.Serialize("application/json", model, stream),
+                StatusCode = HttpStatusCode.OK
+            }.WithHeader("Link", "</context.jsonld>; rel=\"http://www.w3.org/ns/json-ld#context\"; type=\"application/ld+json\"");
+        }
+    }
+}
--- a/test/Nancy.Tests.Functional.MSBuild/Nancy.Tests.Functional.csproj
+++ b/test/Nancy.Tests.Functional.MSBuild/Nancy.Tests.Functional.csproj
@@ -94,6 +94,9 @@
     <Compile Include="..\Nancy.Tests.Functional\Properties\AssemblyInfo.cs">
       <Link>Properties\AssemblyInfo.cs</Link>
     </Compile>
+    <Compile Include="..\Nancy.Tests.Functional\Tests\JsonLdProcessor.cs">
+      <Link>Tests\JsonLdProcessor.cs</Link>
+    </Compile>
     <Compile Include="..\Nancy.Tests\xUnitExtensions\RecordAsync.cs">
       <Link>RecordAsync.cs</Link>
     </Compile>
--- a/test/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
+++ b/test/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
@@ -461,9 +461,9 @@ namespace Nancy.Tests.Functional.Tests
             var response = await browser.Get("/");
 
             // Then
-            Assert.True(response.Headers["Link"].Contains(@"</.foo>; rel=""foo/bar"""));
-            Assert.True(response.Headers["Link"].Contains(@"</.json>; rel=""application/json"""));
-            Assert.True(response.Headers["Link"].Contains(@"</.xml>; rel=""application/xml"""));
+            Assert.True(response.Headers["Link"].Contains(@"</.foo>; rel=""alternate""; type=""foo/bar"""));
+            Assert.True(response.Headers["Link"].Contains(@"</.json>; rel=""alternate""; type=""application/json"""));
+            Assert.True(response.Headers["Link"].Contains(@"</.xml>; rel=""alternate""; type=""application/xml"""));
         }
 
         [Fact]
--- a/test/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
+++ b/test/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
@@ -467,6 +467,28 @@ namespace Nancy.Tests.Functional.Tests
         }
 
         [Fact]
+        public async Task Should_preserve_existing_link_header()
+        {
+            // Given
+            var browser = new Browser(with =>
+            {
+                with.ResponseProcessors(typeof(XmlProcessor), typeof(JsonLdProcessor));
+
+                with.Module(new ConfigurableNancyModule(x =>
+                {
+                    x.Get("/", CreateNegotiatedResponse());
+                }));
+            });
+
+            // When
+            var response = await browser.Get("/");
+
+            // Then
+            Assert.True(response.Headers["Link"].Contains(@"</context.jsonld>; rel=""http://www.w3.org/ns/json-ld#context""; type=""application/ld+json"""));
+            Assert.True(response.Headers["Link"].Contains(@"</.xml>; rel=""alternate""; type=""application/xml"""));
+        }
+
+        [Fact]
         public async Task Should_set_negotiated_status_code_to_response_when_set_as_integer()
         {
             // Given


asbjornu:
Ok, I'll obviously look into why Travis and TC is exploding.

asbjornu:
Ah. Seems like TeamCity and Travis requires C# 5 while AppVeyor has no problem building C# 6. Should I revert to C# 5? If so, I'd argue this should be added to the `.DotSettings` file so ReSharper doesn't suggest C# 6 features in its code inspection. /cc @khellang

khellang:
> Should I revert to C# 5?

Yes, please. We're literally just making changes to support C# 6, but we'd like to keep master building for now.

Nancy.sln.DotSettings
```diff
@@ -33,6 +33,8 @@
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/STICK_COMMENT/@EntryValue">False</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CSharpUsing/AddImportsToDeepestScope/@EntryValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CSharpUsing/QualifiedUsingAtNestedScope/@EntryValue">True</s:Boolean>
+	<s:String x:Key="/Default/CodeStyle/FileHeader/FileHeaderRegionName/@EntryValue"></s:String>
+	<s:String x:Key="/Default/CodeStyle/FileHeader/FileHeaderText/@EntryValue"></s:String>
```
khellang: What are these?

test/Nancy.Tests/Unit/HttpLinkHeaderFixture.cs
```diff
@@ -0,0 +1,136 @@
...
+    public class HttpLinkHeaderFixture
+    {
+        private const string LinkHeader = "<http://nancyfx.org/>; "
+            + "rel=\"http://nancyfx.org/rel/Nancy\"; "
+            + "title=\"Nancy\", "
+            + "</alternate.jsonld>; "
+            + "rel=\"alternate\"; "
+            + "title=\"JSON LD Representation\"";
+
+
```
khellang: Can we remove the double blank lines here?

test/Nancy.Tests/Unit/HttpLinkHeaderFixture.cs
```diff
@@ -0,0 +1,136 @@
+namespace Nancy.Tests.Unit
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
```
khellang: :fire: this blank line

test/Nancy.Tests/Unit/HttpLinkFixture.cs
```diff
@@ -0,0 +1,148 @@
...
+            link.Relation.ShouldEqual(rel);
+            link.TargetUri.ShouldEqual(uri);
+            link.Title.ShouldBeNull();
+        }
+
+
+        [Fact]
+        public void Parse_link_with_title_and_type_should_return_link_object_with_title_and_type()
+        {
+            var rel = new HttpLinkRelation("http://nancyfx.org/rels/Nancy");
```
khellang: We try to please @thecodejunkie's OCD by adding

```
// Given

// When

// Then
```

to our test. Mind adding them?

khellang:
Woah. This is quite the (breaking) change. Good thing we're going 2.0 right now :smile: I pointed out some minor things while skimming through. I'll have to take some time to review this later.

test/Nancy.Tests/Unit/HttpLinkHeaderFixture.cs
khellang: This obviously goes for all files :smile:

test/Nancy.Tests/Unit/HttpLinkHeaderFixture.cs
khellang: This obviously goes for all files :smile:

test/Nancy.Tests/Unit/HttpLinkFixture.cs
khellang: This obviously goes for all files :smile:

Nancy.sln.DotSettings
asbjornu: The commit message for c90d342e334e93848750c714c5fe6ae5f5595f83 explains it quite well, I think?

> Explicitly set the ReSharper file header stuff to blank so global config is not used instead.

I have ReSharper with a standard file header that is added when I do cleanup since Nancy does not explicitly set the header to blank. Since Nancy has no headers, I think it make sense setting this to nothing, agree?

test/Nancy.Tests/Unit/HttpLinkHeaderFixture.cs
asbjornu: All files are formatted according to the accompanying ReSharper `.DotSettings` file. Do you want me to tweak the settings in it so they don't add this whitespace?

Nancy.sln.DotSettings
thecodejunkie: This is change is unrelated to the pull-request. For the sake of tractability it would be better to send a separate pull-request with these changes

asbjornu: @thecodejunkie Ok, sure. I'll split it out to a separate PR. Should I put more R# related stuff into it, such as d064f3b09ac5110c7f60c34be8cd42268797a99d too?

asbjornu:
@khellang Thanks for the code skimming. I've corrected the things you pointed out, but Travis is still failing. Seems like `System.Uri.TryCreate()` is doing something very weird on Mono, causing relative paths to be interpreted as `file://` URIs. Expected output:

```
</.foo>; rel="alternate"; type="foo/bar"
```

Actual output:

```
<file:///.foo>; rel="alternate"; type="foo/bar"
```

Since this behaviour differs from Windows, I can only assume this is a bug in Mono's `System.Uri` implementation, but we obviously have to figure out a way around it. Sigh.

khellang:
This might provide some background: http://www.mono-project.com/docs/faq/known-issues/urikind-relativeorabsolute/

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,504 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+    using Nancy.Responses.Negotiation;
+
+    /// <summary>
+    /// Represents one of possibly many RFC 5899 HTTP Links contained in a <see cref="HttpLinkHeader"/>.
+    /// </summary>
+    public class HttpLink : Dictionary<string, object>, IEquatable<HttpLink>
```
thecodejunkie: This class has _a lot_ of constructor overloads - 11 if I am counting correctly. All off these can't be needed?

thecodejunkie: Generally I think the parameter documentation needs some attention "The x, The y, The z" is not really providing any value. This goes for all methods / constructors

test/Nancy.Tests.Functional/Tests/JsonLdProcessor.cs
```diff
@@ -0,0 +1,49 @@
+﻿namespace Nancy.Tests.Functional.Tests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    using Nancy.Responses.Negotiation;
+
+    public class JsonLdProcessor : IResponseProcessor
```
thecodejunkie: Lots of double blank lines in this file

test/Nancy.Tests/Unit/HttpLinkHeaderFixture.cs
```diff
@@ -0,0 +1,160 @@
...
+        {
+            // Given
+            var linkHeader = new HttpLinkHeader();
+
+            // When
+            linkHeader.Add(new HttpLink("/first", "ALTERNATE"));
+            linkHeader.Add(new HttpLink("/first", "AlterNate"));
+            linkHeader.Add(new HttpLink("/first", "alternate"));
+            linkHeader.Add(new HttpLink("/first", HttpLinkRelation.IanaLinkRelationPrefix + "alternate"));
+            // TODO: Figure out what to do with duplicates as per https://github.com/mnot/I-D/issues/170. @asbjornu
```
thecodejunkie: Was this resolved?

asbjornu:
[This Mono bug](https://bugzilla.xamarin.com/show_bug.cgi?id=30854) references [this explanation](http://www.mono-project.com/docs/faq/known-issues/urikind-relativeorabsolute/):

>In .NET `new Uri ("/foo", UriKind.RelativeOrAbsolute)` returns a relative Uri whereas Mono assumes it is an absolute file path.
>
>Mono diverges from .NET because UNIX file paths starting with ‘/’ are absolute file paths.
>
>Changing the behavior to match the .NET one is not possible without breaking existing Mono applications that rely on `UriKind.RelativeOrAbsolute` to parse absolute file paths.

There is, thankfully, a few workarounds available, such as:

```csharp
var uri = new Uri (str, (str.StartsWith ("/")) ? UriKind.Relative : UriKind.RelativeOrAbsolute)
```

I'll experiment with which works better and rework the PR.

asbjornu:
@thecodejunkie Thanks for the review. I've reworked the `HttpLink` class so there's now only two constructors with optional parameters and reworded the documentation to what RFC 5899 actually says about the parameters. All extraneous whitespace should also be removed.

test/Nancy.Tests/Unit/HttpLinkHeaderFixture.cs
asbjornu: @thecodejunkie I'm still discussing with @mnot in mnot/I-D#170. Once that's resolved and there's no more comments from the Custodians of the Super-Duper-Happy-Path, I'll squash (and rebase, if necessary) and force-push.

src/Nancy.MSBuild/Nancy.csproj
```diff
@@ -1364,6 +1373,7 @@
     <Compile Include="..\Nancy\Configuration\ConfigurationException.cs">
       <Link>Configuration\ConfigurationException.cs</Link>
     </Compile>
+    <Compile Include="HttpLinkParser.cs" />
```
khellang: This file should sit in `..\Nancy.HttpLinkParser.cs` :smile:

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,257 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="HttpLink" /> class.
+        /// </summary>
+        /// <param name="targetUri">The target URI of the link.</param>
+        /// <param name="relation">The relation that identifies the semantics of the link.</param>
+        /// <param name="type">The optional type parameter is a hint indicating what the media type of the result of dereferencing the link should be. Note that this is only a hint; for example, it does not override the HTTP Content-Type header of a HTTP response obtained by actually following the link.</param>
+        /// <param name="title">The optional title parameter is used to label the destination of a link such that it can be used as a human-readable identifier (e.g., a menu entry) in the language indicated by the HTTP Content-Language header (if present).</param>
+        /// <exception cref="System.ArgumentNullException"><paramref name="targetUri" /> or <paramref name="relation" /></exception>
+        public HttpLink(string targetUri, string relation, string type = null, string title = null)
```
khellang: If we can avoid it, I'd like to get rid of these optional parameters. [They don't version very well](http://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/)

src/Nancy.MSBuild/HttpLinkParser.cs
```diff
@@ -0,0 +1,147 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
+
+    internal class HttpLinkParser
+    {
+        private readonly string httpLink;
+
+        public HttpLinkParser(string httpLink)
```
khellang: Why does this need state? Can't it just be a static class?

khellang:
I'm worried about all the allocations we're doing here. I see some LINQ and lots of `ToLowerInvariant`, `Concat`, `Trim`, `Split`, `ToArray`, `StringBuilder`. It's also allocating several dictionaries. This is something that'll happen for _every single negotiated request_. I think we need to do some profiling and see how bad it is.

src/Nancy/HttpLink.cs
asbjornu: @khellang Yea, that's kind of why I had the load of constructors @thecodejunkie complained about. :wink: But by reducing the number of types to basically just allow `string` for most parameters, I've managed to get the number of constructors down to 4. I hope that's a fine number. :smile:

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,274 @@
...
+        private readonly Uri targetUri;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="HttpLink" /> class.
+        /// </summary>
+        /// <param name="targetUri">The target URI of the link.</param>
+        /// <param name="relation">The relation that identifies the semantics of the link.</param>
+        /// <exception cref="System.ArgumentNullException"><paramref name="targetUri" /> or <paramref name="relation" /></exception>
+        public HttpLink(string targetUri, string relation)
+            : this(ParseUri(targetUri), relation, null, null)
```
khellang: Could `type` and `title` be `string.Empty` instead? I'm generally not a fan of keeping `null` around :smile: Same goes for the other constructors. Might also be worth doing a null check before the values are used, just in case someone tries to pass `null`.

asbjornu:
@khellang We should be talking about very small strings here, but I get your point. If you have any guidance on how I can improve the code, I'll get right to it. I'm no performance expert, though, so I'm not sure what to look for when profiling or even how I should profile this.

src/Nancy/HttpLink.cs
asbjornu: Sure, I can change the `null`s to `string.Empty`. But the last constructor does `null` check every parameter, so I'm not sure what more I should do?

khellang: If the last constructor null-checks every parameter, how can you pass `null` here, then? :stuck

khellang:
The more I think about this, the more I'm leaning towards just taking a quick fix for the bugs you mentioned and ask if you'd like to publish this a community package or something instead.
It's quite a bit of code and I, personally, have no intimate knowledge of RFC 5899. I'm not sure I could support it.
I can't speak for the rest of the team, but I know we've taken some over-the-top PRs earlier and ended up not knowing how to properly support it and regretted taking it in the first place, instead of encouraging community packages.

@NancyFx/most-valued-minions @thecodejunkie @grumpydev  What do you think?

asbjornu:
@khellang I completely understand your position, but the fact remains: Nancy is currently producing a `Link` header. The way it produces this header not only violates RFC 5899, but it makes it hard for users of Nancy to add their own `Link` header. To fix the latter problem, I don't see any way around fixing the former.

Just treating existing content in the `Link` header as an opaque `string` and concatenating Nancy's `Link` header content at the end of it, is imho a bad solution and what I would consider a dirty hack. Instead, we can go down the route of providing users of Nancy with a working implementation of RFC 5899 and thus being able to produce a valid `Link` header. That is of course the route I'm voting for.

If you want someone to support this, then look no further. I'd be more than happy to do it. :smile:

Now, to give more weight to the argument of having proper `Link` header support in Nancy; the header is becoming an essential component to HTTP, being used in all sorts of ways to add custom, application-specific metadata to the response. [JSON-LD is using it](https://www.w3.org/TR/json-ld/#interpreting-json-as-json-ld), it's being used as a generic way to add [profiles](http://words.steveklabnik.com/the-profile-link-relation-and-you) to your application, to name a few.

The usage of the `Link` header will only increase in the years to come and not treating it as a first class citizen in an HTTP server framework will, if my guess is right, prove to be a mistake.

grumpydev:
I can see how it overwriting existing link headers is a bad thing, but I'm struggling to see how the link headers we already add are in violation of the spec.

khellang:
> I completely understand your position, but the fact remains: Nancy is currently producing a Link header. The way it produces this header not only violates RFC 5899, but it makes it hard for users of Nancy to add their own Link header. To fix the latter problem, I don't see any way around fixing the former.

That's why we suggested to
1. Don't add a `Link` header if one has already been added by the user.
2. If a `Link` header is absent, add a properly formatted (according to RFC 5899) header.

That's all we need, really.

If you want to publish a package that include these types, and optionally a custom response negotiator that undestands them, that's cool as well :smile:

asbjornu:
@grumpydev:
>I'm struggling to see how the link headers we already add are in violation of the spec.

The current `Link` header has `rel="application/json"`. This is an invalid `rel`. What should have been output is `rel="alternate"; type="application/json"`. While that's an easy fix to do in the current `string.Format()`, it's not that easy if you want to preserve an existing `Link` header all the while ensuring an RFC 5899 compliant output. To do this, I believe you have to ask yourself: Does the existing `Link` header;

1. already contain `rel="alternate"` with the same `type` parameter, referencing same Target URI?
2. end with a `,`?
3. comply to RFC 5899 at all, or is it just a load of gibberish?

There's probably more questions I can't think of on an early Monday morning, but I believe these are important enough to warrant proper RFC 5899 parsing of the header. How else are you going to figure out the answers to the above questions?

@khellang:
> Don't add a `Link` header if one has already been added by the user.

So if there's an existing `Link` header, no `rel="alternate"` links should be added? Sound weird to me. What the existing `Link` header is for is something Nancy knows nothing about. It could be a `rel="profile"`, `rel="http://www.w3.org/ns/json-ld#context"` or any of the [80 registered link relations](http://www.iana.org/assignments/link-relations/link-relations.xhtml).

> If a `Link` header is absent, add a properly formatted (according to RFC 5899) header.

Yes, had the fix been this simple, I'd agree. But I don't think we can ignore the semantics or format of the existing `Link` header.

khellang:
> So if there's an existing `Link` header, no `rel="alternate"` links should be added? Sound weird to me. What the existing `Link` header is for is something Nancy knows nothing about. It could be a `rel="profile"`, `rel="http://www.w3.org/ns/json-ld#context"` or any of the [80 registered link relations](http://www.iana.org/assignments/link-relations/link-relations.xhtml).

Yes. Exactly. If you've set a `Link` header, you've opted out of the response negotiator setting the `rel="alternate"` values for you. You've effectively said; "I'm controlling this `Link` header now" :smile:

> But I don't think we can ignore the semantics or format of the existing `Link` header.

"If a `Link` header is absent" means there's no semantics of an existing `Link` header to ignore :wink:

asbjornu:
> Yes. Exactly. If you've set a `Link` header, you've opted out of the response negotiator setting the `rel="alternate"` values for you. You've effectively said; "I'm controlling this Link header now"

Say you're dependent on the `rel="alternate"` links Nancy produces. How would you know the target URLs to set for the current resource?

khellang:
I'm also not so sure we can't just append what we need to an existing `Link` header...

> Does the existing Link header;
> already contain `rel="alternate"` with the same `type` parameter, referencing same Target URI?

This is a really strange scenario. If this would be the case, you'd be better off with my proposed solution, which is just fully opting out of setting the header if one is present. How would your alternative solve this? What's the performance overhead of doing all these checks?

> end with a `,`?

This would be easy to check. If it doesn't end with a `,`, we could add one and append our values, no?

> comply to RFC 5899 at all, or is it just a load of gibberish?

Do we care? We're not trying to force our users to "comply" with random RFCs. If a dev wants to use the `Link` header for some "gibberish", why can't they?

khellang:
> Say you're dependent on the `rel="alternate"` links Nancy produces. How would you know the target URLs to set for the current resource?

If you're dependent on the `rel="alternate"` links Nancy produces, then don't explicitly set the `Link` header? Otherwise, you're controlling it yourself?

The problem I have is that we're forcing each and every user of Nancy to conform to this RFC. Not only that, we're sacrificing performance in the process.

Remember; this is the defaults. If people want to write their own negotiator, or derive from the default and customize it, they're free to do so. In Nancy, we've gone pretty far to make sure everything is composable and easy to swap out. We have to find a good balance for users on the SDHP, where everything Just Works™ and the advanced users. _I_ think that an explicit opt-out is the way to go here. When setting the `Link` header in "user-land", nothing unexpected happens, while the users on the SDHP gets the `rel="alternate"` links "for free".

thecodejunkie:
Given the following two facts

1. The vast majority of Nancy users have zero or no interest of any of this
2. The current implementation has been in place for years without any bug reports etc

What arguments are there against not putting this in the core, and instead leave it as an opt-in package? Like @khellang said, this is going to have a performance impact on each and every one of the negotiated responses that Nancy produces. We need to make sure that the SDHP is jus that - sensible defaults that _can_ be replaces when you need something more specific / elaborate. 

There is a lot of string manipulation and LINQ allocations going on in that code, on what is a hot path in Nancy. 

Yes, there could be some changes made here in the core, but perhaps not the full swiss army knife? Maybe the changes that goes into core is a more light weight alternative, and then you maintain this as a community package for when you need some more heavy lifting?

asbjornu:
>> Does the existing Link header; already contain `rel="alternate"` with the same `type` parameter, referencing same Target URI?
> This is a really strange scenario. If this would be the case, you'd be better off with my proposed solution, which is just fully opting out of setting the header if one is present.

I respectfully disagree.

> How would your alternative solve this?

My solution parses the content of the `Link` header, produces a dictionary of `HttpLink` instances keyed on the `rel` value and outputs them again when doing `.ToString()` on the `HttpLinkHeader` dictionary.

> What's the performance overhead of doing all these checks?

A dictionary lookup.

>> end with a `,`?
>This would be easy to check. If it doesn't end with a `,`, we could add one and append our values, no?
Yes, all of the things are easy enough to check in isolation, but combine them and you've got yourself a pretty nasty, bug-infested ball of hair.

>> comply to RFC 5899 at all, or is it just a load of gibberish?
> Do we care?

I certainly hope so!

> We're not trying to force our users to "comply" with random RFCs.

So compliance with [RFC 2616](https://tools.ietf.org/html/rfc2616) (now obsoleted by [RFC 7230](https://tools.ietf.org/html/rfc7230), [RFC 7231](https://tools.ietf.org/html/rfc7231), [RFC 7232](https://tools.ietf.org/html/rfc7232), [RFC 7233](https://tools.ietf.org/html/rfc7233) [RFC 7234](https://tools.ietf.org/html/rfc7234) and [RFC 7235](https://tools.ietf.org/html/rfc7235)) is not important for an HTTP Server Framework?

[These "random" RFCs](https://en.wikipedia.org/wiki/List_of_RFCs) are what makes the internet work. They specify things such as the already mentioned HTTP, [TCP](https://www.ietf.org/rfc/rfc793.txt), [IP](https://tools.ietf.org/html/rfc791), [Media Types](https://tools.ietf.org/html/rfc6838), [URI](https://tools.ietf.org/html/rfc3986), etc.

The two latter RFC's (6838 and 3986) are already implemented in Nancy's `MediaRange` and `Url` classes. It is therefore not unprecedented, not even in Nancy, to implement an RFC in one or more support classes.

The reason I'm proposing that Nancy should implement another "random" RFC, is because just like with other basic building blocks of HTTP (such as Media Types and URI), the `Link` header is (becoming) essential to the web and how people build applications with it.

> If a dev wants to use the Link header for some "gibberish", why can't they?

If a dev wants to produce a URI or Media Type that is just gibberish, should they not at least meet a minimum of resistance from the framework? I believe Nancy provides that resistance in order to comply to RFC 6838 and RFC 3986 already, which is good. Which is why I think it should do the same for the `Link` header.

asbjornu:
>What arguments are there against not putting this in the core, and instead leave it as an opt-in package?

Because Nancy is already outputting a `Link` header.

> There is a lot of string manipulation and LINQ allocations going on in that code, on what is a hot path in Nancy.

That can be fixed. I can get to work on that if you want this in.

> Yes, there could be some changes made here in the core, but perhaps not the full swiss army knife?

It's not a full swiss army knife. It's a **very** focused implementation of a rather small RFC.

> Maybe the changes that goes into core is a more light weight alternative, and then you maintain this as a community package for when you need some more heavy lifting?

I disagree that this is "heavy lifting". It's just four small classes.

thecodejunkie:
> So compliance with RFC 2616 (now obsoleted by RFC 7230, RFC 7231, RFC 7232, RFC 7233 RFC 7234 and RFC 7235) is not important for an HTTP Server Framework?

This is where `opt-in` would enter the picture. If and when you need this, then you could pull in your package and get all the bells and whistles to make sure the links you produce are compliance. There a a billion specifications, RFCs and what not that all make varying sense for people to be compliant with, while to others it doesn't matter at all. Take FIPS compliance for instance, it's important to a few but not the many - there's always going to be trade offs.

Like I said, this is a hot path and adding complexity to that, for each request, is not something that sounds like the best of ideas.

Nobody is arguing against the value of being able to do this, the argument is "should this be in the core or should this be an opt-in thing"

khellang:
> A dictionary lookup.

Let me just quote what you said above:

> My solution parses the content of the `Link` header, produces a dictionary of `HttpLink` instances keyed on the `rel` value and outputs them again when doing `.ToString()` on the `HttpLinkHeader` dictionary.

That sounds like it's doing an awful lot more than just "a dictionary lookup" :stuck_out_tongue_closed_eyes: 

> So compliance with [RFC 2616](https://tools.ietf.org/html/rfc2616) (now obsoleted by [RFC 7230](https://tools.ietf.org/html/rfc7230), [RFC 7231](https://tools.ietf.org/html/rfc7231), [RFC 7232](https://tools.ietf.org/html/rfc7232), [RFC 7233](https://tools.ietf.org/html/rfc7233) [RFC 7234](https://tools.ietf.org/html/rfc7234) and [RFC 7235](https://tools.ietf.org/html/rfc7235)) is not important for an HTTP Server Framework?

> [These "random" RFCs](https://en.wikipedia.org/wiki/List_of_RFCs) are what makes the internet work. They specify things such as the already mentioned HTTP, [TCP](https://www.ietf.org/rfc/rfc793.txt), [IP](https://tools.ietf.org/html/rfc791), [Media Types](https://tools.ietf.org/html/rfc6838), [URI](https://tools.ietf.org/html/rfc3986), etc.

Now you're just being silly. `Link` headers are not what makes the internet work, and people are already using them in many different ways. The RFC is just one "standard". Why break them by validating that they conform to an RFC?

> The two latter RFC's (6838 and 3986) are already implemented in Nancy's `MediaRange` and `Url` classes. It is therefore not unprecedented, not even in Nancy, to implement an RFC in one or more support classes.

> The reason I'm proposing that Nancy should implement another "random" RFC, is because just like with other basic building blocks of HTTP (such as Media Types and URI), the `Link` header is (becoming) essential to the web and how people build applications with it.

You are correct that we have types for this in Nancy already. What you've failed to notice is that we're not *forcing* anyone to use them. If someone wanted to set the `Content-Type` header to "hello my name is Alice", they could. I think we'd be fine with having a "builder" for RFC compliant `Link` header values in Nancy, but it's way too much to parse existing headers, go though lots of logic in order to "merge" our values with existing values etc. for *each request*.

khellang:
Just to sum up here...

  1. Yes, we want to fix the bug where Nancy always overrides the `Link` header during negotiation.
  2. No. we don't want to bring in the overhead and break people by trying to parse existing `Link` header values and forcing them to conform to an RFC.
  3. Yes, we'd be fine with having a `Link` class (or similar) to build RFC compliant `Link` header values, which in turn could be used in the negotiator. This should probably be a different PR.

Anything more than that would have to be shipped as a community maintained package :smile:

There is one question that needs to be answered: Do we simply noop if there's already a `Link` header present, or do we append our values, regardless? I'm fine with either.

asbjornu:
> 1. Yes, we want to fix the bug where Nancy always overrides the Link header during negotiation.
> 2. No. we don't want to bring in the overhead and break people by trying to parse existing Link header values and forcing them to conform to an RFC.
> 3. Yes, we'd be fine with having a Link class (or similar) to build RFC compliant Link header values, which in turn could be used in the negotiator. This should probably be a different PR.

Ok, I'm fine with doing the builder approach instead. I'll rebase and squash everything into a new PR and fix list item no. 1 by repurposing this PR.

> There is one question that needs to be answered: Do we simply noop if there's already a Link header present, or do we append our values, regardless? I'm fine with either.

We can do an (imho, **very**) naïve approach and checking if the existing `Link` value ends with a comma and append our links at the end of it. Or perhaps even simpler; append the existing value at the end of Nancy's `alternate` links, separated by a comma.

khellang:
> Or perhaps even simpler; append the existing value at the end of Nancy's alternate links, separated by a comma.

Yes, I think that would be the best solution. In that way, if the user wants "gibberish" `Link` headers, they'll get what they want, and then some :smile:

asbjornu:
@khellang Rebased, repurposed, rewritten, re-squashed, re-tested and re-pushed. I hope it satisfied the "good enough" criteria. :smile:

khellang:
@asbjornu Looks much better! :heart: Eagerly awaiting the `Link` builder :grin:

src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
```diff
@@ -328,16 +328,18 @@ private static NegotiationContext GetNegotiationContext(object routeResult, Nanc
...
-        private static string CreateLinkHeader(Url requestUrl, IEnumerable<KeyValuePair<string, MediaRange>> linkProcessors)
+        protected virtual string CreateLinkHeader(Url requestUrl, IEnumerable<KeyValuePair<string, MediaRange>> linkProcessors, string existingLinkHeader)
         {
             var fileName = Path.GetFileNameWithoutExtension(requestUrl.Path);
             var baseUrl = string.Concat(requestUrl.BasePath, "/", fileName);
 
             var links = linkProcessors
-                .Select(lp => string.Format("<{0}.{1}>; rel=\"{2}\"", baseUrl, lp.Key, lp.Value));
+                .Select(lp => string.Format("<{0}.{1}>; rel=\"alternate\"; type=\"{2}\"", baseUrl, lp.Key, lp.Value))
+                .Concat(new[] { existingLinkHeader });
```
khellang: What happens if `existingLinkHeader` is `null`? Will the `string.Join` call below append an extra `,`?

khellang: Hmm. I just did a quick test:

```csharp
Console.WriteLine(string.Join(", ", new[] { "hello", "testing", null }));
```

Result: `"hello, testing, "`

That's not good. I think we need to check for `null` before concatenating here.

asbjornu: Good point. Fixed.

src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
```diff
@@ -286,15 +285,16 @@ private static NegotiationContext GetNegotiationContext(object routeResult, Nanc
...
-            IEnumerable<CompatibleHeader> compatibleHeaders,
-            Response response,
-            Url requestUrl)
+        private void AddLinkHeader(IEnumerable<CompatibleHeader> compatibleHeaders, Response response, Url requestUrl)
         {
             var linkProcessors = GetLinkProcessors(compatibleHeaders, response.ContentType);
             if (linkProcessors.Any())
             {
-                response.Headers["Link"] = CreateLinkHeader(requestUrl, linkProcessors);
+                var existingLinkHeader = response.Headers.ContainsKey("Link")
```
khellang: Also, can we make this a `TryGetValue`?

asbjornu: Sure.

asbjornu:
Typical. Just as I was going to squash and force-push this branch again now, my Windows VM broke down. With only half an hour left at work before I leave for easter holiday, there's no chance I'll be able to finish this, nor the `Link` builder before easter. I hope that doesn't ruin the chances of this making it into 2.0? :grin:

asbjornu:
@khellang, There we go. Your comments have now been taken care of. I think this should be ready for merge now. On to the `Link` builder in a separate PR. :smile:

khellang:
Thanks for this @asbjornu! :heart:

asbjornu:
@khellang Brilliant, thanks!

PR #2370, opened 2016-03-17
RFC 5899 Compliant Link Header (1.x)

asbjornu:
Please see #2369 for an explanation of the value and content of this PR. This is in essence the same PR, just rebased on top of `1.x-WorkingBranch`.

The reason I believe this should be merged and warrants a new release of `1.x` is that the current implementation of `DefaultResponseNegotiator.AddLinkHeader()` doesn't check if a `Link` header already exists, effectively replacing any existing Link headers with its own.

I'm using Nancy 1.x in a project and the `Link` header replacement is what made me investigate this problem and start a PR to begin with. I would highly appreciate a merge of this and new release of 1.x so this bug is fixed, because it allows me to move forward with my project without having to upgrade it to Nancy 2.0.

Thanks! :smiley:


asbjornu:
TeamCity seems to be configured to build `master` and thus expects a different folder layout and placement of the `.sln` file. Not much I can do about that. I don't understand why Travis is failing, but it might be as with #2369 and the use of C# 6 features?

khellang:
We won't be pushing more 1.x releases unless there are some really serious bugs, as I've [mentioned before](https://github.com/NancyFx/Nancy/commit/3d81957aef70d747e7703c418f12b332b9f356dc#commitcomment-16702980). I don't think this warrants a new 1.x release, given we're days/hours away from shipping the first 2.0 pre-release. This behavior has been there since the dawn of time without a lot of people complaining, and this change is a bit too big to review and pull for a patch release. Sorry :disappointed:

asbjornu:
@khellang I'm sad about that decision since this is obviously a bug and it means I have to wait for 2.0 to become stable as well as rework my code to be 2.0 compatible in order to move forward with my project, which was going to have a release in April until we discovered this `Link` header problem. We're pretty much guaranteed to miss that release now. :disappointed:

thecodejunkie:
@asbjornu First of all, thank you for your work on this! 

Now to the more pressing concern. There is _absolutely_ no reason for you to miss your deadline, with the bug fix in place. At least not because of Nancy. Had Nancy been a closed source project then, yes, I would have totally agreed with you. The fact that Nancy is open-source means the direct opposite. Your company can, safely, fork the [1.x-WorkingBranch](https://github.com/NancyFx/Nancy/tree/1.x-WorkingBranch), apply the fix, use the provided `nuspec` to build a Nuget (can even do `rake nuget_package` if you want) and _confidently_ use that version.

The fact that Nancy `1.4.x` was officially but in [maintenance mode 5 months ago](http://thecodejunkie.com/2015/11/08/our-plans-for-nancy-1-x-and-beyond/) only means that your work load, for maintaining a fork, is virtually going to be zero work. We will only put out a new `1.4.x` release if a _critical_ bug appears and if we do, it's very unlikely that it would overlap this code, so it would be trivial for you to rebase your fork and just package up a new Nuget.

We're talking about less than 30 minutes of work to setup your own fork with the fix. If you have CI server (like TeamCity) then you can even set it up to act as a private nuget feed.

Hopefully this should put your worries a side and please let us know if we can assist you in any way to accomplish this.

Best of luck with your release

jchannon:
Could you not create an extension method and use it in the AfterRequest
pipeline to handle the link header? Just an idea.

grumpydev:
Or even replace our implementation with your own using the container.

asbjornu:
@thecodejunkie Good points wrt. Nancy being open source and maintaining our own fork. That is indeed an option, although not one I would prefer.

@jchannon Thanks for the tip, I'll definitely investigate that option.

@grumpydev How would I do that, exactly? An example would be highly appreciated!

:smiley:

thecodejunkie:
@asbjornu with regards to @grumpydev 's suggestion. You can replace this https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Bootstrapper/NancyInternalConfiguration.cs#L60 with your own implementation and Nancy will use that at runtime.

Simply override the `InternalConfiguration` property, of your bootstrapper and return something like

```c#
return NancyInternalConfiguration.WithOverrides(x => x.ResponseNegotiator = typeof(MyAwesomeImplementation));
```

asbjornu:
@thecodejunkie Brilliant, thanks!

PR #2378, opened 2016-03-22
Don't add any blank lines between using groups

asbjornu:
As it has been pointed out as unwanted whitespace in code reviews, I think it's good if the R# settings don't add blank lines between `using` groups.


khellang:
Thanks! :heart:

PR #2381, opened 2016-03-24
Re-implemented TryCastResultToResponse to boost performance

thecodejunkie:
The current implementation of `DefaultResponseNegotiator.TryCastResultToResponse` uses exception handling to control the flow. This is (no surprise) slow, but as it turns out it was having a much bigger impact that I had anticipated. The new implementation uses reflection to figure out, up front, if the cast is going to be valid or not. Due to reflection caching in .NET 4.0+ there is really no need to do any caching on our own and the performance gain is amazing.

All requests, that ended up being negotiated (i.e not directly castable to `Response`) would hit this code path and had to take the penalty

This is part of a 100k iteration sample, compared before and after

BEFORE
image

AFTER
image

Perhaps we can squeeze in additional tweaks?

### DIFF
--- a/src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
+++ b/src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
+    using System.Reflection;
     using System.Text;
 
     using Nancy.Conventions;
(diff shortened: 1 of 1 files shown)


grumpydev:
:+1:

phillip-haydon:
Splooge

khellang:
The only thing that worries me is that this might've regressed on Mono...

phillip-haydon:
How so?

khellang:
@jchannon has been complaining a lot about reflection perf on Mono, I guess he should check :)

jchannon:
Only way to check it to pull Andreas branch and console app that does 100k
requests and run it on a mono box

PR #2385, opened 2016-03-28
Altered datetime serialisation tests to cope with offsets better

jchannon:
With the Easter 2016 clocks going forward the serialisation tests for `DateTime` failed.

The serialiser converts `Unspecified` dates to Local and then formats using the `ToString("o")` formatter.

This will append the timezone differences if a date is Local.

This PR uses the [recommended](http://stackoverflow.com/a/17974704/84539) `TimeZoneInfo.Local.GetUtcOffset(myDate);` to find the timezone offset.  We then use this in the tests to assert the value the serialiser's `ToString("o")` formatter returns with the datetime passed into the SUT.

### DIFF
--- a/test/Nancy.Tests/Unit/Json/Simple/NancySerializationStrategyFixture.cs
+++ b/test/Nancy.Tests/Unit/Json/Simple/NancySerializationStrategyFixture.cs
@@ -44,13 +44,13 @@ namespace Nancy.Tests.Unit.Json.Simple
         {
             // Given
             var strategy = CreateStrategy();
-            strategy.RegisterConverters(new [] { new DateTimeJavaScriptConverter() });
+            strategy.RegisterConverters(new[] { new DateTimeJavaScriptConverter() });
             var expectedValue = new DateTime(2016, 2, 27);
             object serialized;
 
             // When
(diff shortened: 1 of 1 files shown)


xt0rted:
When I run these tests (en-US & -4 eastern daylight)  I'm getting two failing at [NancySerializationStrategyFixture.cs#L80-112](https://github.com/NancyFx/Nancy/blob/2c07700b2d999d20c60f524480c48b2dc19b678c/test/Nancy.Tests/Unit/Json/Simple/NancySerializationStrategyFixture.cs#L80-112)

**Should_serialize_unspecified_datetime_object**
```
Expected: 2014-03-09T17:03:25.2340000+-04:00
Actual:   2014-03-09T17:03:25.2340000-04:00
```

**Should_serialize_local_datetime_object**
```
Expected: 2014-03-09T17:03:25.2340000+-04:00
Actual:   2014-03-09T17:03:25.2340000-04:00
```

jchannon:
Those expected values look,um, interesting :)

jchannon:
Need to test those when #2427 is complete

xt0rted:
The issue above has been fixed in #2525

PR #2389, opened 2016-03-29
RFC 5988 HTTP Link builder

asbjornu:
As discussed in #2369, here's a pull request for the `HttpLinkBuilder` class.


src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        /// <summary>
+        /// Indicates whether the current object is equal to another object of the same type.
+        /// </summary>
+        /// <param name="other">An object to compare with this object.</param>
+        /// <returns>
+        /// true if the current object is equal to the <paramref name="other" /> parameter; otherwise, false.
+        /// </returns>
+        bool IEquatable<HttpLink>.Equals(HttpLink other)
```
khellang: Why is this implemented explicitly?

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,262 @@
...
+        /// A <see cref="System.String"/> that represents this instance.
+        /// </returns>
+        public override string ToString()
+        {
+            var linkBuilder = new StringBuilder();
+            linkBuilder.Append('<');
+            linkBuilder.Append(this.TargetUri);
+            linkBuilder.Append('>');
+
+            foreach (var parameter in this.OrderBy(parameter => parameter.Key, new LinkParameterComparer()))
```
khellang: Can we cache this comparer somewhere?

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,262 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+    using Nancy.Responses.Negotiation;
+
+    /// <summary>
+    /// Represents one of possibly many RFC 5899 HTTP Links contained in a <see cref="HttpLinkBuilder"/>.
+    /// </summary>
+    public class HttpLink : Dictionary<string, object>, IEquatable<HttpLink>
```
khellang: Why is this deriving from a dictionary instead of keeping a dictionary internally?

src/Nancy/HttpLinkBuilder.cs
```diff
@@ -0,0 +1,49 @@
+namespace Nancy
+{
+    using System.Collections.Generic;
+    using System.Linq;
+
+    /// <summary>
+    /// Builds an RFC 5899 Link HTTP header as a <see cref="List{T}" /> of <see cref="HttpLink" /> objects.
+    /// </summary>
+    /// <seealso cref="HttpLink" />
+    public class HttpLinkBuilder : List<HttpLink>
```
khellang: Same here... does this need all of `List<HttpLink>`'s members?

src/Nancy/HttpLinkBuilder.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Returns a <see cref="System.String" /> that represents an RFC 5899 Link HTTP header.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="System.String" /> that represents an RFC 5899 Link HTTP header.
+        /// </returns>
+        public override string ToString()
+        {
+            var links = this.Select(link => link.ToString());
```
khellang: Can we use a `StringBuilder` in this method and just do a regular `foreach`?

src/Nancy/HttpLinkRelation.cs
```diff
@@ -0,0 +1,180 @@
...
+    using System;
+    using System.Runtime.Serialization;
+
+    /// <summary>
+    /// The relation that identifies the semantics of a <see cref="HttpLink"/> contained in an RFC 5899 Link HTTP header,
+    /// as built by the <see cref="HttpLinkBuilder"/>.
+    /// </summary>
+    /// <seealso cref="System.Uri" />
+    [Serializable]
+    public class HttpLinkRelation : Uri, IEquatable<HttpLinkRelation>
```
khellang: Same here... Not sure we should derive from `Uri` :confused:

src/Nancy/HttpLinkRelation.cs
```diff
@@ -0,0 +1,180 @@
...
+            this.prefix = prefix;
+            this.value = value;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="HttpLinkRelation"/> class.
+        /// </summary>
+        /// <param name="serializationInfo">An instance of the <see cref="T:System.Runtime.Serialization.SerializationInfo" /> class containing the information required to serialize the new <see cref="T:System.Uri" /> instance.</param>
+        /// <param name="streamingContext">An instance of the <see cref="T:System.Runtime.Serialization.StreamingContext" /> class containing the source of the serialized stream associated with the new <see cref="T:System.Uri" /> instance.</param>
+        protected HttpLinkRelation(SerializationInfo serializationInfo, StreamingContext streamingContext)
```
khellang: This (and the `[Serializable]` attribute above) won't fly on .NET Core. I'm not really sure if we should keep this for now or just kill it sooner rather than later.

src/Nancy/HttpLink.cs
asbjornu: @khellang It's just a personal preference to not have several overloads publicly available that do the exact same thing.

src/Nancy/HttpLink.cs
asbjornu: Sure, I'll add it.

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,262 @@
...
+
+        /// <summary>
+        /// Returns a hash code for this instance.
+        /// </summary>
+        /// <returns>
+        /// A hash code for this instance, suitable for use in hashing algorithms and data structures like a hash table.
+        /// </returns>
+        public override int GetHashCode()
+        {
+            return this.ToString().ToLowerInvariant().GetHashCode();
```
khellang: `StringComparer.OrdinalIgnoreCase.GetHashCode(this.ToString())`?

src/Nancy/HttpLink.cs
asbjornu: It's an easy (but perhaps hard to unintuitive, now that you mention it) way to expose the link's parameters. Perhaps I should move it into a `Parameters` property?

khellang:
What happened to `ContentNegotiationFixture`?

> 924 additions, 917 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

src/Nancy/HttpLinkBuilder.cs
asbjornu: Here I think it most definitely make sense. A link builder (and the header it's outputting to) **is** an enumeration of links. I think this is very intuitive and makes the builder easy to understand and use, since it implements a well-known interface.

src/Nancy/HttpLinkBuilder.cs
asbjornu: Absolutely.

src/Nancy/HttpLink.cs
khellang: Yes, I think that makes more sense :smile:

src/Nancy/HttpLinkRelation.cs
asbjornu: Well... According to RFC 5899, it **is** an URI (or an IRI, to be exact), so I thought it made sense to inherit to make that relationship obvious.

src/Nancy/HttpLink.cs
khellang: But this is the one you'd most likely want to call?

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,262 @@
...
+                linkBuilder.Append('=');
+                linkBuilder.Append('"');
+                linkBuilder.Append(parameter.Value);
+                linkBuilder.Append('"');
+            }
+
+            return linkBuilder.ToString();
+        }
+
+        private T GetParamValue<T>(string paramName)
```
khellang: Unless "param" is something that's specific to the RFC, I think we should rename to "parameter". We generally try to avoid that kind of shortening. More descriptive == better :smile:

src/Nancy/HttpLinkRelation.cs
asbjornu: Ok, good to know. I just added it because `System.Uri` had it and that it's been a good best practice to add these serializer constructors for ye olden runtimes. I have no preference either way and can remove it without hesitation.

src/Nancy/HttpLink.cs
asbjornu: Yes, absolutely.

asbjornu:
@khellang Yowza. That's a great question. I'll investigate!

src/Nancy/HttpLinkBuilder.cs
khellang: > A link builder (and the header it's outputting to) is an enumeration of links.

You don't need to inherit `List<T>` to implement `IEnumerable<T>` :wink:

src/Nancy/HttpLink.cs
asbjornu: @khellang It's the name used in the [ABNF of RFC 5899](https://tools.ietf.org/html/rfc5988#section-5) and a remnant from when I also had a parser in the file that actually cared about the ABNF syntax. Now I don't, so I don't mind renaming it to `Parameter`. I agree that more descriptive is better for sure! :smiley:

src/Nancy/HttpLinkRelation.cs
khellang: I think we can just keep it for now, and #ifdef it on the coreclr branch :smile:

src/Nancy/HttpLink.cs
asbjornu: Yes, but I can't hide `Object.Equals()` in any way, so then we'll end up with two. I'm not religious about it, I just find an API easier to use if there's less choice.

src/Nancy/HttpLinkBuilder.cs
asbjornu: No, but `List<T>` has a nice `.Add()` method that makes a lot of sense to a `Builder` class, don't you agree?

src/Nancy/HttpLinkRelation.cs
asbjornu: :+1:

asbjornu:
@khellang Thanks for the review! Besides the discussions we haven't reached a conclusion on yet (see above), I think everything else should now be in shape. The problem with `ContentNegotiationFixture` was line-ending changes that had sneakily snuck in undetected. If you think everything now looks good, I'll squash all the code review commits and force-push something that should be ready for merging.

asbjornu:
Also, 136283ef6c314aa622f702f91d6b24e51aa7ee96 seems to be related to #2390, so if that is merged first, I can try to tackle the `TODO` in said commit.

src/Nancy/HttpLinkBuilder.cs
khellang: Again, yes, I agree, but it also provides a lot of other methods that might not make sense. We shouldn't inherit from `List<T>` just because of laziness :smile:

src/Nancy/HttpLinkRelation.cs
khellang: If it is a `Uri`, do we need a separate type for it? As far as I understand, it's basically a `Uri` consisting of a base and relative `Uri`? I think the `Uri` parsing code is a lot more efficient than what's in this type...

src/Nancy/HttpLinkBuilder.cs
asbjornu: True, but I don't know of an interface that is a middle-ground between `IList<T>` and `IEnumerable<T>`. While the other methods and properties of `List<T>` aren't currently in use by Nancy itself, who knows what consumers might want to do with the `HttpLinkBuilder`? They might find the indexer useful, for instance. If we shouldn't cater to thought use cases, I can of course instead implement `IEnumerable<T>` and add an interface-free `Add()` method, if you think that's cleaner?

src/Nancy/HttpLinkRelation.cs
asbjornu: It's a conditionally relative or absolute `Uri`. That's what the code in `Parse()` deals with. According to RFC 5899, all `rel` values not [registered in IANA](http://www.iana.org/assignments/link-relations/link-relations.xhtml), should be prefixed with a base URI, while `rel` values registered in IANA should be unprefixed, opaque strings.

asbjornu:
Since #2390 is now merged, I've rebased on top of 1677b5fbaf6d171c5c124c6b40964d4c78985af0 and squashed the PR into 3 commits. I wish I knew why AppVeyor is failing the build, but I don't understand how what it complains about is related to any of the code I've touched. Ideas, @khellang?

xt0rted:
The build is failing because of the use of `SerializableAttribute` and `SerializationInfo` as @khellang mentioned previously.

https://ci.appveyor.com/project/khellang/nancy/build/2.0.0-ci000182-cfyeeuok#L359
```
=== TEST EXECUTION SUMMARY ===
   Nancy.Tests  Total: 1838, Errors: 0, Failed: 0, Skipped: 0, Time: 26.105s
SUMMARY: Total: 1 targets, Passed: 1, Failed: 0.
Project Nancy (.NETStandard,Version=v1.5) will be compiled because expected outputs are missing
Compiling Nancy for .NETStandard,Version=v1.5
dotnet.exe : C:\projects\nancy\.dotnet\win7-x64\dotnet.exe compile-csc @C:\projects\nancy\src\Nancy\obj\Release\netstandard1.5\dotnet-compile.rsp returned Exit Code 1
At C:\projects\nancy\build.ps1:53 char:5
+     & dotnet test -c Release;
+     ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (C:\projects\nan...ned Exit Code 1:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
C:\projects\nancy\src\Nancy\HttpLinkRelation.cs(11,6): error CS0246: The type or namespace name 'SerializableAttribute' could not be found (are you missing a using directive or an assembly reference?)
 
 
C:\projects\nancy\src\Nancy\HttpLinkRelation.cs(11,6): error CS0246: The type or namespace name 'Serializable' could not be found (are you missing a using directive or an assembly reference?)
 
 
C:\projects\nancy\src\Nancy\HttpLinkRelation.cs(49,36): error CS0246: The type or namespace name 'SerializationInfo' could not be found (are you missing a using directive or an assembly reference?)
 
 
 
Compilation failed.
    0 Warning(s)
    3 Error(s)
```

khellang:
> I think we can just keep it for now, and #ifdef it on the coreclr branch

Yeah. I think that time has come. The coreclr branch has been merged into master now.

asbjornu:
@xt0rted Thanks! @khellang All green. Any more comments about this, or is it good to go?

src/Nancy/HttpLink.cs
khellang: By hiding this overload, it means that everyone that calls `Equals` will end up with the `object` overload and do extra checks, unless you reference a link as `IEquatable<HttpLink>`, which I doubt anyone would do.

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,283 @@
...
+            {
+                throw new ArgumentNullException("targetUri");
+            }
+
+            if (relation == null)
+            {
+                throw new ArgumentNullException("relation");
+            }
+
+            this.parameters = new Dictionary<string, object>
```
khellang: Should this dictionary be case-insensitive?

src/Nancy/HttpLink.cs
asbjornu: Ok, fair enough. I'll change it.

src/Nancy/HttpLink.cs
asbjornu: Good question. mnot/I-D@825bf92424a977a040f69f2945b9f1d4a24c9504 attempts to clarify this a bit, by basically saying nothing. RFC 5988 itself defines all attributes as case insensitive and singular, though (see mnot/I-D#170 for clearer language on this), and I believe that to be the case with most other applications of the `Link` header as well. 

So, the most "correct" implementation would probably be something like `IEnumerable<KeyValuePair<string, string>>`, but the more pragmatic and usable thing would be to initialize a `Dictionary<string, string>` with `StringComparer.InvariantCultureIgnoreCase`.

khellang: > So, the most "correct" implementation would probably be something like IEnumerable<KeyValuePair<string, string>>

Are you saying that ideally you'd like a `MultiValueDictionary<string, string>`?

> the more pragmatic and usable thing would be to initialize a `Dictionary<string, string>` with `StringComparer.InvariantCultureIgnoreCase`.

Why `StringComparer.InvariantCultureIgnoreCase`? Wouldn't `StringComparer.OrdinalIgnoreCase` be "safer"?

asbjornu: @khellang Yes, absolutely. Just a typo, thanks for noticing!

asbjornu:
@khellang I think I'm done with the review (for now). Anything else I need to do? :smiley:

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,283 @@
...
+    using System.Linq;
+    using System.Text;
+    using Nancy.Responses.Negotiation;
+
+    /// <summary>
+    /// Represents one of possibly many RFC 5988 HTTP Links contained in a <see cref="HttpLinkBuilder"/>.
+    /// </summary>
+    public class HttpLink : IEquatable<HttpLink>
+    {
+        private static readonly HttpLinkParameterComparer parameterComparer = new HttpLinkParameterComparer();
```
khellang: We'll probably just let this slip, but I'm curious... Didn't R# suggest this name to be `ParameterComparer`? If not, we might look into the DotSettings here.

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,283 @@
...
+        /// <summary>
+        /// Determines whether the specified <see cref="System.Object" />, is equal to this instance.
+        /// </summary>
+        /// <param name="other">The <see cref="System.Object" /> to compare with this instance.</param>
+        /// <returns>
+        ///   <c>true</c> if the specified <see cref="System.Object" /> is equal to this instance; otherwise, <c>false</c>.
+        /// </returns>
+        public override bool Equals(object other)
+        {
+            return ((IEquatable<HttpLink>)this).Equals(other as HttpLink);
```
khellang: Now that the method isn't implemented explicitly anymore, we can remove this horrible cast :wink:

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,283 @@
...
+
+        /// <summary>
+        /// Returns a hash code for this instance.
+        /// </summary>
+        /// <returns>
+        /// A hash code for this instance, suitable for use in hashing algorithms and data structures like a hash table.
+        /// </returns>
+        public override int GetHashCode()
+        {
+            return StringComparer.OrdinalIgnoreCase.GetHashCode(this.ToString());
```
khellang: Just something to think about. Do we really want this to generate a potentially big string each time it needs the hash code? I think that might be breaking one of the "rules" of hash code functions...

src/Nancy/HttpLinkRelation.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <summary>
+        /// Determines whether the specified <see cref="System.Object" />, is equal to this instance.
+        /// </summary>
+        /// <param name="other">The <see cref="System.Object" /> to compare with this instance.</param>
+        /// <returns>
+        ///   <c>true</c> if the specified <see cref="System.Object" /> is equal to this instance; otherwise, <c>false</c>.
+        /// </returns>
+        public override bool Equals(object other)
+        {
+            return ((IEquatable<HttpLinkRelation>)this).Equals(other as HttpLinkRelation);
```
khellang: Same here...

src/Nancy/HttpLinkRelation.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        /// <summary>
+        /// Returns a hash code for this instance.
+        /// </summary>
+        /// <returns>
+        /// A hash code for this instance, suitable for use in hashing algorithms and data structures like a hash table.
+        /// </returns>
+        public override int GetHashCode()
+        {
+            return StringComparer.OrdinalIgnoreCase.GetHashCode(this.ToString());
```
khellang: Same here...

khellang:
@asbjornu Almost there. Just a couple of nits and the `GetHashCode` functions.Think there's any way we can avoid doing all the work/allocations in there?

src/Nancy/HttpLink.cs
asbjornu: It did. Fixed.

src/Nancy/HttpLink.cs
asbjornu: Yea, the `GetHashCode` implementations were rubbish. Redone.

asbjornu:
@khellang There we go. I can squash 5c0b2a7864a7d5437d3c43d6cf5b2a4f55840034 into 847300743c405ba5a7db3fc0c4e66d6b7732231a if you want to, I just thought it might be easier to review as a separate commit.

src/Nancy/HttpLink.cs
```diff
@@ -180,7 +185,22 @@ public bool Equals(HttpLink other)
         /// </returns>
         public override bool Equals(object other)
         {
-            return ((IEquatable<HttpLink>)this).Equals(other as HttpLink);
+            if (ReferenceEquals(null, other))
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(this, other))
```
khellang: Both this and the null check is covered in the other equals method. I think `return other is HttpLink && this.Equals((HttpLink) other)` should be enough here...

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,321 @@
...
+        ///   <c>true</c> if the specified <see cref="System.Object" /> is equal to this instance; otherwise, <c>false</c>.
+        /// </returns>
+        public override bool Equals(object other)
+        {
+            if (ReferenceEquals(null, other))
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(this, other))
```
khellang: Both this and the null check is covered in the other equals method. I think `return other is HttpLink && this.Equals((HttpLink) other)` should be enough here...

src/Nancy/HttpLink.cs
```diff
@@ -0,0 +1,321 @@
...
+        /// <returns>
+        /// A hash code for this instance, suitable for use in hashing algorithms and data structures like a hash table.
+        /// </returns>
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                var p = 0;
+
+                foreach (var parameter in this.parameters)
```
khellang: Doesn't `this.parameters` already have an efficient `GetHashCode` implementation?

src/Nancy/HttpLinkRelation.cs
```diff
@@ -0,0 +1,195 @@
...
+        ///   <c>true</c> if the specified <see cref="System.Object" /> is equal to this instance; otherwise, <c>false</c>.
+        /// </returns>
+        public override bool Equals(object other)
+        {
+            if (ReferenceEquals(null, other))
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(this, other))
```
khellang: Same goes for this.

src/Nancy/HttpLink.cs
asbjornu: I agree.

src/Nancy/HttpLink.cs
asbjornu: It might be efficient, but it doesn't take the keys and values into consideration, especially not in a key-insensitive way for the key.

asbjornu:
@khellang I've changed the `Equals(object)` methods as you suggested, but left the `HttpLink .GetHashCode()` since I think it's valuable to have a hash code that reflects the case insensitivity of the parameter keys.

khellang:
Awesome! Thanks :heart:

asbjornu:
@khellang Yay! Thanks! :heart:

PR #2390, opened 2016-03-29
Optimization of Content Negotiation

thecodejunkie:
This is part of a 100k requests sample

There are a bunch of changes in this one
- `RequestHeaders.ContentType` was changed from `string` to `MediaRange` to reduce the number of `string -> MediaRange` implicit casts that are preformed during route execution
- `RequestHeaders.ContentType` was changed to read-only
- `MediaRange` now implements `IEquatable<MediaRange>`
- Added `MediaRange.Empty`
- Added `MediaRange.IsEmpty` - Together with `Empty` these replaces a couple of `IsNullOrEmpty` checks we performed in places where `contentType` was still a `string`
- Removed parts of the LINQ statements inside the `DefaultResponseNegotiatior` (there are more fish to fry in there, that's for sure)

NOTE: This pull-request does not convert all occurrences where the `contentType` is represented as `string`

Before
image

After
image

Not bad, if I may say so myself :)

### DIFF
--- a/src/Nancy/Request.cs
+++ b/src/Nancy/Request.cs
@@ -234,21 +234,20 @@ namespace Nancy
 
         private void ParseFormData()
         {
-            if (string.IsNullOrEmpty(this.Headers.ContentType))
+            if (this.Headers.ContentType == null)
             {
                 return;
             }
 
-            var contentType = this.Headers["content-type"].First();
-            var mimeType = contentType.Split(';').First();
-            if (mimeType.Equals("application/x-www-form-urlencoded", StringComparison.OrdinalIgnoreCase))
+            var contentType = this.Headers.ContentType;
+            if (contentType.Matches("application/x-www-form-urlencoded"))
             {
                 var reader = new StreamReader(this.Body);
                 this.form = reader.ReadToEnd().AsQueryDictionary();
                 this.Body.Position = 0;
             }
 
-            if (!mimeType.Equals("multipart/form-data", StringComparison.OrdinalIgnoreCase))
+            if (!contentType.Matches("multipart/form-data"))
             {
                 return;
             }
--- a/src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
+++ b/src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
@@ -1,6 +1,7 @@
 ﻿namespace Nancy.Responses.Negotiation
 {
     using System;
+    using System.CodeDom.Compiler;
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
@@ -15,7 +16,7 @@
     /// </summary>
     public class DefaultResponseNegotiator : IResponseNegotiator
     {
-        private readonly IEnumerable<IResponseProcessor> processors;
+        private readonly IReadOnlyCollection<IResponseProcessor> processors;
         private readonly AcceptHeaderCoercionConventions coercionConventions;
 
         /// <summary>
@@ -25,7 +26,7 @@
         /// <param name="coercionConventions">The Accept header coercion conventions.</param>
         public DefaultResponseNegotiator(IEnumerable<IResponseProcessor> processors, AcceptHeaderCoercionConventions coercionConventions)
         {
-            this.processors = processors;
+            this.processors = processors.ToArray();
             this.coercionConventions = coercionConventions;
         }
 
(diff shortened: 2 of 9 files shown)


test/Nancy.Tests/Unit/RequestHeadersFixture.cs
```diff
@@ -744,10 +744,9 @@ public void Should_return_null_date_when_date_headers_are_not_available()
             var headers = new RequestHeaders(rawHeaders);
 
             // Then
-            headers.Date.ShouldBeNull();;
+            headers.Date.ShouldBeNull(); ;
```
xt0rted: extra `;`

khellang:
> `RequestHeaders.ContentType` was changed to read-only

Why was this change made? It seems reasonable to me that request headers are mutable, so middleware (pipeline stuff) can change headers on their way to the final handler.

> `RequestHeaders.ContentType` was changed from `string` to `MediaRange` to reduce the number of `string` -> `MediaRange` implicit casts that are preformed during route execution

Shouldn't it be a `MediaRange?` (nullable)? That way we could remove the `Empty` and `IsEmpty` properties, which IMO, make little sense. Either it's there, or it's not. I don't think an empty `Content-Type` makes sense?

phillip-haydon:
Why would a header change, if the client said "This is `application/json`" then the owin pipeline should capture this and change it before it ever gets to Nancy. Once inside Nancy, the request header should never have to change for `Content-Type`

IMO

phillip-haydon:
I'm OK for this to be merged, but @khellang disagrees because he's wrong. What do we do? :grin:

asbjornu:
As this seems related to 136283ef6c314aa622f702f91d6b24e51aa7ee96, I have an interest in ensuring that nothing I do makes the performance worse. Perhaps adding a [`[Benchmark]` test](http://www.hanselman.com/blog/BenchmarkingNETCode.aspx) in this PR would be a good idea, so I can run it on my branch and compare results?

khellang:
> Why would a header change, if the client said "This is application/json" then the owin pipeline should capture this and change it before it ever gets to Nancy. Once inside Nancy, the request header should never have to change for Content-Type

I agree that it's a it's a weird case in this scenario, but I'd like the request headers to be mutable in principle. I don't think it's up to us to decide what can be changed and what can't.

src/Nancy/Responses/Negotiation/DefaultResponseNegotiator.cs
```diff
@@ -357,15 +362,29 @@ protected virtual string CreateLinkHeader(Url requestUrl, IEnumerable<KeyValuePa
...
+            var result = string.Empty;
+
+            foreach (var linkProcessor in linkProcessors)
+            {
+                if (result.Length > 0)
+                {
+                    result += ", ";
+                }
+
+                result += string.Concat("<", baseUrl, ".", linkProcessor.Key, ">; rel=\"alternate\"; type=\"", linkProcessor.Value, "\"");
```
khellang: I think we have to make this `StringBuilder`-based. Currently this will allocate a lot of strings.

psibernetic:
Request headers should be immutable inside of Nancy imo, changes to the headers could be confusing for debugging and I don't see much reason to mutate them, rather mutate response.

psibernetic:
There are edge cases where you do mutate request, but owin pipeline seems much more appropriate for that as it would apply to multiple middleware, not just Nancy.

khellang:
What if you don't run on OWIN?

psibernetic:
That is a fair point.

psibernetic:
Am I missing something in the commits or is the header actually not readonly in this pull?

psibernetic:
@thecodejunkie

thecodejunkie:
@psibernetic it was changed before the discussion kicked off again

psibernetic:
Gotcha, ship-it!

thecodejunkie:
:eyes:

jchannon:
Not sure why you would change the request's ContentType but I guess we
should allow it. Will making it not read only effect that perf graph?

thecodejunkie:
When it's a `MediaRange` we do not need to do string manipulation all over the place because it does expose some methods on it.

jchannon:
Sounds good. Guess this is ready to go @philliphaydon @khellang ?

src/Nancy/Request.cs
```diff
@@ -234,21 +234,20 @@ public void Dispose()
...
 
-            var contentType = this.Headers["content-type"].First();
-            var mimeType = contentType.Split(';').First();
-            if (mimeType.Equals("application/x-www-form-urlencoded", StringComparison.OrdinalIgnoreCase))
+            var contentType = this.Headers.ContentType;
+            if (contentType.Matches("application/x-www-form-urlencoded"))
             {
                 var reader = new StreamReader(this.Body);
                 this.form = reader.ReadToEnd().AsQueryDictionary();
                 this.Body.Position = 0;
```
phillip-haydon: SCOPE CREEP!!!

I believe this should return rather than falling through to the next if statement check before returning.

phillip-haydon:
Besides my 1 comment that @thecodejunkie's refuses to fix because he hates being a boyscout. 

Everything looks good.

PR #2393, opened 2016-04-01
Added templates for pull-requests and issues

thecodejunkie:
_Do not pull this before it has been discussed_

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,28 @@
+### Prerequisites
+
+- [ ] I have written a descriptive issue title
+- [ ] I have verified that I am running the latest version of Nancy
+- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
+- [ ] I have searched [open](https://github.com/NancyFx/Nancy/issues) and [closed](https://github.com/NancyFx/Nancy/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported
+
+### Description
+<!-- A description of the bug or feature -->
+
+### Steps to Reproduce
+<!-- List of steps, sample code, failing test or link to a project that reproduces the behavior -->
+
+### System Configuration
+<!-- Tell us about the environment where you are experiencing the bug -->
+
+- Nancy version:
+- Nancy host
+  - [ ] ASP.NET
+  - [ ] OWIN
+  - [ ] Self-Hosted
+  - [ ] Other:
+- Other Nancy packages and versions:
+- Environment (Operating system, version and so on):
+- .NET Framework version:
+- Additional information:
+
+<!-- Thanks for reporting the issue to Nancy! -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,11 @@
+### Prerequisites
+
+- [ ] I have written a descriptive pull-request title
+- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open
+- [ ] I have verified made sure that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)
+- [ ] I have provided test coverage for your change (where applicable)
+
+### Description
+<!-- A description of the changes proposed in the pull-request -->
+
+<!-- Thanks for contributing to Nancy! -->


phillip-haydon:
ISSUE_TEMPLATE - This should include the hosting environment (ASP.NET, OWIN, Self Hosting, etc)

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
+### Prerequisites
+
+- [ ] Have you written a descriptive issue title?
+- [ ] Have you verified that you are running the latest version of Nancy?
+- [ ] Have you verified if the problem exist in both `debug` and `release` mode?
+- [ ] Have you searched [open issues](https://github.com/NancyFx/Nancy/issues) to ensure it has not already been reported?
+
+### Description
+[A description of the bug or feature]
```
khellang: I think we should make these HTML comments. Those will show up in the text, but not in a final render.

thecodejunkie: @khellang you are talking about HTML5 placeholders, right? How would we do that? This is just a markdown file that gets entered into the text field when you create a new issue/pull-request. I checked a whole bunch of other repos how they did it and this seemed to be the general approach.

I kind of like it being actual text as well, it will force people to actively replace it :D

khellang: > you are talking about HTML5 placeholders, right?

Nah, I'm just talking about HTML comments.

> I kind of like it being actual text as well, it will force people to actively replace it :D

The problem is that a lot of people won't, so we'll get back loads and loads of issues saying "Thanks for reporting the issue to Nancy!" at the bottom :stuck_out_tongue_closed_eyes:

thecodejunkie: > Nah, I'm just talking about HTML comments. Just wrap it in 

So the idea would be that it showed up in edit-mode, but was hidden once it was posted? I see your point, but I don't think we should make them optional in that way

> The problem is that a lot of people won't, so we'll get back loads and loads of issues saying "Thanks for reporting the issue to Nancy!" at the bottom 

The whole point of the templates is to try and enforce a higher standard of the reported issues. If we call failure on it before we've even tried, it then what's the point? Also, if we have a template, and people fail to fill it out, we'll simply ask them to go back and update the description or the issue will be closed

khellang: > So the idea would be that it showed up in edit-mode, but was hidden once it was posted? I see your point, but I don't think we should make them optional in that way

Yes, that's the point. Optional? What makes them optional by doing it this way? The only thing it does is make sure that it's not echoed back to us as if the issue/PR author wrote it :stuck_out_tongue_closed_eyes: I'm not saying we comment out *everything*, just the placeholders.

> The whole point of the templates is to try and enforce a higher standard of the reported issues. If we call failure on it before we've even tried, it then what's the point? Also, if we have a template, and people fail to fill it out, we'll simply ask them to go back and update the description or the issue will be closed

I'm not "calling failure" on it, I'm simply saying that it's inevitable that a lot of people will fail to fill out everything. Either because it doesn't apply to them, or they're just lazy. I'd just like to keep things that are only targeted at the issue/PR author to *not* show up in the final issue, like "Thanks for reporting the issue to Nancy!".

thecodejunkie: FINE! If you insist on being the voice of reason! :wink: :heart:

thecodejunkie:
Updated to change placeholders to HTML comments + added @phillip-haydon idea about Nancy hosting environment

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
+### Prerequisites
+
+- [ ] Have you written a descriptive issue title?
```
khellang: I'd like to see these as `I have...`. This is just a santiy-check for the author, right?

thecodejunkie: At one point I had them like that.. can't really recall why I went back on it, because reading it again I agree with you =/ Will update

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+<!-- A description of the bug or feature -->
+
+### Steps to Reproduce
+<!-- List of steps, sample code, failing test or link to a project that reproduces the behavior -->
+
+### System Configuration
+<!--
+Tell us about the environment where you are experiencing the bug]
+
+- Nancy version
```
khellang: I'm thinking that these shouldn't be comments, but actually something that the user should fill out. Something like

```
- Nancy version: <!-- Please specify Nancy version -->
- Nancy host
  - [ ] ASP.NET
  - [ ] OWIN
  - [ ] Self-Hosted
  - [ ] Other: <!-- Please specify the host -->
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
+### Prerequisites
+
+- [ ] I have written a descriptive issue title
+- [ ] I have verified that I am running the latest version of Nancy
+- [ ] I have verified if the problem exist in both `debug` and `release` mode
```
khellang: `DEBUG` and `RELEASE`? (caps) :lipstick:

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Description
+<!-- A description of the bug or feature -->
+
+### Steps to Reproduce
+<!-- List of steps, sample code, failing test or link to a project that reproduces the behavior -->
+
+### System Configuration
+<!--
+Tell us about the environment where you are experiencing the bug]
```
khellang: bug~~**]**~~ :lipstick:

.github/ISSUE_TEMPLATE.md
phillip-haydon: I like this idea, riot.js uses it for asking you which labels should be applied to the issue.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
...
+- [ ] I have searched [open issues](https://github.com/NancyFx/Nancy/issues) to ensure it has not already been reported
+
+### Description
+<!-- A description of the bug or feature -->
+
+### Steps to Reproduce
+<!-- List of steps, sample code, failing test or link to a project that reproduces the behavior -->
+
+### System Configuration
+Tell us about the environment where you are experiencing the bug
```
khellang: Comment again? :stuck_out_tongue:

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
+### Prerequisites
+
+- [ ] I have written a descriptive pull-request title
+- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open
+- [ ] I have verified made sure that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)
+- [ ] I have provided test coverage for your change (where applicable)
+
+### Description
+<!--
+[A description of the changes proposed in the pull-request]
```
khellang: Can we split this comment in two, so it's easier to delete this line, while still keeping the "Thanks for contributing to Nancy!" comment in balance? Also, probably remove the wrapping `[]`

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
+### Prerequisites
+
+- [ ] I have written a descriptive issue title
+- [ ] I have verified that I am running the latest version of Nancy
+- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
+- [ ] I have searched [open issues](https://github.com/NancyFx/Nancy/issues) to ensure it has not already been reported
```
xt0rted: Would it be better to say open and closed issues? I saw the [angular-ui team](https://github.com/angular-ui/bootstrap/blob/master/.github/ISSUE_TEMPLATE.md) mentioned both since the issue could have been resolved already.

jchannon: Fair point

PR #2398, opened 2016-04-05
Fixed some reflection problems for CoreCLR in NancySerializationStrategy

thecodejunkie:
These changes are needed so that we can rebase the `coreclr` branch on `master` without compilation errors

### DIFF
--- a/src/Nancy/Json/Simple/NancySerializationStrategy.cs
+++ b/src/Nancy/Json/Simple/NancySerializationStrategy.cs
@@ -16,7 +16,7 @@
         private readonly bool retainCasing;
         private readonly List<JavaScriptConverter> converters = new List<JavaScriptConverter>();
         private readonly List<JavaScriptPrimitiveConverter> primitiveConverters = new List<JavaScriptPrimitiveConverter>();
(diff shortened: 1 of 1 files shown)


src/Nancy/Json/Simple/NancySerializationStrategy.cs
```diff
@@ -148,12 +148,12 @@ protected override bool TrySerializeKnownTypes(object input, out object output)
             }
 
             var inputType = input.GetType().GetTypeInfo();
-            if (this.TrySerializeJavaScriptConverter(input, out output, inputType))
```
khellang: You have `var inputType = input.GetType().GetTypeInfo();` above, can't you use that? :stuck_out_tongue:

thecodejunkie: That spits out a `TypeInfo` which is what I can't be using, I need the `Type`. We could split it up so we do something like

```
var foo = input.GetType();
var bar = foo.GetTypeInfo();
```

and use `foo` and `bar` when needed. Would safe us two calls to `.AsType()` ?

khellang: Yeah, that's what I meant.

```csharp
var inputType = input.GetType();
var inputTypeInfo = inputType.GetTypeInfo();
```

Issue #2401, opened 2016-04-07
NUnit and NCrunch are not ignored during DefaultNancyBootstrapper.AutoRegister


westonized:
I have just tracked a large amount of slow tests down to this.

xunit and other test frameworks are ignored, but not nunit and ncrunch. ("MonoDevelop.NUnit" is excluded, but not regular "NUnit").

They are commonly used, make no sense to have auto discover anything from them, and though I can fix it for my project it would be a general benefit to fix for all users.

e.g. this is where `"xunit"` is ignored:

/src/Nancy/DefaultNancyBootstrapper.cs:

```
asm => asm.FullName.StartsWith("xunit", StringComparison.Ordinal),
```

I'm happy to make a PR if you agree?

thecodejunkie:
Happy to take a PR :)

khellang:
Closed with #2402

PR #2402, opened 2016-04-07
Exclude nunit and nCrunch from automatic bootstrap registration

westonized:
Pr for #2401

I have just tracked a large amount of slow tests down to this.

xunit and other test frameworks are ignored, but not nunit and ncrunch. ("MonoDevelop.NUnit" is excluded, but not regular "NUnit").

They are commonly used, make no sense to have auto discover anything from them, and though I can fix it for my project it would be a general benefit to fix for all users.

e.g. this is where "xunit" is ignored:

/src/Nancy/DefaultNancyBootstrapper.cs:

```
asm => asm.FullName.StartsWith("xunit", StringComparison.Ordinal),
```


jchannon:
LGTM

Issue #2355, opened 2016-03-14
Performances issues with multipart upload


ClemPi:
This has been tested on Windows 10, Linux running mono 3.12 and 4.4.2 with Nancy 1.4.3 and master.

I experienced some serious performances issues uploading files with NancyFX running Hosting.Self on localhost.

eg : 30MB/s for 1GB, 20MB/s for 4GB and 16MB/s for 16GB.

I started profiling the application with VisualStudio 2015, here are some screenshots : 

![d1](https://cloud.githubusercontent.com/assets/1401705/13750649/b2d35308-ea06-11e5-8902-a70c88b21b1b.jpg)
![d2](https://cloud.githubusercontent.com/assets/1401705/13750648/b2ccef2c-ea06-11e5-9639-3ff758c8d87a.jpg)
![d3](https://cloud.githubusercontent.com/assets/1401705/13750647/b2ccbad4-ea06-11e5-9079-e4ffabaf7460.jpg)

Here is the main hot point : https://github.com/NancyFx/Nancy/blob/master/src/Nancy/HttpMultipart.cs#L112

jchannon:
Awesome work, thanks!

PR #2409, opened 2016-04-08
Change the multipart lib to improve performances

ClemPi:
Related to #2355 I reworked the Multipart handling and used https://github.com/Vodurden/Http-Multipart-Data-Parser to do so.

Here are some performances test I made:

|    | before | after |
| ----- | -------- | ------- |
| 2GB | 55s    | 33s   |
| 4GB | 120s   | 75s   |
| 8GB | 314s   | 285s  |

Please let me know if this look OK for you before I start working on the tests. I included all the lib, maybe we want to only pick relevant classes.

Regards


thecodejunkie:
Hi,

Nice! This is one of those things that you just don't pull in without giving it a proper code review and step the code a bit. Even then it's a nightmare to dare to pull the merge button because of the uncertainty of that scenarios that might be broken :D

So before we start reviewing that, there are some pretty big code style violations that would need to be fixed. I would prefer if they were addressed before we even started looking at the actual implementation logic. You can find the style guides [here](https://github.com/NancyFx/Nancy/blob/master/.github/CONTRIBUTING.md#style-guidelines) (they were also linked during the pull-request process)

Thanks

ClemPi:
I'm aware that the code added in HttpMultipart directory is not coding style compliant.
As I said the code is coming from a lib, and I think it will be easier to maintain this kind of "external" dependency if we don't change the code too much. Looks like it's already the case for TinyIOC.

Regards

psibernetic:
Rather than directly pull in the codebase, can we learn from it and apply changes to the existing handling?

PR #2418, opened 2016-04-20
Update to run on rc2 and dotnet cli

thecodejunkie:
### Description
This is bringing the `coreclr` branch up to the `rc2` bits, including `dotnet cli` (things such as the new `DependencyContext`) and the correct package versions of `corefx`. This will be WIP (and tests will keep failing) until the next drop of [dotnet-test-xunit](https://github.com/dotnet/coreclr.xunit) is available on myget, so that we can run the tests of both the full-framework and coreclr. That pull-request is ready but is waiting for https://github.com/dotnet/cli/pull/2527 which will be available on the latest dotnet builds as soon as it has been pulled in.

#### Modifications
- Normal projects now target `net452` (instead of `net451` because that's quickly coming to its end) and `netstandard1.5`. The reason we had to bump from `netstandard1.3` is that Microsoft is also doing the same of their ASP.NET / MVC Core 1 bits and we need to follow along because their packages won't target anything less and we have to use `Microsoft.Extensions.DependencyModel` for the `DependencyContext` bits (for the `IAssemblyCatalog` implementation on coreclr). This is fine, since we still have `net452` defined for the full-framework and it makes sense to use `netstandard1.5` for Nancy on core
- All test projects have been updated to a combination of `net452` and `netcoreapp1.0` and have had their `project.json` files updated to support the new test runner setup approach used by the dotnet cli

#### Package versions
- I have switched over the `aspnetcidev` nuget feed over to `aspnetcirelease` feed as that is what will contain the `rc2` packages that are guaranteed to work together without any package downgrades and so on. The `aspnetcidev` feed is already containing post-rc2 stuff and we do not want to get involved with those at this time.
- Due to the above change, we can now use `xxxxxxx-rc2-*` as the package versions

#### Development environment
Right now, Visual Studio is not an option because it still uses `dnx` and we need to use `dotnet cli`. For the time being, the best we can get Visual Studio Code with Omnisharp and the VSCode C# Debugger. It let's use use the rc2 stuff because it uses the `dotnet cli`. You are going to have to use an insiders build of VSCode https://code.visualstudio.com/Download#insiders and of the OmniSharp extension https://github.com/OmniSharp/omnisharp-vscode/releases (do not try and run it, download the `vsix` and open it with VSCode using `open file` from the menu to install it)

#### Remaining before we can pull this in

- [x] Rename `DNX` compiler directive to `CORE`
- [x] Add XML comments to `DependencyContextAssemblyCatalog`
- [x] Split files in `Nancy.Demo.Hosting.Kestrel`
- [x] Remove `UseDefaultHostingConfiguration` call in `Nancy.Demo.Hosting.Kestrel/Program.cs`
- [x] Make sure `Microsoft.NETCore.App` is pegged to `1.0.0-rc2-*` in `Nancy.Demo.Hosting.Kestrel`
- [x] Check so `preserveCompilationContext` is only declared where it is needed
- [x] Rebase `master` ➡ `coreclr` ➡ `coreclr-update-tfm-and-testing`
- [ ] Squash commits


jchannon:
Failed on appveyor :smile

PR #2419, opened 2016-04-21
Removed JabbR and added slack information

Sphiecoh:
### Description
I removed JabbR info since its practically dead now and we want people to join us on Slack


khellang:
Thanks! :heart:

PR #2421, opened 2016-04-21
[WIP] Updated ViewRenderer to return Task

thecodejunkie:
### Prerequisites

- [X] I have verified made sure that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)
- [X] I have provided test coverage for your change (where applicable)

### Description
Updated `ViewRenderer` to return `Task<Negotiator>` instead of `Negotiator`. This means that from an `NancyModule` you can now call `await View["index"]`. Also created a `LegacyViewRenderer` with the old behavior that is used by `LegacyModule`. I also had to remove the `INancyModule.View` declaration since there is no `IViewRenderer` interface, nor can there be since they differ on the return value. To be honest this should not be on `INancyModule` at all (could prune more, but that's out of the context of this pull-request), it should be an implementation detail.

Also sneaked in a couple of file fixes in `Nancy.sln` for three files that weren't included as links

### DIFF
--- a/samples/Nancy.Demo.CustomModule/UglifiedNancyModule.cs
+++ b/samples/Nancy.Demo.CustomModule/UglifiedNancyModule.cs
@@ -43,9 +43,9 @@
 
         public string ModulePath { get; private set; }
 
-        public ViewRenderer View
+        public LegacyViewRenderer View
         {
-            get { return new ViewRenderer(this); }
+            get { return new LegacyViewRenderer(this); }
         }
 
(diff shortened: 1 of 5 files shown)


thecodejunkie:
I'll investigate if this can be replaced by the same fix applied in #2422

thecodejunkie:
Will be fixed by #2422

PR #2441, opened 2016-05-11
Introduced new route declaration syntax

thecodejunkie:
### Prerequisites

- [X] I have verified made sure that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)
- [X] I have provided test coverage for your change (where applicable)

### Description

In `2.0-alpha` we changed the route syntax and made all the routes `async` which meant they all had to return `Task<T>`. We also temporarily introduced `LegacyNancyModule` which was a custom module class which mimicked the old, synchronous, route behavior with the intent to help people migrate to `2.0` and then we would remove it in a later release (perhaps even before 2.0 RTM). As it turns out, forcing `Task<T>` on all routes wrecked our syntax a bit it made it difficult to have routes that returned primitives such as `int` (status code), `HttpStatusCode`, `string` (response body) or even to call `View` or `Negotiate`. We did try some tricks, like custom awaiters and so on which made the syntax a bit nicer but it still forced you to declare all routes as `async` and then `await`

```c#
Get["/"] = async (args, ct) => await 200;

Get["/"] = async (args, ct) => await View["index"];
```

Not too shabby, but it still forces you to declare these simple routes, which aren't async by nature, as async which would cause the entire async state machine to be generated even though you did not need them. Clearly not ideal, and most certainly not very Super-Duper-Happy-Path

### Proposal

For quite some time now, we have talked internally about a new route declaration syntax. Not a massive change, but one that would mean we would trade in the indexer syntax `Get["/"]` (they have been a signature mark for Nancy since day one) for a method based approach `Get("/")`.

See https://github.com/NancyFx/Nancy/pull/2441#issuecomment-219382476 for samples

### Code change mentions

- ~~I have introduced a delegate `T RouteAction<T>(dynamic parameters, CancellationToken token)` which represents the body of a route~~
- I have made `Route` abstract and introduced `Route<T>`. The base class `Route` does not contain an `Action` anymore, the idea is that a Route doesn't per-say need to have an action it just needs to be invocable so you could inherit different kinds of `Route` if you have any other desires. The `NancyModule` declarations uses `Route<T>` which has the `Action` property of type `RouteAction<T>` that in turn will be called when `Route.Invoke` is invoked.
- ~~Introduced `RouteResultHelper` which is used to take a route response, which may or may not be a Task, and coereces the result in a format what we can use it. It does some funky casting with the help of reflection and generic method signatures. This class is used by `DefaultRouteInvoker` and `DiagnosticsHook` + in a test where we needed to mock the `IRouteInvoker` and have it return a proper response.~~
- All route declaration method (`Get()`, `Post()` and so on) on `NancyModule` have been made `virtual` so you can override and intercept the declarations if you'd like to do something to it before it is stored
- `dynamic` is only used in the route lambdas now, from there on it should be `DynamicDictionairy`

### Worth noting

- ~~Since we are inferring `T` you cannot return instances of a private model~~
- ~~For the same reason you need to explicitly define a model `object` for routes that returns an anonymous model `Get<object>("/", (args, ct) => new { Name = "Nancy" });`~~
- I have removed `LegacyNancyModule` as it serves very little point now and converting to the new syntax is very fast


jchannon:
Should we add an overload to remove the `CancellationToken` from sync routes?
```
Get("/", (args, ct) => {
   return "Hi"
});
```

So it becomes

```
Get("/", (args) => {
   return "Hi"
});
```

thecodejunkie:
@jchannon yeah, doing that now

thecodejunkie:
@jchannon pushed + updated description

src/Nancy/Routing/Route.cs
```diff
@@ -56,7 +56,7 @@ public class Route<T> : Route
         /// </summary>
         /// <param name="description"></param>
         /// <param name="action">The action that should take place when the route is invoked.</param>
-        public Route(RouteDescription description, RouteAction<T> action)
+        public Route(RouteDescription description, Func<dynamic, CancellationToken, T> action)
```
khellang: Shouldn't these return a Task<T>?

src/Nancy.Testing/ConfigurableNancyModule.cs
```diff
@@ -251,9 +252,9 @@ public ConfigurableNancyModuleConfigurator Put(string path, Func<dynamic, Legacy
             /// <param name="condition">The condition that has to be fulfilled in order for the route to be invoked</param>
             /// <param name="action">The action that should be invoked by the route.</param>
             /// <returns>An instance to the current <see cref="ConfigurableNancyModuleConfigurator"/>.</returns>
-            public ConfigurableNancyModuleConfigurator Put(string path, Func<NancyContext, bool> condition, Func<dynamic, LegacyNancyModule, dynamic> action)
+            public ConfigurableNancyModuleConfigurator Put(string path, Func<NancyContext, bool> condition, Func<dynamic, NancyModule, dynamic> action)
             {
-                this.module.Post[path, GetSafeRouteCondition(condition)] = GetSafeRouteAction(action);
+                this.module.Put(path, this.GetSafeRouteAction(action), null, GetSafeRouteCondition(condition));
```
khellang: Was this a bug? :stuck_out_tongue_closed_eyes:

jchannon:
I see you made the CancellationToken optional but would there be a case where you'd want it on a non-async route?

khellang:
Why would you cancel a synchronous operation using a `CancellationToken`?

jchannon:
@khellang who's that for?

khellang:
You :smile:

jchannon:
You wouldn't that's why I'm asking why make it available on a sync route

khellang:
> You wouldn't that's why I'm asking why make it available on a sync route

Oh, I thought you were asking if we could add it. The main problem right now is that all the methods accept non-async delegates (no `Task<T>`) return.

src/Nancy/NancyModule.cs
```diff
@@ -263,80 +380,35 @@ public virtual ModelValidationResult ModelValidationResult
...
+        /// Declares a route for the module
         /// </summary>
-        public class RouteBuilder : IHideObjectMembers
+        /// <typeparam name="T"></typeparam>
+        /// <param name="name">Name of the route</param>
+        /// <param name="method">The HTTP method that the route will response to</param>
+        /// <param name="path">The path that the route will respond to</param>
+        /// <param name="action">Action that will be invoked when the route it hit</param>
+        /// <param name="condition">A condition to determine if the route can be hit</param>
+        protected void AddRoute<T>(string name, string method, string path, Func<NancyContext, bool> condition, Func<dynamic, CancellationToken, T> action)
```
horsdal: Seems odd to swap the `condition` and `action` args around compared to the verb methods.

jchannon: Ha ha, well spotted! 😄

src/Nancy/Routing/RouteResultHelper.cs
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            Type innerType = resultType.GetGenericArguments()[0];
+
+            var converter = ConverterCache.GetOrAdd(innerType, t =>
+            {
+                return ConvertMethod.MakeGenericMethod(t);
+            });
+
+            return await (Task<object>)converter.Invoke(null, new [] { routeResult });
```
horsdal: Is it necessary to `await` here? Why not return the task and drop the `async` from the signature (and add `Task.FromResult` on the other `return`s)?

src/Nancy/Routing/DefaultRouteInvoker.cs
```diff
@@ -39,7 +37,10 @@ public DefaultRouteInvoker(IResponseNegotiator negotiator)
 
             try
             {
-                result = await route.Invoke(parameters, cancellationToken).ConfigureAwait(false);
+                dynamic routeResult =
+                    route.Invoke(parameters, cancellationToken);
+
+                result = await RouteResultHelper.CoerceResult(routeResult);
```
horsdal: Needs `.ConfigureAwait(false)`

src/Nancy/Routing/DefaultRouteInvoker.cs
```diff
@@ -58,7 +59,7 @@ public DefaultRouteInvoker(IResponseNegotiator negotiator)
                 result = new Response();
             }
 
-            return this.negotiator.NegotiateResponse(result, context);
+            return await this.negotiator.NegotiateResponse(result, context);
```
horsdal: Needs `.ConfigureAwait(false)`

src/Nancy/Routing/RouteResultHelper.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        private static readonly MethodInfo ConvertMethod = typeof(RouteResultHelper).GetMethod("Convert", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static);
+        private static readonly ConcurrentDictionary<Type, MethodInfo> ConverterCache = new ConcurrentDictionary<Type, MethodInfo>();
+
+        /// <summary>
+        /// Coerces the result.
+        /// </summary>
+        /// <param name="routeResult">The result that should be coereced.</param>
+        /// <returns>A <see cref="Task{TResult}"/> of <see cref="object"/>.</returns>
+        public static async Task<object> CoerceResult(object routeResult)
```
horsdal: This is quite clever, btw. :sunglasses:

src/Nancy/NancyModule.cs
thecodejunkie: They weren't really swapped around, but moved out of necessity. The `AddRoute` parameter order is left unchanged while the verb methods have been rearranged because `condition` and `name` are optional.

src/Nancy.Testing/ConfigurableNancyModule.cs
thecodejunkie: Sure was :smile:

thecodejunkie:
@khellang me and @grumpydev spent some time this morning verifying the behavior of it not being `Task<T>` under the premiss that it would be inferred + verified the behavior of the result coercion (found a bug). It all seem to work as expected in terms of asynchronicity https://gist.github.com/grumpydev/048e8ab6fb390dfa53d77d3cc68a4b19

src/Nancy/NancyModule.cs
horsdal: Makes sense. Guess I'd have swapped them in `AddRoute` too then.

thecodejunkie:
> :information_source:  Update the route declaration overloads to support both sync and async routes, both with or without a cancellation token (ignore the silly sample routes, just needed compiling routes to show what it is now capable of)

```
public class RoutingModule : NancyModule
{
    public RoutingModule()
    {
        Get("/", args =>
        {
            return 200;
        });

        Get("/", (args, ct) =>
        {
            return Task.FromResult(200);
        });

        Get("/", async args =>
        {
            return await Task.FromResult(200);
        });

        Get("/", async (args, ct) =>
        {
            return await Task.FromResult(200);
        });
    }
}
```

thecodejunkie:
All CI builds are green but needs a proper reviewing before we pull it in

ping @NancyFx/most-valued-minions @NancyFx/owners

src/Nancy.MSBuild/Nancy.csproj
```diff
@@ -1364,6 +1358,9 @@
     <Compile Include="..\Nancy\Configuration\ConfigurationException.cs">
       <Link>Configuration\ConfigurationException.cs</Link>
     </Compile>
+    <Compile Include="..\Nancy\Routing\Route.cs">
+      <Link>Route.cs</Link>
```
jchannon: This needs to be Routing\Route.cs

test/Nancy.Tests/Unit/Routing/DefaultRouteResolverFixture.cs
```diff
@@ -395,7 +395,7 @@ public class DefaultRouteResolverFixture
         }
 
         [Fact]
-        public async Task Should_set_allowed_method_on_response_when_returning_405()
+        public async Task Shouldargssetargsallowedargsmethodargsonargsresponseargswhenargsreturningargs405()
```
jchannon: Might want to rename this

test/Nancy.Tests/Unit/Routing/DefaultRouteResolverFixture.cs
```diff
@@ -375,7 +375,7 @@ public class DefaultRouteResolverFixture
         }
 
         [Fact]
-        public async Task Should_not_return_405_if_requested_method_is_not_permitted_but_others_are_available_and_disabled()
+        public async Task Shouldargsnotargsreturnargs405argsifargsrequestedargsmethodargsisargsnotargspermittedargsbutargsothersargsareargsavailableargsandargsdisabled()
```
jchannon: Might want to rename this

test/Nancy.Tests/Unit/Routing/DefaultRouteResolverFixture.cs
```diff
@@ -355,7 +355,7 @@ public class DefaultRouteResolverFixture
         }
 
         [Fact]
-        public async Task Should_return_405_if_requested_method_is_not_permitted_but_others_are_available_and_not_disabled()
+        public async Task Shouldargsreturnargs405argsifargsrequestedargsmethodargsisargsnotargspermittedargsbutargsothersargsareargsavailableargsandargsnotargsdisabled()
```
jchannon: Might want to rename this

test/Nancy.Tests/Unit/Routing/DefaultRouteResolverFixture.cs
```diff
@@ -342,7 +342,7 @@ public class DefaultRouteResolverFixture
         }
 
         [Fact]
-        public async Task Should_return_404_if_no_root_found_when_requesting_it()
+        public async Task Shouldargsreturnargs404argsifargsnoargsrootargsfoundargswhenargsrequestingargsit()
```
jchannon: Might want to rename this

test/Nancy.Tests/Unit/Routing/DefaultRouteResolverFixture.cs
```diff
@@ -329,7 +329,7 @@ public class DefaultRouteResolverFixture
         }
 
         [Fact]
-        public async Task Should_handle_options_requests()
+        public async Task Shouldargshandleargsoptionsargsrequests()
```
jchannon: Might want to rename this

test/Nancy.Tests/Unit/Routing/DefaultRouteResolverFixture.cs
```diff
@@ -9,7 +9,7 @@
     public class DefaultRouteResolverFixture
```
jchannon: All the tests are fooked! someone do a string replace on underscore?

src/Nancy/Routing/Route.cs
```diff
@@ -69,77 +42,67 @@ public Route(string method, string path, Func<NancyContext, bool> condition, Fun
...
-                };
+        /// <summary>
+        /// Invokes the route with the provided <paramref name="parameters"/>.
+        /// </summary>
+        /// <param name="parameters">A <see cref="DynamicDictionary"/> that contains the parameters that should be passed to the route.</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns>A (hot) task of <see cref="Response"/> instance.</returns>
+        public override Task<object> Invoke(DynamicDictionary parameters, CancellationToken cancellationToken)
+        {
+            return this.Action.Invoke(parameters, cancellationToken).ContinueWith<object>(t => t.Result, TaskContinuationOptions.OnlyOnRanToCompletion);
```
jchannon: Why do you need the `ContinueWith`

thecodejunkie: Because `this.Action` returns `Task<T>` and we need convert it to a `Task<object>`

jchannon:
:shipit

psibernetic:
Looking forward to testing this one, I would imagine this will make it simpler to add Swaggeresque metadata overloads in the future or even as an extension.

jchannon:
Yup

samples/Nancy.Demo.Authentication.Forms/MainModule.cs
```diff
@@ -6,23 +6,23 @@ namespace Nancy.Demo.Authentication.Forms
...
+                dynamic model = new ExpandoObject();
+                model.Errored = this.Request.Query.error.HasValue;
 
-                    return View["login", model];
-                };
+                return View["login", model];
+            });
 
-            Post["/login"] = x => {
+            Post("/login", args => {
```
phillip-haydon: OMG @thecodejunkie can you stop with the inconsistent braces!!! :)

samples/Nancy.Demo.Authentication.Stateless/AuthModule.cs
```diff
@@ -1,27 +1,28 @@
...
     {
         public AuthModule() : base("/auth/")
         {
             //the Post["/login"] method is used mainly to fetch the api key for subsequent calls
-            Post["/"] = x =>
-                {
-                    string apiKey = UserDatabase.ValidateUser((string) this.Request.Form.Username,
-                                                              (string) this.Request.Form.Password);
+            Post("/", args =>
+            {
```
phillip-haydon: I know these are demo project's but the file above this has the brace open on the same line, then this opens on the next line!

I'm becoming the stylecop now.

samples/Nancy.Demo.Authentication.Stateless/RootModule.cs
```diff
@@ -1,15 +1,15 @@
 namespace Nancy.Demo.Authentication.Stateless
 {
-    public class RootModule : LegacyNancyModule
+    public class RootModule : NancyModule
     {
         public RootModule()
         {
-            Get["/"] = _ => this.Response.AsText("This is a REST API. It is in another VS project to " +
+            Get("/", args => this.Response.AsText("This is a REST API. It is in another VS project to " +
                                                  "demonstrate how a common REST API might behave when " +
```
phillip-haydon: Thing's don't line up anymore!

samples/Nancy.Demo.Hosting.Aspnet/MainModule.cs
```diff
@@ -237,9 +273,15 @@ public MainModule(IRouteCacheProvider routeCacheProvider, INancyEnvironment envi
                 }
 
                 return View["FileUpload", new { Posted = fileDetails }];
-            };
+            });
 
-            Get["NamedRoute", "/namedRoute"] = _ => "I am a named route!";
+            Get(
+                name: "NamedRoute",
```
phillip-haydon: This
  line is not like the other lines.

src/Nancy/Diagnostics/Modules/MainModule.cs
```diff
@@ -4,12 +4,15 @@ public class MainModule : DiagnosticModule
     {
         public MainModule()
         {
-            Get["/"] = async (_, __) =>
+            Get("/", _ =>
```
phillip-haydon: I'm going to be anal now and ask, are we going to be consistent and call it `args` or `_` everywhere?

src/Nancy/Diagnostics/Modules/TraceModule.cs
```diff
@@ -12,17 +12,17 @@ public TraceModule(IRequestTracing sessionProvider)
...
 
-            Get["/sessions"] = async (_, __) =>
+            Get("/sessions", _ =>
             {
                 return this.Response.AsJson(this.sessionProvider.GetSessions().Select(s => new { Id = s.Id }).ToArray());
-            };
+            });
 
-            Get["/sessions/{id}"] = async (ctx, __) =>
+            Get("/sessions/{id}", ctx =>
```
phillip-haydon: `args` `_` `ctx`...

src/Nancy/Routing/DefaultRouteInvoker.cs
```diff
@@ -35,7 +33,7 @@ public DefaultRouteInvoker(IResponseNegotiator negotiator)
         /// <returns>A <see cref="Response"/> instance that represents the result of the invoked route.</returns>
         public async Task<Response> Invoke(Route route, CancellationToken cancellationToken, DynamicDictionary parameters, NancyContext context)
         {
-            dynamic result;
+            object result;
```
phillip-haydon: Why is this now object?

src/Nancy/Routing/Route.cs
phillip-haydon: I noticed further up you called an action like `action(...)` but here you're using `Action.Invoke(...)` I don't believe it makes much different but, consistency? :)

test/Nancy.Metadata.Modules.Tests/FakeNancyModule.cs
```diff
@@ -2,34 +2,34 @@
...
-            public FakeNancyModuleConfigurator(LegacyNancyModule module)
+            public FakeNancyModuleConfigurator(NancyModule module)
             {
                 this.module = module;
             }
 
             public FakeNancyModuleConfigurator AddDeleteRoute(string path)
             {
-                this.module.Delete[path] = parameters => {
+                this.module.Delete(path, args => {
```
phillip-haydon: AHHH BRACE INCONSISTENCY AGAIN! LOL

test/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
```diff
@@ -184,35 +183,33 @@ public class ContentNegotiationFixture
...
+                with.Get("/headers", (x, m) =>
+                {
+                    var context =
+                        new NancyContext();
 
-              var negotiator =
-                  new Negotiator(context);
-              negotiator.WithContentType("text/xml");
+                    var negotiator =
+                        new Negotiator(context);
```
phillip-haydon: Why is this on a new line?

phillip-haydon:
Splooge

samples/Nancy.Demo.Authentication.Forms/MainModule.cs
thecodejunkie: I've not changed a single brace

src/Nancy/Routing/DefaultRouteInvoker.cs
thecodejunkie: Because outside of the actual route lambdas, we don't need `dynamic` .. using `dynamic` in other places has side effects such as the entire method body being treated as dynamic so the compiler cannot infer a return type tc

test/Nancy.Tests.Functional/Tests/ContentNegotiationFixture.cs
thecodejunkie: Nothing but indentation change

davidfowl:
I approve of this syntax 👍

grumpydev:
Fowler is agreeing with us.. Anyone else suspicious? 😂

phillip-haydon:
So we can close this PR now?

thecodejunkie:
@phillip-haydon if we have _properly_ reviewed the changes around async stuff then yes.. if no, then we need to review it :P For example are we using `ConfigureAwait(false);` in relevant places in the change etc?

horsdal:
Reviewed a second time. Didn't find anything :smile_cat:

phillip-haydon:
Held up by @khellang as always. :eyes:

jchannon:
tumblr_ln9eat2mn61qa9fb5o1_500

psibernetic:
May I request a new release for this?

thecodejunkie:
@psibernetic we're working on it. in the meantime you can always grab it of our myget feed which gets built on each new commit to `master`

psibernetic:
Yeap thank you, fellas in chat pointed me there, been testing away.

canton7:
The wiki still needs updating, BTW

thecodejunkie:
@canton7 the wiki will not be updates as the 2.0 packages comes out of pre-release, until then all changes are considered as being pending =)

canton7:
My bad, apologies for the noise.

steve2:
Are there ever going to be doc updates that use the new syntax?

PR #2438, opened 2016-05-06
Added DateTimeStyleConversion to JsonConfiguration

jchannon:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open
- [x] I have verified made sure that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)
- [x] I have provided test coverage for your change (where applicable)

### Description
Addresses #2417 & #2427

PR #2454, opened 2016-05-24
Use GlobalizationConfiguration.DateTimeStyles in DynamicDictionaryValue DateTime parsing

thecodejunkie:
### Prerequisites

### Description
This will make `DynamicDictionaryValue` use the `GlobalizationConfiguration.DateTimeStyles` when parsing dates, represented as strings, to `DateTime`. This will let you control how you want dates to be handled in your application. 

Once pull in, we should update #2438 to use the new settings as well

### DIFF
--- a/src/Nancy/DynamicDictionaryValue.cs
+++ b/src/Nancy/DynamicDictionaryValue.cs
@@ -5,21 +5,34 @@
     using System.Dynamic;
     using System.Globalization;
     using System.Linq.Expressions;
-
     using Microsoft.CSharp.RuntimeBinder;
-    using Nancy.Routing.Trie.Nodes;
 
+    /// <summary>
+    /// A value that is stored inside a <see cref="DynamicDictionary"/> instance.
+    /// </summary>
     public class DynamicDictionaryValue : DynamicObject, IEquatable<DynamicDictionaryValue>, IHideObjectMembers, IConvertible
     {
         private readonly object value;
+        private readonly GlobalizationConfiguration globalizationConfiguration;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DynamicDictionaryValue"/> class.
         /// </summary>
         /// <param name="value">The value to store in the instance</param>
         public DynamicDictionaryValue(object value)
+            : this(value, GlobalizationConfiguration.Default)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DynamicDictionaryValue"/> class.
+        /// </summary>
+        /// <param name="value">The value to store in the instance</param>
+        /// <param name="globalizationConfiguration">A <see cref="GlobalizationConfiguration"/> instance.</param>
+        public DynamicDictionaryValue(object value, GlobalizationConfiguration globalizationConfiguration)
         {
             this.value = value;
+            this.globalizationConfiguration = globalizationConfiguration;
         }
 
         /// <summary>
@@ -52,13 +65,12 @@
             {
                 try
                 {
-                    return (T)value;
+                    return (T)this.value;
                 }
                 catch
                 {
-                    var typeName = value.GetType().Name;
-                    var message = string.Format("Cannot convert value of type '{0}' to type '{1}'",
-                                                typeName, typeof(T).Name);
+                    var typeName = this.value.GetType().Name;
+                    var message = string.Format("Cannot convert value of type '{0}' to type '{1}'", typeName, typeof(T).Name);
 
                     throw new InvalidCastException(message);
                 }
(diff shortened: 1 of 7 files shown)


jchannon:
LGTM

jchannon:
@phillip-haydon any comments?

src/Nancy/DynamicDictionaryValue.cs
```diff
@@ -5,21 +5,34 @@
     using System.Dynamic;
     using System.Globalization;
     using System.Linq.Expressions;
-
     using Microsoft.CSharp.RuntimeBinder;
-    using Nancy.Routing.Trie.Nodes;
 
+    /// <summary>
+    ///
```
khellang: :frowning:

test/Nancy.Tests/Unit/DynamicDictionaryValueFixture.cs
```diff
@@ -987,6 +987,38 @@ public void Should_return_default_value_if_implicit_convert_fails_on_datetime()
         }
 
         [Fact]
+        public void Should_adjust_to_universal_time_when_globalizationconfiguration_datetimestyles_requires_it()
+        {
+            // Given
+            var expected = DateTime.Parse("2016-05-24T08:41:37.1661565");
+
+            var config = new GlobalizationConfiguration(new [] {"en-US"}, dateTimeStyles: DateTimeStyles.AdjustToUniversal);
+            var value = new DynamicDictionaryValue("2016-05-24T10:41:37.1661565+02:00", config);
```
khellang: We're sure this won't bite us when DST goes away, or when this test runs on a machine not in CEST?

jchannon: Double check what the SimpleJSON tests do that I fixed when DST kicked in a few months ago

jchannon: https://github.com/NancyFx/Nancy/pull/2385/files not sure if this helps tbh

thecodejunkie: My brain hurts.. plz to tell how to fix :D

jchannon: This should be ok shouldn't it? If we say we are 2 hours ahead of UTC no matter the DST it will always take that 2 hours off.

phillip-haydon:
Looks good to me, hopefully fixes that guys problem who raised the issue in slack.

jchannon:
@khellang happy?

test/Nancy.Tests/Unit/DynamicDictionaryValueFixture.cs
thecodejunkie: I call on the powers of the @jskeet :D

jskeet: Not sure I understand what the test is trying to do, but a) I would set the expected value by calling the `DateTime` constructor and specifying the kind explicitly and b) if you're assuming a particular time zone, that's pretty worrying. However, if the time zone *is* set explicitly somewhere, then it's reasonable to assume that the rules for May 24th 2016 won't be changed in the future.

thecodejunkie: Basically just want to have a test that makes sure it can adjust to universal and @khellang where worried that the test might break once daylight savings change because of the explicit `+2` offset

jskeet: If the test is being run in a time zone which is UTC+2 *on May 24th 2016* then it won't (shouldn't) matter if it's some other offset when the test is run. If the date were in the future, you would be at the mercy of governments changing time zone rules - but it's a bit too late for them to change the rules for May 24th 2016.

The call to `DateTime.Parse` still bothers me though - I have to look up the documentation to know *exactly* what you're expecting.

jchannon: The past date the test is using doesn't really matter we just wanted to make sure if the string passed in with an offset of +2, that our code will convert that to UTC.  I think we should up change the DateTime to `new DateTime(2016, 5, 24, 08, 41, 37, DateTimeKind.UTC)`

khellang: I think what @jskeet said makes sense. Instead of `DateTime.Parse`, use a ctor and specify universal explicitly, or add `Z` to the string, but that's probably less obvious.

I now realize that time zones won't be an issue here. We're basically just testing that a local timestamp, `2016-05-24T10:41:37.1661565+02:00`, will be adjusted to a universal timestamp, `2016-05-24T08:41:37.1661565Z`. This is fiiiine...

jskeet: Ah, true - yes, I hadn't spotted that the offset is in the string as well. That should be fine.

jchannon: Agreed, @thecodejunkie make the change to `new DateTime(2016, 5, 24, 08, 41, 37, DateTimeKind.UTC)` and we'll merge

thecodejunkie: cheers @jskeet sorry for dragging you into a random discussion, but we don't know many whom are as comfortable around timezones as you are :D

thecodejunkie:
It's :white_check_mark: :smile:

PR #2456, opened 2016-05-25
Catch exception in Middleware ans set 500 and write exception message

jchannon:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open
- [x] I have verified that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)
- [ ] I have provided test coverage for my change (where applicable)

### Description
Addresses #1594

Issue #1594, opened 2014-07-05
owin.ResponseStatusCode not set on uncaught exceptions


jchannon:
By default if an exception is thrown Nancy captures it and sets `owin.ResponseStatusCode` to 500.  If for example, you throw an exception in a `IStatusCodeHandler` `owin.ResponseStatusCode` is set to 200.  If the exception is not caught in the owin middleware the owin host will convert it to a 500.  This is fine because the final response is 500 but if you have middleware inspecting the owin environment dictionary its misleading.

I've looked at the `NancyMiddleware.cs` https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Owin/NancyMiddleware.cs#L98 but I can't see where to set the status code as the `NancyEngine` catches the exception.  I guess we could check if `context.Items` has the owin dictionary in it and set the response status code where the exception gets caught but the whole Task voodoo going on in there seems delicate so I didn't want to change anything without checking where the best place to do it is first.

Thoughts?

thecodejunkie:
So you guys agreed on a solution, but never implemented it? :smiley_cat:

jchannon:
I don't know, is that the solution?

jchannon:
I've just hit this again debugging an production issue!!

jchannon:
Checked with @psibernetic and its still in master

psibernetic:
IMO Nancy should catch exceptions and set the response status code to 500 if nothing else intervened so that other middleware have a chance to take part in the response party rather than needing to catch exceptions.

thecodejunkie:
@jchannon @psibernetic perfect opportunity to become bff's and work on a pull-request together :smile:

jchannon:
I'll let you do it tomorrow. It's in the scary bit of the code base :p

khellang:
Closed with #2457

PR #2457, opened 2016-05-25
Catch calls to statuscode handler and write error response if caught

jchannon:
### Prerequisites

### Description
Addresses #1594 This can be merged or #2456 instead.

### DIFF
--- a/src/Nancy/NancyEngine.cs
+++ b/src/Nancy/NancyEngine.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.IO;
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
@@ -252,7 +253,19 @@
                 return;
             }
 
-            handler.Handle(context.Response.StatusCode, context);
+            try
+            {
+                handler.Handle(context.Response.StatusCode, context);
+            }
+            catch (Exception ex)
+            {
+                if (defaultHandler == null)
+                {
+                    throw;
+                }
+
+                defaultHandler.Handle(context.Response.StatusCode, context);
+            }
         }
 
         private async Task<NancyContext> InvokeRequestLifeCycle(NancyContext context, CancellationToken cancellationToken, IPipelines pipelines)
--- a/test/Nancy.Tests/Unit/NancyEngineFixture.cs
+++ b/test/Nancy.Tests/Unit/NancyEngineFixture.cs
@@ -340,6 +340,53 @@ namespace Nancy.Tests.Unit
         }
 
         [Fact]
+        public async Task Should_catch_exception_inside_status_code_handler_and_return_default_error_response()
+        {
+            // Given
+            A.CallTo(() => this.requestDispatcher.Dispatch(A<NancyContext>.Ignored, A<CancellationToken>.Ignored))
+                .Returns(new Response() { StatusCode = HttpStatusCode.InternalServerError });
+
+            var statusCodeHandlers = new[]
+            {
+                this.statusCodeHandler,
+                new DefaultStatusCodeHandler(this.negotiator, this.environment),
+            };
+
+            var engine =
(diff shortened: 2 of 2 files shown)


src/Nancy/NancyEngine.cs
```diff
@@ -252,7 +253,24 @@ private void CheckStatusCodeHandler(NancyContext context)
...
             }
 
-            handler.Handle(context.Response.StatusCode, context);
+            try
+            {
+                handler.Handle(context.Response.StatusCode, context);
+            }
+            catch (Exception ex)
+            {
+                context.Response = new Response()
```
khellang: Instead of doing this, we should call `defaultHandler.Handle` so we get the beautiful Nancy error page instead of just a simple string.

khellang:
And also... we need a test here, I think.

src/Nancy/NancyEngine.cs
```diff
@@ -252,7 +253,21 @@ private void CheckStatusCodeHandler(NancyContext context)
...
+            }
+            catch (Exception ex)
+            {
+                if (defaultHandler != null)
+                {
+                    defaultHandler.Handle(context.Response.StatusCode, context);
+                }
+                else
+                {
+                    throw new Exception("No default status code handler found", ex);
```
khellang: `throw;`

src/Nancy/NancyEngine.cs
```diff
@@ -252,7 +253,21 @@ private void CheckStatusCodeHandler(NancyContext context)
...
             }
 
-            handler.Handle(context.Response.StatusCode, context);
+            try
+            {
+                handler.Handle(context.Response.StatusCode, context);
+            }
+            catch (Exception ex)
+            {
+                if (defaultHandler != null)
```
khellang: Can you flip this condition and remove the `else` clause?

src/Nancy/NancyEngine.cs
jchannon: not throw a new one with the original in tact?

khellang: Nah, that just obfuscates the real exception with a more generic one...

jchannon: Doesnt it add more info to why this was thrown and keep the original too?

khellang: Not really. I just mangles the stack trace. The fact that a status code handler wasn't found isn't the source of the exception. That's fine.

test/Nancy.Tests/Unit/NancyEngineFixture.cs
```diff
@@ -340,6 +340,45 @@ public void HandleRequest_should_get_context_from_context_factory()
...
 
         [Fact]
+        public async Task Should_catch_exception_inside_status_code_handler_and_return_default_error_response()
+        {
+            // Given
+            A.CallTo(() => this.requestDispatcher.Dispatch(A<NancyContext>.Ignored, A<CancellationToken>.Ignored))
+                .Returns(new Response() { StatusCode = HttpStatusCode.InternalServerError });
+
+            var engine =
+                new NancyEngine(this.requestDispatcher, this.contextFactory, new[] { this.statusCodeHandler, new DefaultStatusCodeHandler(this.negotiator, this.environment), }, A.Fake<IRequestTracing>(), new DisabledStaticContentProvider(), this.negotiator, this.environment)
```
khellang: Just one last nit; can we shorten this line down a bit? Either wrapping it or making the status code handler array a variable? Or both?

jchannon: I shortened it. this is the only test that has this line, if we have 2+ then I think it'd make sense for a variable

khellang: A variable, not a field. The `request` variable below is also just used one time. :smile:

jchannon: Done 😄

PR #2458, opened 2016-05-26
Updated the look and feel of the error pages

thecodejunkie:
### Prerequisites

### Description
This is something that I have been wanting to do for a long time now. The old pages, while they've generated a lot of positive reactions, have felt a bit dated to me for a while. I've been fiddling around with a new, modern, look for the error pages for the 2.0 release. The pages are all in HTML5, uses things like CSS3 to create zebra stripes in PRE-tags (but will have a fallback for non compliant browsers). 

### Pages that needs updating
- [X] 404
- [X] 500
- [X] Razor
- [X] DotLiquid
- [X] Formatting of the `ViewNotFoundException` message

### Screenshots

image

image

image

Sample of when error traces are disabled (same applies to the 500 page)

image

Updated `ViewNotFoundException` message formatting to make it more readable

image


phillip-haydon:
These are looking awesome!

src/Nancy/ViewEngines/ViewNotFoundException.cs
```diff
@@ -29,14 +30,14 @@ public ViewNotFoundException(string viewName, string[] availableViewEngineExtens
...
 
-            this.message = String.Format(
-                    "Unable to locate view '{0}'{4}Currently available view engine extensions: {1}{4}Locations inspected: {2}{4}Root path: {3}{4}" +
-                    "If you were expecting raw data back, make sure you set the 'Accept'-header of the request to correct format, for example 'application/json'",
-                    this.ViewName,
-                    string.Join(",", this.AvailableViewEngineExtensions),
-                    string.Join(",", this.InspectedLocations),
-                    this.rootPathProvider.GetRootPath(),
-                    Environment.NewLine);
+            this.message = string.Format(
```
khellang: `StringBuilder`? Or do we care? There's a ton of `string.Join`, `string.Concat` and `string.Format` here...

jchannon:
Apart from @khellang point :shipit:

src/Nancy/ViewEngines/ViewNotFoundException.cs
phillip-haydon: I don't care. It's an exception. Would be worried if this was happy path code.

06b:
What if the Tumbeast were removed and replaced with the Nancy logo?

psibernetic:
Being able to bootstrap config the ErrorPage Img and funny text for a more professional feel would be simple but darn good improvements, me thinks.

thecodejunkie:
@psibernetic you can just swap out the `DefaultStatusCodeHandler` or use your own, for instance like this http://mike-ward.net/2013/04/11/custom-error-pages-in-nancyfx/

psibernetic:
@thecodejunkie I was wondering how to replace them too so thanks for that, but personally I'd like to keep all the good work you have done with zebra stripes and such, just rebrand a bit. 

It is just a thought. I could always just copy your files from the repo and modify them too, just more work and no more inline updates.

BrandonClapp:
Awesome job. The new error pages will look great!

06b:
Is it possible to move the css into an external style sheet so the styles could be maintained in one location?

phillip-haydon:
@06b if we move it to an external stylesheet then installing nancy would require including that style sheet. 

The code is on github for you to take and modify to your requirements, but I'm against trying to pull this out, the beauty of this error page is it's all contained within the assembly and pushed out giving a nice appearance.

06b:
@phillip-haydon Ahhh I see... In that case - I'm against my own suggestion about moving it into an external style sheet as well.

src/Nancy/ViewEngines/ViewNotFoundException.cs
thecodejunkie: Yeah I don't really care either.. it should be a rare thing to end up here, even more so in production

PR #2512, opened 2016-07-19
Replace dependency on RequestStream in Request.Body with Stream

danbarua:
### Prerequisites

- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open -  overlaps with #1920

### Description
Use case:
We're uploading large blob files via a Nancy API
The default `RequestStream` implementation spools out to disk where the size of the incoming request is >85kb to avoid pressure on the Large Object Heap
When requests are aborted, or generally when errors occur, these temp files can leak and cause %TEMPDIR% to run out of space. (I have submitted a few PRs around this in the past but it seems we can't avoid this behaviour, we can't clean up the file in the `RequestStream` destructor as in the finalizer we have lost the reference to the `File`)
Since `RequestStream` is a drop-in-replacement for `Stream`, it seems logical to use the base class as the `Body` property on `Request`, this will allow me to swap in a different `Stream` implementation such as https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream
This also includes tracing so I can troubleshoot where Nancy may be leaking the `Request.Body` stream.
I can't inherit from/override `RequestStream` since it has no paramaterless ctor, and this would be a cleaner approach anyhow.

Warning: possible conflicts with #1920

### DIFF
--- a/src/Nancy/Request.cs
+++ b/src/Nancy/Request.cs
@@ -47,7 +47,7 @@ namespace Nancy
         /// <param name="protocolVersion">The HTTP protocol version.</param>
         public Request(string method,
             Url url,
-            RequestStream body = null,
+            Stream body = null,
             IDictionary<string, IEnumerable<string>> headers = null,
             string ip = null,
             X509Certificate certificate = null,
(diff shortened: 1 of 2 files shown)


danbarua:
With this change, I can swap out `RequestStream` with another implementation here: https://github.com/NancyFx/Nancy/blob/a9989a68db1c822d897932b2af5689284fe2ceb8/src/Nancy/Owin/NancyMiddleware.cs#L84

danbarua:
I've reviewed #1920 and I'm fairly sure this change will have no impact on that PR.

jchannon:
By making Request use Stream and not RequestStream won't that undo the work put into RequestStream.cs?

jchannon:
This looks fine to me, we will still use RequestStream but others can use another inherited Stream class and write their own Middleware extension to use that

danbarua:
Assuming this is ok to merge, any chance of getting it backported to 1.x?

jchannon:
Pinging @thecodejunkie

horsdal:
Agree that this is good to go. Merging

PR #2520, opened 2016-07-28
Added enum serializing/deserializing as string

jchannon:
### Prerequisites

### Description
This serializes/deserialzies a enum to string/string to enum

I previously used ServiceStack.Text which serilaizes enums as strings and theforefore moving to netcore this broke my application as SSTack does not work on coreCLR.

### DIFF
--- a/test/Nancy.Tests/Unit/Json/SimpleJsonFixture.cs
+++ b/test/Nancy.Tests/Unit/Json/SimpleJsonFixture.cs
@@ -1,5 +1,6 @@
 ﻿namespace Nancy.Tests.Unit.Json
 {
+    using System.Globalization;
     using Nancy.Json.Simple;
     using Xunit;
 
@@ -31,6 +32,31 @@
 
             //Then
             actual.ShouldEqual(@"42");
+        }
+
+        [Fact]
+        public void Should_serialize_enum_to_string()
+        {
+            //Given
+            var model = new ModelTest { EnumModel = TestEnum.Freddy };
+
+            //When
+            var result = SimpleJson.SerializeObject(model, new NancySerializationStrategy(false, true));
+
+            //Then
+            result.ShouldEqual("{\"enumModel\":\"Freddy\"}");
+        }
+
+        public class ModelTest
+        {
+            public TestEnum EnumModel { get; set; }
+        }
+        
+        public enum TestEnum
+        {
+            Rod = 1,
+            Jane = 2,
+            Freddy = 3
         }
     }
 }
--- a/src/Nancy/Json/JavaScriptSerializer.cs
+++ b/src/Nancy/Json/JavaScriptSerializer.cs
@@ -61,7 +61,7 @@ namespace Nancy.Json
         {
             this.jsonConfiguration = jsonConfiguration;
             this.globalizationConfiguration = globalizationConfiguration;
-            this.serializerStrategy = new NancySerializationStrategy(jsonConfiguration.RetainCasing);
+            this.serializerStrategy = new NancySerializationStrategy(jsonConfiguration.RetainCasing, jsonConfiguration.SerializeEnumToString);
         }
 
         /// <summary>
--- a/src/Nancy/Json/JsonConfiguration.cs
+++ b/src/Nancy/Json/JsonConfiguration.cs
@@ -17,7 +17,8 @@ namespace Nancy.Json
             Converters = new List<JavaScriptConverter> { new TimeSpanConverter(), new TupleConverter() },
             DefaultEncoding = Encoding.UTF8,
             PrimitiveConverters = new List<JavaScriptPrimitiveConverter>(),
-            RetainCasing = false
+            RetainCasing = false,
+            SerializeEnumToString = false
         };
 
         private JsonConfiguration()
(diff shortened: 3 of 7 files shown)


khellang:
This is a major breaking change, no?

jchannon:
I'm not sure what v1 does

danbarua:
Might be a major breaking change but returning enums as ints == wat.

Yantrio:
it is going from 1.x to 2.x, so semver allows for breaking changes :)

khellang:
Oh, I agree that strings are nicer. We just have to make sure that the other serializers default to the same.

jchannon:
Isn't that the point of having separate serializers? They work differently and separately?

khellang:
Yes and no. I still think we should strive to have the same defaults, i.e. camelCase, string enums, omit null-values etc.

jchannon:
I'm sure @danbarua will do that for us.

#mergethismotha

danbarua:
Review my pull request and I'll think about it :)

test/Nancy.Tests/Unit/Json/SimpleJsonFixture.cs
```diff
@@ -32,5 +33,42 @@ public void Integer_dictionary_values_are_Json_serialized_as_integers()
...
+
+            //Then
+            result.EnumModel.ShouldEqual(TestEnum.Freddy);
+        }
+
+        public class ModelTest
+        {
+            public TestEnum EnumModel { get; set; }
+        }
+        public enum TestEnum
```
thecodejunkie: Missing blank line ;)

thecodejunkie:
Couple of thoughts

1. In principal I am happy with us taking this in and breaking the behavior for 2.x but I think that @khellang is correct and that we should make it a consistent behavior between our `Nancy.Serialization.*` projects
2. There is _zero_ context in this pull-request as to _why_ this change was proposed in the first place. Not comfortable taking a pull-request without knowing why we should introduce a breaking change.

thecodejunkie:
image

Typo! 😄

jchannon:
1. Maybe we could add some UpForGrabs issues in the Json.Net serializer and ServiceStack repos to make sure it behaves as this does, ISSTack already does this (although I think we should kill that as it wont work on core and do many people still use it?) but JSON.Net serializes as integers

2. Context added

thecodejunkie:
So the reason we're looking at introducing a breaking change is not because it is the most sensible change, but rather that one of your dependencies cannot be used anymore so you propose a change to the build in serializer so it works the same as said dependency? 😄 Not sure that's a very good reason to introduce a breaking change to be honest. Now is we can agree that `strings` are preferred over `int` either way, then we're approaching it from a reasonable angle 😄

jchannon:
I think comments do agree that strings are better

thecodejunkie:
ping @NancyFx/most-valued-minions @NancyFx/owners are we happy on changing the default to string values instead of int? If we do this, then I agree with @khellang that the other `Nancy.Serialization.*` projects should be updated to have the same default behavior (they should have the same defaults for as much as possible).

phillip-haydon:
It should support both. I use ints so this is a breaking change for me.

jchannon:
Anyone else have an opinion?

danbarua:
What the fuck is wrong with Philip Haydon is my opinion, but probably best expressed elsewhere.

phillip-haydon:
I've never worked on a project where enums were not mapped in the database as a table so as to keep the size of the table smaller and contributing to page spilling. While being mapped in code via an enum. So all UI requests are, a drop down list built from an enum passing the int value.


Weather this is the right or wrong approach is up for debate, but doesn't change that it's very common.


IMO it's bad to not support ints by default.

src/Nancy/Json/Simple/SimpleJson.cs
```diff
@@ -1383,6 +1383,17 @@ public virtual object DeserializeObject(object value, Type type, DateTimeStyles
             
             bool valueIsLong = value is long;
             bool valueIsDouble = value is double;
+            if (type.GetTypeInfo().IsEnum)
+            {
+                 if (value is double || value is int || value is long)
+                 {
+                     return Enum.ToObject(type, Convert.ToInt32(value.ToString()));
```
phillip-haydon: You check for `long` and cast to `int32`...

jchannon: good spot, just copy/pasted from simplejson repo

thecodejunkie:
We could (should?) support both.. the question is should we change the default. Perhaps not. Perhaps enable support to change how enums are serialized and keep the default as int. That will have the smallest impact? I guess your project decides which is the more appropriate to use. If you map them to DB values, like @phillip-haydon says then int makes sense.. if you spit them out into an external API then the strings might be more useful.. kinda guessing that no matter what we choose as the default, we'll be screwing over the other 50% .. hence leaving int as default, but letting you change it to string (or some other strategy I guess) of you have that requirement

phillip-haydon:
Right now this is supporting deserialization only, so converting the JSON to the object, which will support both when @jchannon fixes his bug. 

In regards to serialization, if we make the change for 2.0 then I'm fine as long as I can switch it back to int, but we really should make this decision before RTM 2.0

phillip-haydon:
Oh woops, I missed the `SerializeEnum` method. Yeah ok, we should have config around that!

thecodejunkie:
Yeah this is part of the `2.0-clinteastwood` milestone and we should be closing that by next week

jchannon:
Added config option.

jchannon:
**MERGE ALERT**

thecodejunkie:
@jchannon is that not a proper failed test?

jchannon:
It only fails on TC

PR #2534, opened 2016-08-11
Remove Wait call to prevent deadlock

khellang:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open
- [x] I have verified that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)
- [x] I have provided test coverage for my change (where applicable)

### Description

The call to `Task.Wait()` causes a deadlock on many concurrent requests. This changes the method to do a synchronous copy instead.

Fixes #2533

Issue #2471, opened 2016-05-30
Should ConfigureAwait(false) be called in NancyHost?


xt0rted:
While looking through the code and using the ReSharper [ConfigureAwait Checker](https://github.com/aelij/ConfigureAwaitChecker/) plugin I came across [Nancy.Hosting.Self/NancyHost.cs#L133-134](https://github.com/NancyFx/Nancy/blob/bc44cd30d58ce1b6c7930452304d187248038e18/src/Nancy.Hosting.Self/NancyHost.cs#L133-134).

Should line 133 and/or 134 call `.ConfigureAwait(false)`? The other warning I'm getting is on line 134 because it's not awaited but I wasn't sure if that should be or not.

Those are the only two spots outside of the test projects that don't call ConfigureAwait.

psibernetic:
Since NancyHost is the consumer of the task it does need to capture context.

thecodejunkie:
@khellang @damianh @grumpydev

damianh:
Probably should be called from everywhere.

Explicit or assembly rewrite with a Fody plugin?

On 25 Aug 2016 12:51 p.m., "Andreas Håkansson" <notifications@github.com>
wrote:

@khellang <https://github.com/khellang> @damianh
<https://github.com/damianh> @grumpydev <https://github.com/grumpydev>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/NancyFx/Nancy/issues/2471#issuecomment-242359437>, or mute
the thread

horsdal:
I think line 133 should have a CAF.
Basically every `await` outside of tests should have a CAF AFAIK.

thecodejunkie:
Been out of the loop a bit, but would adding that `ConfigureAwait` call fix #2534 ?

jchannon:
What if you host by something else?

thecodejunkie:
Well Nancy is a library and if you consume it yourself you are subjected to following the laws of async, this making sure you properly do your own `ConfigureAwait`-handling. Or am I thinking about this in the wrong way?

damianh:
You are thinking about this in the correct way :)

thecodejunkie:
@damianh so we should get this in, right?

damianh:
Yep!

thecodejunkie:
Oh wait.. this is an issue, not a pull-request.. @xt0rted mind sending a PR and I'll get it in asap?

PR #2535, opened 2016-08-13
Setup code analysis

xt0rted:
This is based on the discussion in #2534 and sets up code analysis with a custom ruleset that includes [DotNetAnalyzers/AsyncUsageAnalyzers](https://github.com/DotNetAnalyzers/AsyncUsageAnalyzers). All the other rules are turned off right now.

In the classic project setup you can view the code analysis rules in the project's properties and when building the errors will show in the Error List.

In the new project setup there's no code analysis settings in the project's properties, but it does allow you to detect and run the rules. You do that by right clicking on the solution/folder/project and selecting `Find Code Issues` which should bring up the Inspection Results window which is similar to the old Error List.

I'm not sure if the new project setup can run these on build or in a separate step, or if we'll have to wait for the rollback to msbuild for that.

The results of this show the only issue being in [NancyHost.cs#L133-L134](https://github.com/NancyFx/Nancy/blob/d76e7ef6ee938c8ac1f3c42b50b610c158fdeccc/src/Nancy.Hosting.Self/NancyHost.cs#L133-L134) as per #2471


jchannon:
Is this whole thing a VS feature ie. Can you get analysis from dotnet build?

xt0rted:
I just turned on [CS4014](https://msdn.microsoft.com/en-us/library/hh873131.aspx) and that's giving two new warnings. One is in NancyHost.cs, the other is [DiagnosticsHook.cs#L186](https://github.com/NancyFx/Nancy/blob/d76e7ef6ee938c8ac1f3c42b50b610c158fdeccc/src/Nancy/Diagnostics/DiagnosticsHook.cs#L186)

@jchannon the msbuild output has these warnings and errors, you can see them in the [teamcity build log](http://nancy-ci.cloudapp.net/viewLog.html?buildId=7476&buildTypeId=Nancy_Pulls&tab=buildLog#_state=47,177&focus=497) (search for `CS4014`). I'm not sure why the async usage ones I just added aren't showing up there though.

I'm also not sure about dotnet build since I've been waiting for them to stop changing every release before I started looking into it.

jchannon:
But interestingly it doesn't show an issue around the code in the PR that
@khellang did

xt0rted:
It looks like that code isn't being detected by these rules because of how it's invoked. Since it's in a constructor and is using `.Wait()` instead of `await` you can't call `.ConfigureAwait(false)` so it's not being flagged.

xt0rted:
After looking into this some more it seems the analyzers don't work with .net core. There's a number of issues about this accross the various dotnet & roslyn projects which seem to indicate it's a tooling issue and/or clr issue and should be fixed in the 1.1 release.

thecodejunkie:
So what do we want to do with this?

thecodejunkie:
@xt0rted is this still in limbo? If so then we'll push it forward to the next milestone. If not then it needs rebasing (and updating?) so we can pull it in. Trying to wrap up the two remaining issues on the `-clinteastwood` milestone :)

xt0rted:
@thecodejunkie this can wait, I need to look into why the build is failing after my rebase but I don't really have the time right now.

xt0rted:
@thecodejunkie rebased and passing

PR #2537, opened 2016-08-13
Correctly accept all json and xml subtypes

lvermeulen:
### Prerequisites

### Description
Currently Nancy accepts all **vendor-specific** json and xml media subtypes. As per [RFC 6839 (json)](https://tools.ietf.org/html/rfc6839#page-4) and [RFC 3023 (xml)](https://tools.ietf.org/html/rfc3023#page-32) all +json and +xml subtypes should be regarded as such. Otherwise many [IANA registered media types](http://www.iana.org/assignments/media-types/media-types.xhtml) will not be handled correctly.

### DIFF
--- a/src/Nancy/Responses/Negotiation/JsonProcessor.cs
+++ b/src/Nancy/Responses/Negotiation/JsonProcessor.cs
@@ -107,8 +107,7 @@
 
             var subtypeString = requestedContentType.Subtype.ToString();
 
-            return (subtypeString.StartsWith("vnd", StringComparison.OrdinalIgnoreCase) &&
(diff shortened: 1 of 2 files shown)


khellang:
:shipit:

Anyone know why TC is failing?

lvermeulen:
@khellang It seems to be related to a roslyn bug, it's referenced [here](https://github.com/xunit/xunit/issues/646), [here](https://github.com/dotnet/roslyn/issues/10718) and [here](https://github.com/dotnet/roslyn/pull/9963).
Suggestion is to use the CDSCollectionETWBCLProvider event source directly (rather than indirectly), e.g. by referencing ConcurrentDictionary. **If** I understood correctly. :-)

PR #2547, opened 2016-08-30
Fix for issue-2546: Allow chunked transfer-encoding for ASP.Net hosted requests

yahehe:
### Prerequisites
### Description

This allows incoming requests with a header of Transfer-Encoding = chunked to properly set the Request Body

### DIFF
--- a/src/Nancy.Hosting.Aspnet/NancyHandler.cs
+++ b/src/Nancy.Hosting.Aspnet/NancyHandler.cs
@@ -98,7 +98,7 @@ namespace Nancy.Hosting.Aspnet
 
             RequestStream body = null;
 
-            if (expectedRequestLength != 0)
+            if (expectedRequestLength != 0 || HasChunkedEncoding(incomingHeaders))
             {
                 body = RequestStream.FromStream(context.Request.InputStream, expectedRequestLength, StaticConfiguration.DisableRequestStreamSwitching ?? true);
             }
@@ -112,6 +112,17 @@ namespace Nancy.Hosting.Aspnet
                 context.Request.UserHostAddress, 
                 certificate,
                 protocolVersion);
+        }
+
+        private static bool HasChunkedEncoding(IDictionary<string, IEnumerable<string>> incomingHeaders)
+        {
+            IEnumerable<string> transferEncodingValue;
+            if (incomingHeaders == null || !incomingHeaders.TryGetValue("Transfer-Encoding", out transferEncodingValue))
+            {
+                return false;
+            }
+            var transferEncodingString = transferEncodingValue.SingleOrDefault() ?? string.Empty;
+            return transferEncodingString.Equals("chunked", StringComparison.OrdinalIgnoreCase);
         }
 
         private static long GetExpectedRequestLength(IDictionary<string, IEnumerable<string>> incomingHeaders)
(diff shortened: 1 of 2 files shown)


src/Nancy.Hosting.Aspnet/NancyHandler.cs
```diff
@@ -114,6 +114,18 @@ private static Request CreateNancyRequest(HttpContextBase context)
                 protocolVersion);
         }
 
+        private static bool IsChunkEncoded(IDictionary<string, IEnumerable<string>> incomingHeaders)
```
khellang: `HasChunkedEncoding`?

src/Nancy.Hosting.Aspnet/NancyHandler.cs
```diff
@@ -114,6 +114,18 @@ private static Request CreateNancyRequest(HttpContextBase context)
...
         }
 
+        private static bool IsChunkEncoded(IDictionary<string, IEnumerable<string>> incomingHeaders)
+        {
+            if (incomingHeaders == null)
+            {
+                return false;
+            }
+            IEnumerable<string> transferEncodingValue;
+            var isTransferEncoded = incomingHeaders.TryGetValue("Transfer-Encoding", out transferEncodingValue);
```
khellang: Can we just bail out early here?

```csharp
if (!incomingHeaders.TryGetValue("Transfer-Encoding", out transferEncodingValue))
{
    return false;
}
```

src/Nancy.Hosting.Aspnet/NancyHandler.cs
```diff
@@ -114,6 +114,18 @@ private static Request CreateNancyRequest(HttpContextBase context)
...
+            var transferEncodingString = isTransferEncoded ? transferEncodingValue.SingleOrDefault() : string.Empty;
+            return transferEncodingString != null && transferEncodingString.ToLower().Equals("chunked");
```
khellang: Let's compare this using proper casing, instead of doing `ToLower` and allocating an extra string.

```csharp
string.Equals(transferEncodingString , "chunked", StringComparison. OrdinalIgnoreCase)
```

yahehe:
Updated the PR with the suggestions. Two of the build checks are failing but as far as I can tell they aren't related to this?

thecodejunkie:
@khellang got a moment to revisit this? Your initial review comments have been fixed

khellang:
I think this looks good (besides the failing builds)

thecodejunkie:
The failing build it due to some issue with Travis =/

thecodejunkie:
Shouldn't this primarily go into `master` instead of being back-ported to `1.x`? @khellang

yahehe:
I'd be fine making another PR to fix this in master if the issue exists there as well. I made this initially because my team was using 1.4.3 at the time

thecodejunkie:
@yahehe yeah it should be.. no changes should have been made to that. For `1.x` we really just want to put out bug fixes since it has been put into maintenance mode

yahehe:
Ok, I'll make the same change against master when I get a chance then (you can decline this if it's not enough of a bug-fix)

yahehe:
@jnallard are you still using the version of Nancy.1.4.3 I made this change for?

jnallard:
@yahehe Nope, I'm using Nancy 2 for all my projects.

yahehe:
Closing so I can figure out what I have to do to make the same functional change into v2

PR #2548, opened 2016-08-30
Add missing await and add ConfigureAwait(false) calls

xt0rted:
### Prerequisites

### Description

This is a fix for #2471 along with some missing `this.`'s.

I wasn't sure if the call to `Process()`  on line 134 was supposed to be awaited or not. Trying out the self host demo worked but that was as far as I went (I'm not familiar with the self host project).

### DIFF
--- a/src/Nancy.Hosting.Self/NancyHost.cs
+++ b/src/Nancy.Hosting.Self/NancyHost.cs
@@ -131,7 +131,7 @@
                     while(!this.stop)
                     {
                         HttpListenerContext context = await this.listener.GetContextAsync().ConfigureAwait(false);
-                        Process(context);
+                        await this.Process(context).ConfigureAwait(false);
                     }
                 }
                 catch(Exception ex)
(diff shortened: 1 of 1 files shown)


jchannon:
👍 Thanks ❤️

davidallyoung:
@jchannon Does this make it so that the nancy self host will really only process one request at a time since it awaits on the Process call? Previously it would have just executed a hot task and forgotten about it and went back into the GetContextAsync loop. Does the fact that the httpListner.GetContextAsync is returned in its own execution thread change how await works?

Code referenced: https://github.com/NancyFx/Nancy/commit/4c48f54262e3449561a0800a500c73ec8ac3fda9#diff-f6b50509473947354cba195b3180ca10R134

jchannon:
ping @xt0rted @NancyFx/most-valued-minions

jchannon:
@davidallyoung No I dont believe so, its just saying it doesn't need the original context in which its executed

davidallyoung:
@jchannon Yeah I think it's good that the call to GetContextAsync was CAF'd, but I really feel like the await on this.Process is going to prevent the NancyHost from processing more than one request at a time as it'll block loop execution until Process completes. The current release of SelfHosting (1.4.1) package is using the async event callbacks on the httpListener so it's not affecting consumers of the package. I can submit a PR removing the await if that's a better way to have the conversation, just let me know!

PR #2555, opened 2016-09-08
Fix for deserializing JSON number to ulong and ushort values

kal:
### Prerequisites

### Description

I found an issue with binding to a model that has a ulong property on it which showed up as a cast exception. Tracing through I found a simple case of a missing cast from the default long value that a number is parsed to to the requested ulong type. While in there I noticed that the same was true for ushort values.

This PR includes a couple of short unit tests for the deserialization of JSON numbers to ulong and ushort values as well as the fix.


kal:
I don't think that the TC build failure is the result of this PR as the master branch is currently also failing on the TC build.

jchannon:
Thanks! ❤️

PR #2577, opened 2016-09-28
WIP - First stab at creating a BodyDelegate type

grumpydev:
### Prerequisites

### Description

Initial stab at pulling out the body delegate to its own type to allow for casts to/from the actual async and non-async delegate signatures - idea being we patch up everything by adding a cast to Action<Stream> to make it build (unfortunately implicit casts and delegates still need an explicit cast), then we can work through bit by bit making them async, then remove the cast option from the sync delegate type at a later date.

Already fixed up the Nancy project itself, which shows both "asyncing" the bits that make sense, and just straight casting the bits that don't (such as when writing a string).

### DIFF
--- a/src/Nancy.MSBuild/BodyDelegate.cs
+++ b/src/Nancy.MSBuild/BodyDelegate.cs
@@ -1,0 +1,31 @@
+namespace Nancy
+{
+    using System;
+    using System.IO;
+    using System.Threading.Tasks;
+
+    public class BodyDelegate
+    {
+        public Func<Stream, Task> Body { get; set; }
+
+        public static implicit operator BodyDelegate(Func<Stream, Task> body)
+        {
+            return new BodyDelegate { Body = body };
+        }
+
+        public static implicit operator BodyDelegate(Action<Stream> body)
+        {
+            return new BodyDelegate { Body = Wrap(body) };
+        }
+
+        private static Func<Stream, Task> Wrap(Action<Stream> body)
+        {
+            return s =>
+            {
+                body.Invoke(s);
+
+                return Task.FromResult(new object());
+            };
+        }
+    }
+}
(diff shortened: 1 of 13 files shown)


src/Nancy.MSBuild/BodyDelegate.cs
```diff
@@ -0,0 +1,31 @@
+namespace Nancy
```
jchannon: This is in the wrong location, needs to be under Nancy

src/Nancy.MSBuild/Nancy.csproj
```diff
@@ -1376,6 +1376,7 @@
     <Compile Include="..\Nancy\Routing\Route.cs">
       <Link>Routing\Route.cs</Link>
     </Compile>
+    <Compile Include="BodyDelegate.cs" />
```
jchannon: `../Nancy` required

src/Nancy.MSBuild/BodyDelegate.cs
grumpydev: *shrug* VS2015 is garbage, what can I say? :)

thecodejunkie:
@grumpydev @jchannon yeah unfortunately there is a bit of fud right now if you add new files.. due to and old bug with VS tooling when `project.json` and `*.csproj` files existed side by side https://github.com/dotnet/corefx/issues/4544 I'm trying to figure out if it has been resolved. If it has then we should revert the `<projectname>.MSBuild` folder change we made as a workaround. For the time being we manually have to move the files to the correct folder and update the `*.csproj` to link them

thecodejunkie:
@grumpydev can you rebase plz?

thecodejunkie:
👀 rebase, pretty please? 😄

PR #2605, opened 2016-10-30
Fix a memory leak

yzhou88:
### Prerequisites

- [ ] I have written a descriptive pull-request title
- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open
- [ ] I have verified that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)

### Description

Dispose the object returned by CreateLinkedTokenSource(...)

I have spent around 6 hours working on memory leak issue. After applying the change here, no memory leak anymore.

### DIFF
--- a/src/Nancy/NancyEngine.cs
+++ b/src/Nancy/NancyEngine.cs
@@ -95,7 +95,8 @@
 
         public Task<NancyContext> HandleRequest(Request request, Func<NancyContext, NancyContext> preRequest, CancellationToken cancellationToken)
         {
-	        var cts = CancellationTokenSource.CreateLinkedTokenSource(this.engineDisposedCts.Token, cancellationToken);
+	        using(var cts = CancellationTokenSource.CreateLinkedTokenSource(this.engineDisposedCts.Token, cancellationToken))
+            {
             cts.Token.ThrowIfCancellationRequested();
 
             var tcs = new TaskCompletionSource<NancyContext>();
@@ -138,10 +139,6 @@
 		                tcs.SetException(ex);
 		                return;
 	                }
-	                finally
-	                {
-		                cts.Dispose();
-	                }
 
                     tcs.SetResult(completeTask.Result);
                 },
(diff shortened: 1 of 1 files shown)


thecodejunkie:
Let's loop in @khellang and this one as he worked on it previously. Thanks

yzhou88:
I understand this issue does not exist in v2.00. But can you update v1.4.3 so Nuget can give me a bug-free 1.x version ?

thecodejunkie:
Oh sorry, you're sending this to the `1.x` branch. I'll have a look

yzhou88:
Wow, I really like to contribute to this project, we have such an active people like you!
When stress testing my project, the memory leak became a serious problem. Please understand that this simple fix at least cost me 6 hours. And I'd like to see v1.4.4 just to have this issue fixed. 

Thanks!

khellang:
This will just bring back the big that https://github.com/NancyFx/Nancy/pull/2150 tried to fix. We need to figure out exactly what and when we're leaking memory here...

jeffthiele-iq:
Looking at the original fix for #2150 it looks like I may have missed a `cts.Dispose();` in the error task for `WhenCompleted()` potentially causing a memory leak when errors occur.  I wonder if that would solve the memory leaks you're seeing.

khellang:
I think this is worth fixing for a new 1.4.x release of Nancy, along with #2616. What's the next step here? We certainly won't bring back the behavior of this PR as it currently stands.

cloudhunter89:
> Looking at the original fix for #2150 it looks like I may have missed a cts.Dispose(); in the error task for WhenCompleted() potentially causing a memory leak when errors occur.

@jeffthiele-iq What about the case of serving static content?  It seems that the `CancellationTokenSource cts` is not disposed of in that case either, or am I missing something else?

jeffthiele-iq:
Yeah you're right @cloudhunter89, it isn't disposed correctly for static content.  None of my testing used static content.

PR #2648, opened 2016-12-09
Build script cleanup

thecodejunkie:
- Updated the way that the `UpdateProjectJsonVersion`, in `build.cake` updates the version number. It will now only update the first occurrence (which should always be the project one)
- Updated the `Publish-NuGet` task so that it filters out `*.symbols.nupkg`
- Some other white-space / indentation tweaks

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -196,23 +196,24 @@ Task("Package-NuGet")
 });
 
 Task("Publish-NuGet")
-  .Description("Pushes the nuget packages in the nuget folder to a NuGet source. Also publishes the packages into the feeds.")
-  .Does(() =>
+    .Description("Pushes the nuget packages in the nuget folder to a NuGet source. Also publishes the packages into the feeds.")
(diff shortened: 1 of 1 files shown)


build.cake
```diff
@@ -292,36 +293,41 @@ Task("Prepare-Release")
...
+    foreach (var file in filePaths)
+    {
+        var project =
+            System.IO.File.ReadAllText(file.FullPath, Encoding.UTF8);
 
-    System.IO.File.WriteAllText(file.FullPath, project, Encoding.UTF8);
-  }
-}
+        var projectVersion =
+            new System.Text.RegularExpressions.Regex("(\"version\":\\s*)\".+\"");
```
jchannon: you tested this?

thecodejunkie: Yes? I have a habit of running code that I change, before I pull-request it 😄 I've not modified the regex

PR #2654, opened 2016-12-16
AppDomainAssemblyCatalog correctly unloads the assemblies that were probed for any Nancy reference.

arturosevilla:
### Prerequisites

### Description

`AppDomainAssemblyCatalog`, in order to load all the assemblies that have a reference to Nancy, probes for the loaded assemblies as well as those that are in the "load path". The probing occurs by purposefully loading the assembly into a temporal `AppDomain` that is unloaded after the loading of useful assemblies is complete.

However this is accomplished by using `AppDomain.Load()`, and as specified by [MSDN](https://msdn.microsoft.com/en-us/library/36az8x58(v=vs.110).aspx), this is just a convenience method for `Assembly.Load()`. This means that the current code loads all the assemblies into both the inspection `AppDomain` and `AppDomain.Current`, when the code intent was clear that this is not the desired behavior.

Moreover, if the assembly is passed between application domains, because it is not serializable (`MarshalByRefObject`), it would be loaded as well into both domains, therefore the solution is to make a remote proxy probing instance which loads the assembly in the inspection domain and locally determines if the desired reference exists.


arturosevilla:
I believe #2653 is breaking my PR.

thecodejunkie:
You should rebase your PR and it should go away

arturosevilla:
That did it 👍

src/Nancy/Helpers/ProxyNancyReferenceProber.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    using System;
+    using System.Reflection;
+    using Nancy.Extensions;
+
+    internal class ProxyNancyReferenceProber : MarshalByRefObject
+    {
+        public bool HasReference(AssemblyName assemblyNameForProbing, AssemblyName referenceAssemblyName)
+        {
+            var assemblyForInspection = Assembly.Load(assemblyNameForProbing);
```
khellang: Can this be reflection only?

khellang:
This is a really nice PR! :sparkles: Thanks :heart:

src/Nancy/Helpers/ProxyNancyReferenceProber.cs
arturosevilla: Ok, then yes. I tested it and it came out correctly, I just pushed these changes.

src/Nancy/Helpers/ProxyNancyReferenceProber.cs
```diff
@@ -0,0 +1,18 @@
+#if !CORE
+namespace Nancy.Helpers
+{
+    using System;
+    using System.Reflection;
+    using Nancy.Extensions;
+
+    internal class ProxyNancyReferenceProber : MarshalByRefObject
```
arturosevilla: @thecodejunkie I don't see which access modifier you're referring to? Besides making it `public`.

thecodejunkie: I'd rather see it `private` than `internal`

arturosevilla: @thecodejunkie But this is a class defined at the namespace level, you can't have a private class unless it's inside another class. Would you rather have this class put inside the `AppDomainAssemblyCatalog` class?

thecodejunkie:
@khellang care to give this one last round of reviewing before we pull it in?

test/Nancy.Tests.MSBuild/Nancy.Tests.csproj
```diff
@@ -651,6 +652,7 @@
     <Compile Include="..\Nancy.Tests\Unit\Responses\GenericFileResponseFixture.cs">
       <Link>Unit\Responses\GenericFileResponseFixture.cs</Link>
     </Compile>
+    <Compile Include="..\Nancy.Tests\Unit\AppDomainAssemblyCatalogFixture.cs" />
```
khellang: I think this should be referenced like the rest, using a `<Link>` element.

thecodejunkie:
@khellang did you happen to kick the tires on it when you were working on it? Quick sanity check that everything is functioning as expected

khellang:
@thecodejunkie Nope, but I'm assuming we have at least *some* test coverage there. Maybe that should be a billet point for 2.0? Do a coverage scan, see if we have big holes anywhere, and fix those?

thecodejunkie:
@khellang it's mostly a runtime thing so hard to put good tests around it. Improving out coverage is always a good thing 👍

PR #2658, opened 2016-12-22
Remove nuspecs as no longer needed

jchannon:
### Prerequisites

### Description
We use `project.json` now so `*nuspecs` are no longer required


jvandertil:
Seems like a good change, I'm assuming this is also the case with the new csproj format?

thecodejunkie:
@jvandertil Our build scripts use `dotnet pack` to build all NuGet packages these days, so the `*.nuspec` files aren't really used anymore. Once Visual Studio 2017 gets closer to RTM then we will have a look at how we will support the new csproj-format. It's not all black and white, I am afraid, since from what I have gathered they will not back port the tooling to Visual Studio 2015 and we will have to assume that for a while that's what most of our users will continue to support. Of course we want to move over to the new csproj format so with a bit of luck we can support both scenarios without too much hassle (supporting csproj and project.json today was a major PITA because of the tooling support)

PR #2682, opened 2017-01-13
Allows SimpleJson to deserialise POCOs with constructors

JoeStead:
### Prerequisites

### Description

Previously, when deserialising a class that looked like:

```
        public class ParameterisedConstructorEchoModel
        {
            public ParameterisedConstructorEchoModel(string someString, int someInt, bool someBoolean)
            {
                this.SomeString = someString;
                this.SomeInt = someInt;
                this.SomeBoolean = someBoolean;
            }

            public string SomeString { get; set; }
            public int SomeInt { get; set; }
            public bool SomeBoolean { get; set; }

        }
```

SimpleJson would throw a `NullReferenceException` as a default ctor could not be found. This change will try and find a suitable constructor (matching json property names with parameter names), and create the object that way. I've refrained from doing anything fancy with order of parameters etc. to try and keep it simple. 

If no suitable constructor is found, a `ValidConstructorNotFoundException` will be thrown, which should ease with debugging.


thecodejunkie:
Can you rebase please? It will hopefully fix the Travis build error since we've pull in #2685

jchannon:
I assume it still works without constructors?

JoeStead:
Yep, existing implementation has remain unchanged (except a nicer exception being thrown when it can't resolve the type).

jchannon:
@NancyFx/most-valued-minions 2nd pair of eyes please

src/Nancy/Json/Simple/SimpleJson.cs
```diff
@@ -1237,6 +1238,25 @@ interface IJsonSerializerStrategy
         string MapDictionaryKeyToFieldName(string stringKey);
     }
 
+#if SIMPLE_JSON_INTERNAL
```
xt0rted: missing `[GeneratedCode("simple-json", "1.0.0")]` like the other internal/public classes

test/Nancy.Tests.Functional/Tests/SerializeTests.cs
```diff
@@ -45,6 +45,29 @@ public SerializeTests()
...
+            //When
+            var actualModel = response.Body.DeserializeJson<ParameterisedConstructorEchoModel>();
+
+            //Then
+            Assert.Equal("Hi", actualModel.SomeString);
+            Assert.Equal(1, actualModel.SomeInt);
+            Assert.Equal(true, actualModel.SomeBoolean);
+        }
+
+
```
xt0rted: Extra blank line

test/Nancy.Tests.Functional/Tests/SerializeTests.cs
```diff
@@ -72,5 +95,20 @@ public class EchoModel
...
+            {
+                this.SomeString = someString;
+                this.SomeInt = someInt;
+                this.SomeBoolean = someBoolean;
+            }
+
+            public string SomeString { get; set; }
+            public int SomeInt { get; set; }
+            public bool SomeBoolean { get; set; }
+
```
xt0rted: Extra blank line

src/Nancy/Json/Simple/SimpleJson.cs
jchannon: Good spot. No idea what it is though

JoeStead: Are you sure we want to mark non-generated code with a generated code attribute? :)

thecodejunkie:
@JoeStead any progress on the review comments?

JoeStead:
Will remove blank lines, just after some clarification on the attribute
before updating though :)

src/Nancy/Json/Simple/SimpleJson.cs
thecodejunkie: @xt0rted @jchannon why would we want to add that attribute? #context

xt0rted: I'm not sure what it's used for, I just noticed it wasn't included like on the existing classes so I pointed it out incase it was needed for something.

thecodejunkie: > I just noticed it wasn't included like on the existing classes

You've completely lost me. What existing classes? Assume I get _nothing_ about the context that this is discussed (because it's the truth! 😄)

xt0rted: The other classes in this file have the `[GeneratedCode("simple-json", "1.0.0")]` attribute applied to them while the new one didn't. I wasn't sure if this attribute was used for something internally or if it was just metadata for tooling.

thecodejunkie: Oh right, gotcha! Let's ping @prabirshrestha and see if he can shed some light on the `[GeneratedCode("simple-json", "1.0.0")]` attribute and whether or not we should add it to the new class that's embedded in the file

prabirshrestha:
@thecodejunkie if you guys want to take owernship of the original SimpleJson let me know :)

thecodejunkie:
@prabirshrestha not right now, we'll see what the future has to bring 😄 do you have any feedback on the question about the `[GeneratedCode("simple-json", "1.0.0")]` attribute?

prabirshrestha:
SimpleJson was meant to ship as a .cs file instead of a dll which means the project's linting rules applies. In order to avoid this we added that attribute so linting is ignored. In case of Nancy you don't need this attribute since you already ship it as part of dll, But that means you need to change the coding style to match Nancy's stylecop/fxcop/resharper rules. Ideally all tools should respect this attribute.

https://github.com/facebook-csharp-sdk/simple-json/commit/a8a275e2c8bb1470a72cde1268ba84b6ec87b0bd

https://github.com/facebook-csharp-sdk/reflection-utils/pull/7

```
 // This class is meant to be copied into other libraries. So we want to exclude it from Code Analysis rules
 // that might be in place in the target project.
 [GeneratedCode("reflection-utils", "1.0.0")]
```

JoeStead:
In which case, shall I fix the whitespace and we can address the style rules in a different PR?

khellang:
I think you should also be able to add

> // <auto-generated />

To the top of the file to exclude it from linting in VS etc.

JoeStead:
Fixed the whitespace comments. Left everything else. 

I think the SimpleJson stuff could do with a general code-style tidy or mark it all as autogenerated for the linting rules.

thecodejunkie:
@JoeStead we've intentionally left it with the original formatting incase we need to patch it from the original source =)

xt0rted:
@thecodejunkie why are there two copies of SimpleJson and do these types of updates need to be made to both instances?
- [Nancy/Json/SimpleJson.cs](https://github.com/NancyFx/Nancy/blob/ceea69506b9270c94cba231e0bf459897077058b/src/Nancy/Json/SimpleJson.cs)
- [Nancy/Json/Simple/SimpleJson.cs](https://github.com/NancyFx/Nancy/blob/ceea69506b9270c94cba231e0bf459897077058b/src/Nancy/Json/Simple/SimpleJson.cs)

jchannon:
One is a mistake and needs removing

thecodejunkie:
what @jchannon said.. we also noticed this yesterday and it's the one in Nancy/Json that needs to be deleted

Issue #2623, opened 2016-11-19
Potential Performance Problem with Nancy.Routing.Route.Invoke


Luo-Liang:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of Nancy
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/NancyFx/Nancy/issues) and [closed](https://github.com/NancyFx/Nancy/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Hi there! We may have found a potential problem with performance in Nancy.Routing.Route.Invoke. The use of `TaskCompletionSource` is really slowing down the performance of Nancy when tested with the sample `Nancy.Demo.Hosting.Aspnet` by as much as 10x. This happens both in Debug and Release mode.

### Steps to Reproduce
Run `wrk`, the workload generator with the following parameters in original code.
`-c 256 -t 1 - d 60 http://localhost:59017/user/liang` (256 connections, 1 thread and run for 60 seconds)
and
`-c 1 -t 1 -d 60 http://localhost:59017/user/liang` (1 connection, 1 thread and run for 60 seconds).
`wrk` reports `132` and `270` requests/sec.
The `http://localhost:59017/user/liang` is simply the site serviced by Nancy.
If we change the source code of `Route.cs` from

````
public override Task<object>  Invoke(DynamicDictionary parameters, CancellationToken cancellationToken)
{
    var task = this.Action.Invoke(parameters, cancellationToken);
    var tcs = new TaskCompletionSource<object
    task.ContinueWith(t => tcs.SetResult(t.Result), TaskContinuationOptions.OnlyOnRanToCompletion);
    task.ContinueWith(t => tcs.SetException(t.Exception.InnerExceptions), TaskContinuationOptions.OnlyOnFaulted);
    task.ContinueWith(t => tcs.SetCanceled(), TaskContinuationOptions.OnlyOnCanceled);
    return tcs.Task;
}
````
to simply
````
public override async Task<object> Invoke(DynamicDictionary parameters, CancellationToken cancellationToken)
{
    var task = this.Action.Invoke(parameters, cancellationToken);
    var configuredTask = task.ConfigureAwait(false);
    return await configuredTask;
 }
````
We get `864` and `1129` requests/second respectively with the same configuration on the same machine.

[...]

PR #2684, opened 2017-01-13
Removed TaskCompletionSource usage from Route<T>.Invoke for #2623

davidallyoung:
### Prerequisites

### Description

This PR implements removal of the `TaskCompletionSource` that was causing a performance issue as reported in #2623. It converts the `Route<T>.Invoke` to be an async method signature and just returns the awaited Action.Invoke call.

I attempted to resolve this using the suggestion to return the raw result of Action.Invoke without awaiting from @damianh  and @grumpydev, however the signature of the base method is `Task<object>` and you cannot cast `Task<T>` (the return type of the `Action`) to `Task<object>` without doing various hoops (such as the `TaskCompletionSource` method that was being used).


thecodejunkie:
👀 @NancyFx/most-valued-minions

thecodejunkie:
Can you rebase please? It will hopefully fix the Travis build error since we've pull in #2685

thecodejunkie:
Does not look like this rebase it correct. It is now showing commits by me and @jchannon as well

image

If you are unsure how to rebase, this guide https://github.com/NancyFx/Nancy/blob/master/.github/CONTRIBUTING.md provides some assistance.

Thanks!

davidallyoung:
@thecodejunkie oops sorry about that, VS github plugin trolled me when I pressed sync :(. We should be good now assuming the build completes.

thecodejunkie:
Would this exhibit the same behavior, as the previous code, in case of an exception?

> `task.ContinueWith(t => tcs.SetException(t.Exception.InnerExceptions), TaskContinuationOptions.OnlyOnFaulted);`

Making sure we end up with the correct exception?

Additional feedback @khellang @damianh ?

khellang:
TBH I can't remember why we did it this way in the first place. I think the problem is that the action returns `Task<T>`, while we want `Task<object>`, so instead of going `async`/`await`, we wanted to try to avoid the state machine. If this gives better perf, I think this change is fine.

khellang:
Oh, and now I read the description which I pretty much re-hashed in my previous comment :stuck_out_tongue_closed_eyes:

PR #2697, opened 2017-01-31
Nancy Self Host: Connection Control over the HttpListener

davidallyoung:
### Prerequisites

### Description

NancyHost in its original implementation did not allow a way for consumers to have control over the number of connections. This PR aims to provide a slice of control to consumers to allow them to specify the number of accept connections that can be listened for at a given time.

This implementation has been adapted from https://github.com/JamesDunne/aardwolf/blob/master/Aardwolf/HttpAsyncHost.cs#L107 , which outlines a simple semaphore to control the total number of Hot Tasks listening for a request. To allow consumers to specify the number of connections to fit their environment, I added a property to the HostConfiguration.cs, and specify a default value for it (4) in the default constructor. I've used JMeter locally pointing at the Nancy Self Hosting demo with 150 threads. The performance on my machine is anecdotal and doesn't really matter, but in all this change does not appear to degrade existing functionality. In the context of my environment, throughput was around 2,000/s for the original version that was awaiting on this.Process and with the default MaximumConnectionCount of 4 was around 4,000/s. Again, anecdotal and nothing anyone should take to the bank, but I wanted to provide that as some form of comparison.

I originally jumped in to remove an await call on this.Process as it would cause only one request to be processed at a time, however while in here I thought I could add a little extra bit of functionality for consumers. This will let consumers of Nancy.SelfHosting to fine tune their connection accept loop to meet their needs, as opposed to just throwing out hot this.Process task calls indefinitely.

I'm certainly open to changing this up, and really not sure what a "good" default value would be for the MaximumConnectionCount, as most consumers would probably want to set this after performing their own analysis of their needs and their service's performance. 

Let me know your feedback as this functionality may not be helpful (though I think it's a nice feature) for Nancy.SelfHost, but at the minimum we need to take the await off of the this.Process call or I think we'll only be processing  one request at a time when this package is released :)

### DIFF
--- a/src/Nancy.Hosting.Self/HostConfiguration.cs
+++ b/src/Nancy.Hosting.Self/HostConfiguration.cs
@@ -54,6 +54,16 @@
         /// </summary>
         public bool AllowAuthorityFallback { get; set; }
 
+        /// <summary>
+        /// Gets or sets a property determining how many total connections the NancyHost can maintain simultaneously.
+        /// Higher values mean more conections can be maintained at a slower average response times; while fewer connections will be rejected.
+        /// Lower values will result in fewer conections, yet will be maintained at a faster average response time.
+        /// </summary>
+        public int MaximumConnectionCount { get; set; }
+
+        /// <summary>
+        /// Initializes the default configuration. MaximumConnectionCount is defaulted to 4.
+        /// </summary>
         public HostConfiguration()
         {
             this.RewriteLocalhost = true;
@@ -65,6 +75,7 @@
                     Debug.Write(message);
                 };
             this.EnableClientCertificates = false;
+            this.MaximumConnectionCount = 4;
         }
     }
 }
--- a/src/Nancy.Hosting.Self/HostConfiguration.cs
+++ b/src/Nancy.Hosting.Self/HostConfiguration.cs
@@ -61,21 +61,51 @@
         /// </summary>
         public int MaximumConnectionCount { get; set; }
 
+
         /// <summary>
-        /// Initializes the default configuration. MaximumConnectionCount is defaulted to 4.
+        /// Gets approximate processor thread count by halfing the Logical Core count to 
+        /// account for hyper-threading.
         /// </summary>
+        private static int ProcessorThreadCount
+        {
+            get
+            {
+                // Divide by 2 for hyper-threading, and good defaults.
+                var threadCount = Environment.ProcessorCount >> 1;
+
+                if (threadCount < 1)
+                {
+                    // Ensure thread count is at least 1.
+                    return 1;
+                }
+
+                return threadCount;
+            }
+        }
+
+        /// <summary>
+        /// Initializes the default configuration.
+        /// MaximumConnectionCount by default is half of the Logical Core count.
+        /// </summary>
+        /// <remarks>
+        /// If the system running NancyHost is not using hyper-threading you may want to consider
+        /// supplying your own values for MaximumConnectionCount as the default assumes 
+        /// hyperthreading is being utilized.
+        /// </remarks>
         public HostConfiguration()
         {
             this.RewriteLocalhost = true;
             this.UrlReservations = new UrlReservations();
             this.AllowChunkedEncoding = true;
             this.UnhandledExceptionCallback = e =>
-                {
-                    var message = string.Format("---\n{0}\n---\n", e);
-                    Debug.Write(message);
-                };
+            {
+                var message = string.Format("---\n{0}\n---\n", e);
+                Debug.Write(message);
+            };
             this.EnableClientCertificates = false;
-            this.MaximumConnectionCount = 4;
+            this.MaximumConnectionCount = ProcessorThreadCount;
         }
+
+
     }
 }
--- a/src/Nancy.Hosting.Self/HostConfiguration.cs
+++ b/src/Nancy.Hosting.Self/HostConfiguration.cs
@@ -61,7 +61,6 @@
         /// </summary>
         public int MaximumConnectionCount { get; set; }
 
-
         /// <summary>
         /// Gets approximate processor thread count by halfing the Logical Core count to 
         /// account for hyper-threading.
@@ -98,10 +97,10 @@
             this.UrlReservations = new UrlReservations();
             this.AllowChunkedEncoding = true;
             this.UnhandledExceptionCallback = e =>
-            {
-                var message = string.Format("---\n{0}\n---\n", e);
-                Debug.Write(message);
-            };
+                {
+                    var message = string.Format("---\n{0}\n---\n", e);
+                    Debug.Write(message);
+                };
             this.EnableClientCertificates = false;
             this.MaximumConnectionCount = ProcessorThreadCount;
         }
--- a/src/Nancy.Hosting.Self/NancyHost.cs
+++ b/src/Nancy.Hosting.Self/NancyHost.cs
@@ -13,6 +13,7 @@
     using Nancy.Extensions;
     using Nancy.Helpers;
     using Nancy.IO;
+    using System.Threading;
 
     /// <summary>
     /// Allows to host Nancy server inside any application - console or windows service.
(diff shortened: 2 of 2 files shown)


khellang:
Does it make sense to make the connection count a function of the processor count? Something like https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs#L86 ?

davidallyoung:
I think tying the connection count to logical processor count is a good idea. Is this something we would want to allow consumers to override/set themselves?

khellang:
Sure, I think this is a perfectly reasonable setting to have for power users. The rest can probably get by fine with the defaults.

davidallyoung:
@khellang I've rebased and added a new commit for settingthe default MaximumConnectionCount to 1/2 of the logical core count, similar to how the KestrelServerOptions were implemented.

davidallyoung:
Any thoughts from anyone on this feature?

src/Nancy.Hosting.Self/HostConfiguration.cs
```diff
@@ -54,17 +54,58 @@ public sealed class HostConfiguration
...
         public bool AllowAuthorityFallback { get; set; }
 
+        /// <summary>
+        /// Gets or sets a property determining how many total connections the NancyHost can maintain simultaneously.
+        /// Higher values mean more conections can be maintained at a slower average response times; while fewer connections will be rejected.
+        /// Lower values will result in fewer conections, yet will be maintained at a faster average response time.
+        /// </summary>
+        public int MaximumConnectionCount { get; set; }
+
+
```
thecodejunkie: Can we kill the extra blank line please

src/Nancy.Hosting.Self/HostConfiguration.cs
```diff
@@ -54,17 +54,58 @@ public sealed class HostConfiguration
...
             this.UrlReservations = new UrlReservations();
             this.AllowChunkedEncoding = true;
             this.UnhandledExceptionCallback = e =>
-                {
-                    var message = string.Format("---\n{0}\n---\n", e);
-                    Debug.Write(message);
-                };
+            {
+                var message = string.Format("---\n{0}\n---\n", e);
+                Debug.Write(message);
```
thecodejunkie: Do we want to keep the `Debug.Write` call?

thecodejunkie:
Left a couple of minor review comments

Also, should this be labeled a `Breaking Change`? @khellang 
Shouldn't be needed, right?

davidallyoung:
Jeez, somehow I messed up this commit. I performed the rebase from master after pulling upstream, but it looks like the other commits from master are still included.  Let me know what steps I should take to clean this up. :(

thecodejunkie:
You should be able to look at your `git reflog` and find the SHA for the commit RIGHT BEFORE your rebase. Once you have that, you can do `git reset --hard <insert the SHA here>`. Be sure to be on your `self-host-blocked` branch when performing this.

We have posted some instructions on how to handle changes on upstream, these should help you avoid this in the future https://github.com/NancyFx/Nancy/blob/master/.github/CONTRIBUTING.md#handling-updates-from-upstreammaster :D

davidallyoung:
Thanks @thecodejunkie  , I'm used to to SVN/TFS so still making noob mistakes with git! I'll try this out this evening after work.

thecodejunkie:
@davidallyoung 👍

thecodejunkie:
@davidallyoung any luck? Do you need any more assistance?

davidallyoung:
@thecodejunkie sorry, I haven't had a chance yet, this week has been one of those..... lol

davidallyoung:
Great success, thanks for the reflog tip @thecodejunkie that is handy! I will achieve non-noob status with git, I have faith! I got turned around because I made the mistake of looking at the "sync" view in the VS github plugin instead of just pushing immediately after I had rebased from master.  I'm stating to think I would be better off not looking at that ever again lol.

davidallyoung:
Another side note, I'm making a todo item for myself to document how this works in the Wiki for Nancy Self Host once it merges.

khellang:
> PS, is it normal to need to git push -f after rebasing?

Yep, totally normal. You've basically ripped out your commits, added the new commits from upstream (NancyFx/Nancy) and then replayed your commits at the end, which means you've effectively rewritten history (feels powerful, doesn't it) :wink:

davidallyoung:
It feels powerful enough to horrify me! Thanks @khellang!

thecodejunkie:
@khellang happy to pull this now? 👍 from me

thecodejunkie:
I can't seem to figure out how to approve my review comments since they're no longer available after the rebase?! =)

src/Nancy.Hosting.Self/NancyHost.cs
```diff
@@ -124,20 +125,27 @@ public void Start()
         {
             this.StartListener();
 
-            Task.Factory.StartNew(async () =>
+            Task.Run(() =>
             {
-                try
+                var semaphore = new Semaphore(this.configuration.MaximumConnectionCount, this.configuration.MaximumConnectionCount);
```
khellang: I'm wondering if it's better to use a `SemaphoreSlim` here?

davidallyoung: I'm certainly open to using slim, however the way I understand slim is that it's intended for use when wait times are expected to be short. I'm not sure that there's a gaurantee for that to be true in all cases, especially for a service that experiences no load during certain times of the day. The wait time in this context is the time between a WaitOne a call and GetContextAsync completing and firing the continuation. What do you think?

khellang: Ah, you're right. It's probably fine to keep it as-it :smile:

khellang:
Thanks a ton for this, @davidallyoung! :heart:

davidallyoung:
No problem, I hope to contribute more! Thanks!

davidallyoung:
Just dropping a comment here to note that documentation has been updated for this.

src/Nancy.Hosting.Self/NancyHost.cs
```diff
@@ -124,20 +125,27 @@ public void Start()
...
-                        await this.Process(context).ConfigureAwait(false);
-                    }
-                }
-                catch(Exception ex)
-                {
-                    this.configuration.UnhandledExceptionCallback.Invoke(ex);
-                    throw;
+                        try
+                        {
+                            semaphore.Release();
```
asbjornu: Shouldn't the `Release()` be in a `finally` clause? If the above `GetcontextAsyn().ContinueWith()` statement throws, afaict, the lock will never be released. Whether you actually want a deadlock or not is a difficult choice to make, but [Eric Lippert outlines the different options in this blog post](https://blogs.msdn.microsoft.com/ericlippert/2009/03/06/locks-and-exceptions-do-not-mix/).

davidallyoung: I think this is a valid concern, however if we were to move the Release to a finally clause, we won't be able to accept new requests until the full request is handled, rather than being able to continue accepting new requests after the context is retrieved. The options for finally clauses would in my mind be: 1. Put a finally clause on the try block inside of the continuation, which wouldn't help if GetContextAsync() were to throw, or put a new try finally around this.listener.GetContextAsync(), which wouldn't really work as we want that to execute as a hot task so we wouldn't be blocking the semaphore. 

Another point I would like to make is that an exception here is likely due to a configuration issue, and is going to bubble out as an unhandled exception, so most likely the application is going to terminate unless someone writes explicit handling around the host start and tries to recover. Here's the exceptions that potentially could come out of the BeginGetContext that GetContextAsync invokes: https://docs.microsoft.com/en-us/dotnet/api/system.net.httplistener.begingetcontext?view=netframework-4.7 . I guess the big question to answer here is: Should an exception thrown from GetContextAsync() be recoverable? I think maybe if the answer to that is yes, the application's scope may be better suited utilizing a different host.

I do think the point is fair though and worth discussion!

Issue #2698, opened 2017-02-02
AsString extension no longer works


jchannon:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of Nancy
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/NancyFx/Nancy/issues) and [closed](https://github.com/NancyFx/Nancy/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
We recently moved the Request.Body to a Stream away from a RequestStream.  We have an extension that allows you to get a request body as string however its an extension method off of RequestStream. 

### Steps to Reproduce
To get this to work you have to do `((RequestStream)this.Body)>AsString()` which is nasty.

This needs fixing https://github.com/NancyFx/Nancy/blob/be4f8d42076e4e568a3742715437868e6c7d05af/src/Nancy/Extensions/RequestStreamExtensions.cs#L18

I think this test needs ammending https://github.com/NancyFx/Nancy/blob/be4f8d42076e4e568a3742715437868e6c7d05af/test/Nancy.Tests/Unit/Extensions/RequestStreamExtensionsFixture.cs

### System Configuration

- Nancy version: 2.0-clinteastwood
- Nancy host
  - [ ] ASP.NET
  - [ ] OWIN
  - [ ] Self-Hosted
  - [x] Other: Kestrel
- Other Nancy packages and versions:
- Environment (Operating system, version and so on): OSX but will apply to all
- .NET Framework version:
- Additional information:

thecodejunkie:
So it should be an extension method on `Stream` instead, is that way you are saying?

jchannon:
Yup

thecodejunkie:
Wouldn't it be quicker to send a PR than to write an issue and link stuff? 🤣

jchannon:
Probably but busy and dont want to forget :)

JosephWoodward:
> Wouldn't it be quicker to send a PR than to write an issue and link stuff?

Not to mention it's always nice having PRs like this up for grabs for first timers :)

jchannon:
Make it so...

ghuntley:
[...]

PR #2699, opened 2017-02-03
Move AsString method to be an extension of Stream

pawelsawicz:
### Prerequisites

### Description
That PR is connected to https://github.com/NancyFx/Nancy/issues/2698

Extension method `AsString()` was moved away from *RequestStream* to *Stream*


src/Nancy/Extensions/StreamExtensions.cs
```diff
@@ -5,17 +5,17 @@
...
+    /// Extensions for Stream.
     /// </summary>
-    public static class RequestStreamExtensions
+    public static class StreamExtensions
     {
         /// <summary>
         /// Gets the request body as a string.
         /// </summary>
-        /// <param name="stream">The request body stream.</param>
+        /// <param name="stream">Stream.</param>
```
jchannon: Can you put the stream param body back

pawelsawicz: Yes 👍 , I wasn't sure if I should change that. Same with "Gets the request body as a string."

thecodejunkie:
Failing CI builds due to a broken test

src/Nancy/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// Gets the request body as a string.
+        /// </summary>
+        /// <param name="stream">The request body stream.</param>
+        /// <param name="encoding">The encoding to use, <see cref="Encoding.UTF8"/> by default.</param>
+        /// <returns>The request body as a <see cref="string"/>.</returns>
+        public static string AsString(this Stream stream, Encoding encoding = null)
+        {
+            var reader = new StreamReader(stream, encoding ?? Encoding.UTF8);
```
pawelsawicz: @jchannon could you tell me if that change is correct?

To make `AsString_should_always_start_from_position_0_and_reset_it_afterwards` passing I had to remove `using` because it dispose `stream`.
That way I couldn't assert `Assert.Equal(initialPosition, streamUnderTest.Position);`

pawelsawicz:
Should I remove every occurrence of casting to `RequetStream` in code if we invoke later `AsString()` method? like [this scenario  - SerializerTestModule](https://github.com/NancyFx/Nancy/blob/be4f8d42076e4e568a3742715437868e6c7d05af/test/Nancy.Tests.Functional/Modules/SerializerTestModule.cs#L29) ?

src/Nancy/Extensions/StreamExtensions.cs
jchannon: Yes I think the RequestStream copies the stream internally so that would make sense why that wouldn't pass

jchannon:
Yup, please change the SerializerTestModule

jchannon:
I'm starting to wonder if this is a bigger job than first thought. We have a lot of tests that create fake `RequestStream` to mock a request and test its disposed etc.  Should these tests be using `Stream` instead? I know it depends where its testing because I think we do actually take the initial `Stream` and turn it into `RequestStream`.

@thecodejunkie do you want to have a quick look in our usage/tests of `RequestStream`?

pawelsawicz:
In my opinion, we should change `RequestStream` to `Stream` only when we if we call in this test `AsString()` as in **SerializerTestModule**. Otherwise, that would be separate PR

jchannon:
Yup that's fine

thecodejunkie:
@jchannon 👀

jchannon:
Looks good to me. Only niggle is we don't close the streamreader but not
sure we can do much about that.

pawelsawicz:
Yeah I have that feeling too, but then we would have to remove existing test if we want to close stream.

Unless... I will make a copy of stream inside `AsString()` method and will close copy and return original stream. What do you think ?

src/Nancy/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+            var initialPosition = stream.Position;
+
+            stream.Position = 0;
```
khellang: I think we should check `CanSeek` now that we are doing this on a generic `Stream`...

src/Nancy/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+            var request = reader.ReadToEnd();
```
khellang: `content`? It's not necessarily a request anymore :smile:

src/Nancy/Extensions/StreamExtensions.cs
khellang: We should pass `true` for `leaveOpen` so the stream is left open and keep disposing the reaser.

khellang:
Also, I'm not sure why there's a test change here. The tests should be identical, but working on `Stream` instead.

pawelsawicz:
Ok I think @khellang is right. Existing test should stay same. I brought back old test, also added new test that will assert that stream is not closed.

src/Nancy/Extensions/StreamExtensions.cs
pawelsawicz: Do you mean variable name ?

thecodejunkie:
Just a heads up.. before we pull this (after the review edits are done) you are going to have to squash your commits. Let us know if you need help

jchannon:
We can squash and merge in the GH UI

thecodejunkie:
Unless we can change the commit message, before we merge, then it won't be an option tbh

pawelsawicz:
Yes we can, commit message will be same as PR name. Example https://github.com/Helbreath/Helbreath.HGServerConsole/commits/master. But I can squash in git as well it's not a problem

jchannon:
When you squash and merge in GH it asks what you want the commit message to
be.

src/Nancy/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+                    stream.Position = 0;
+
+                    var request = reader.ReadToEnd();
+
+                    stream.Position = initialPosition;
+
+                    return request;
+                }
+                return null;
```
khellang: `string.Empty`? I'm not a fan of returning `null` strings.

test/Nancy.Tests/Unit/Extensions/StreamExtensionsFixture.cs
```diff
@@ -29,5 +29,25 @@ public void AsString_should_always_start_from_position_0_and_reset_it_afterwards
...
             }
         }
+
+        [Fact]
+        public void AsString_should_not_close_stream()
+        {
+            // Given
+            using (var innerStream = new MemoryStream())
+            using (var streamWriter = new StreamWriter(innerStream, Encoding.UTF8) { AutoFlush = true })
+            using (var streamUnderTest = new MemoryStream())
```
khellang: Why do we have this extra stream here? Can't we just do `innerStream.AsString`?

jchannon:
Ah I see what @thecodejunkie was getting at maybe. If we squash and merge in the GH UI we won't get a merge commit. Is that a problem going forward, I know we always tend to squash locally and create a merge commit?

@pawelsawicz can you squash locally and push for now?

jchannon:
Good work! 

Thanks ❤️

PR #2702, opened 2017-02-08
[WIP] Investigate tests that may be intermittently failing on master. DO NOT MERGE

blairconrad:
### Prerequisites

- [ ] I have written a descriptive pull-request title
- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open
- [ ] I have verified that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)
- [ ] I have provided test coverage for my change (where applicable)

PR #2634, opened 2016-11-23
Add netcoreapp1.0 test target to Nancy.Tests

blairconrad:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open
- [x] I have verified that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)
- [x] I have provided test coverage for my change (where applicable)

### Description

Supports #2612.

A preliminary attempt to convert Nancy.Tests to target a netcoreapp1.0.
I expect I've done many things wrong, and invite corrections.
I'll highlight what I think are the most egregious violations.

- update FakeItEasy to 3.0.0-alpha001
- add workaround so FakeItEasy gets required Castle.Core-beta002
- poorly patch the build so the project compiles and tests run

- currently
  - ~~`CookieBasedSessionsFixture.Should_load_valid_test_data` fails~~, and
  - ~~`DefaultNancyBootstrapperFixture.Container_should_ignore_specified_assemblies`
    is conditionally compiled out~~

Issue #2659, opened 2016-12-22
Nancy 2.0.0-clinteastwood feedback 


thecodejunkie:
If you find any problems, please [open an issue](https://github.com/NancyFx/Nancy/issues/new) and provide as much information, as possible, to help us identify your problem. Alternatively, feel free to submit a [pull-request](https://github.com/NancyFx/Nancy/pulls) if you think you can help resolve the problem. If you need a guiding hand, on how to get started with pull-requests, please have a look at our [contributing](https://github.com/NancyFx/Nancy/blob/ab80c22b3aa5226a860b84cc65a90660c10fb88f/.github/CONTRIBUTING.md) guidelines. Make sure you don't miss the [style guidelines](https://github.com/NancyFx/Nancy/blob/ab80c22b3aa5226a860b84cc65a90660c10fb88f/.github/CONTRIBUTING.md#style-guidelines), I won't :D

> **NOTE:** Because Nancy 2.0-clinteastwood is a pre-release, you are going to have to install packages using the `-pre` toggle, i.e `install-package Nancy -pre`

Please use this issue for posting comments that are not directly related to any specific reported bug or pull-request.

## TODO

The following is a list of identified task that needs to be performed in the wake of the `2.0.0-clinteastwood` release and should be completed before `2.0.0-dangermouse` is released. Edit the TODO-list and add items as they are found, add links to pull-requests that are sent to close them

- [x] Make sure `AsyncUsageAnalyzers` it set to a build dependency in *all* `project.json` and `packages.config` Correct example to mirror [here](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/project.json#L13-L18) - (pull request #2700 @jchannon
- [x] Remove all old `*.nuspec` files in all projects (including Nancy) (pull-request #2658 @jchannon)
- [x] Update `Nancy.Hosting.Aspnet\project.json` to include `web.config.transform` in the package (pull-request #2661 @thecodejunkie)
- [x] Update `Nancy.ViewEngines.Razor\project.json` to include `app/web.config.transform` in the package (pull-request #2661 @thecodejunkie)
- [ ] Update `Nancy.ViewEngines.Razor\project.json` to include `Nancy.ViewEngines.Razor.BuildProvider` (including build targets) in the package

## Planned changes

- [ ] Remove one of `Request.Cookies` and `Request.Headers.Cookie` see #2628 for discussions
- [ ] Improve the Before and After hook syntax and capabilities #2665
[...]

PR #2703, opened 2017-02-09
Disable parallel test runs for Nancy.Tests

blairconrad:
Several fixtures manipulate (and possibly more depend on) the static field
StaticConfiguration.CaseSensitive, causing interference.

### Prerequisites

### Description
Disable test run parallelization for Nancy.Tests. 
I noticed intermittent test failures in this module (even on master, on Windows) during my work on #2634. After further investigation performed as described in #2702, helped along with comments from @xt0rted  and @jchannon, I see that both `Nancy.Tests.Unit.RequestFixture` and `Nancy.Tests.Unit.Routing.DefaultRouteResolverFixture`
` set `StaticConfiguration.CaseSensitive`. I figured this to be a primary cause of the failures.

I ran `dotnet test -c Release` in Nancy.Tests repeatedly on Windows on the master branch, then disabled test parallelization and repeated.

Here's what I found:
With parallelization enabled (as `master` has it), tests failed after various numbers of runs:
7, 3, 6, 1, 2, 2, 2, 3, 1, 2, 1, 3, 1, 1, 1, 8, 11, 4, 2, 1, 2, 1, 1, 12, 7, 3, 1, 1, 7, 1, 8, 15, 10, 1, 4, 3, 3, 1 , 2, 1, 1, 2, 1, 2, 1, 4, 14, 3, 14, 10, 3, 1, 1, 2, 3, 3, 1, 4, 5, 6, 1, 9, 7, 4, 12, 7, 1, 5, 3, 1, 6, 9, 6, 1, 6, 2, 2, 1, 1, 2, 2, 2, 6, 2, 1 , 5, 11, 4, 4, 2, 1, 3, 7, 3, 3, 6, 2, 10, 1, 1, 26, 2, 1, 7, 4, 10, 2, 1, 12, 2, 1, 1, 7, 1, 1, 4, 1, 7, 6, 2, 4, 1, 1, 4, 3, 2, 2, 1, 3, 1, 2, 3, 1, 2, 5, 2, 7, 7, 2, 6, 6, 1, 1, 1, 2, 4, 3, 6, 4, 2, 13, 1, 8, 2, 2, 10, 8, 1, 6, 5, 3, 6, 2, 1, 5, 1, 2, 1, 10, 9, 4, 3, 5, 6, 2, 9, 3, 1, 2 , 17, 1, 9, 3, 4, 5, 5, 1, 2, 5, 1, 4, 2, 1, 1, 1, 1, 8, 2, 2, 1, 5, 8, 2, 2, 2, 3, 3, 2, 2, 1, 1, 3, 4, 3, 3, 1, 3, 1, 2, 1, 1, 3, 3, 1

With parallelization disabled, the tests did not fail once in over 200 runs.

Here's a picture:

image

It's possible that a more sophisticated fix could be found, but in the interim, I propose disabling parallel test execution.

### DIFF
--- a/test/Nancy.Tests/Properties/AssemblyInfo.cs
+++ b/test/Nancy.Tests/Properties/AssemblyInfo.cs
@@ -1,0 +1,3 @@
+﻿using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]


thecodejunkie:
Copied this from #2702 

@blairconrad It's no unlikely that StaticConfiguration is the cause of the problem. I've previously added a task to remove the last, remaining parts, of StaticConfiguration in #2659

A while back I also created StaticConfigurationContext to help mitigate race conditions, in tests, that used StaticConfiguration values. It looks like DefaultRouteResolverFixture does not use it.. would be interesting to see if adding it would solve the intermittent test failtures

blairconrad:
@thecodejunkie, [StaticConfigurationContext](https://github.com/NancyFx/Nancy/blob/3e70d45e2c06210bb489ebb9c011ccdeb9c22f83/src/Nancy.Testing/StaticConfigurationContext.cs) looks like it would help for serialized tests, but I think won't protect us when the tests are run in parallel. Unless I'm missing something, it performs no synchronization, so the other fixtures could still muck about with the setting, no?

blairconrad:
I'm not sure what to make of the Travis failure. At work, so no time/ability to investigate.

blairconrad:
… at least it's [not the first time it's failed](https://travis-ci.org/NancyFx/Nancy/builds/187582689#L2442)

thecodejunkie:
I'm kind of tempted to think that it's also `StaticConfiguration` that might be causing it. I'm just quasi-guessing here, but [LiteralNode](https://github.com/NancyFx/Nancy/blob/704530c7d8a205234738262da5f6f08e360d7c7c/src/Nancy/Routing/Trie/Nodes/LiteralNode.cs#L37) is also using `StaticConfiguration.CaseSensitive` so maybe the parallel test execution screw it up sometimes so the `BrowserFixture` JSON test doesn't match a route.. 🤔

blairconrad:
Interesting idea, @thecodejunkie. If you'd like, I could try disabling the parallel execution on Nancy.Testing.Tests as well. (But where will it all end!? :wink:)

thecodejunkie:
To be honest, we're working on removing `StaticConfiguration` so we can live with the intermittent failure for the time being

blairconrad:
Fair enough. Should we close this PR then?

jchannon:
Does that assemblyinfo change work for netcore tests?

blairconrad:
Seems to. Test times jump from about 21–22 seconds to about 34–35 seconds when I add the `AssemblyInfo.cs`.

jchannon:
In netcore you pass it to the runner eg `dotnet test -parallel none` plus .netcore doesnt know what assembly info is, it uses project.json

blairconrad:
@jchannon, I would be the first to admit that I don't fully understand .NET Core or XUnit, but I'm confused by your earlier comments.

> In netcore you pass it to the runner eg dotnet test -parallel none

This seems to be _a_ way to disable parallel test execution, but I don't think it's the only way.

> .netcore doesnt know what assembly info is, it uses project.json

I don't know exactly what you're suggesting by "doesn't know what assembly info is". The file is built into the assembly. The attribute does appear when I reflect the assembly. And as I explain below, it seems to make a difference. Also, in a [separate project](https://github.com/blairconrad/SelfInitializingFakes), I'm including an equivalent file in the build, specifying assembly version numbers, and they are respected.

I think it would be up to XUnit to interpret the attribute once it's there; this wouldn't be something directly controlled by .NET Core. I took a peek at the XUnit source, and it seems to be treating the `CollectionBehavior` attribute the same way whether compiled under .NET Core or not, so I'd expect it to work. And consider my evidence:

1. when I include the attribute, Nancy.Tests execution time rises from about 21 seconds to about 34. This is repeatable over many test runs
2. when I run `dotnet test -parallel none`, the time also rises to about 34 seconds per run, which is pretty suggestive
3. ( I also seem to get more test failures without the attribute, but I've not repeated my original experiment, so at this point, my evidence is anecdotal)

blairconrad:
Oh, and is the whole discussion moot, since you're trying to get rid of `StaticConfiguration`, and that may obviate the need for serializing the tests?

blairconrad:
In case we're keeping track, [we've another failed build](https://travis-ci.org/NancyFx/Nancy/builds/204166444#L2712).

But at this I'm more interested in knowing whether

1. we should close the PR because there's no interest, or
2. if there's interest, what more you'd like done before a merge can happen

Thanks!

thecodejunkie:
@NancyFx/most-valued-minions @NancyFx/owners I think we should pull this in for now.. it seems that it's causing test failures in most PRs after we enabled more test projects to run on core. Any objections?

jchannon:
nope do it

blairconrad:
Want a rebase?

blairconrad:
Rather, object to a rebase?

blairconrad:
Ah, to heck with it. I'm rebasing.

blairconrad:
Thanks, @thecodejunkie!

thecodejunkie:
@blairconrad thank you! 👍

PR #2715, opened 2017-03-14
Fix for ViewNotFound error on netcoreapp

Sphiecoh:
### Prerequisites

### Description

When using Nancy with a view engine on netcoreapp , the default root path points to the debug folder which then requires views to be copied to output.
This however has a limitation because updates to views need to be copied.

### DIFF
--- a/src/Nancy/DefaultRootPathProvider.cs
+++ b/src/Nancy/DefaultRootPathProvider.cs
@@ -14,7 +14,7 @@
         public string GetRootPath()
         {
 #if CORE
-            return Microsoft.Extensions.PlatformAbstractions.PlatformServices.Default.Application.ApplicationBasePath;
+            return System.IO.Directory.GetCurrentDirectory();
 #else
             return AppDomain.CurrentDomain.BaseDirectory;
 #endif


khellang:
Hmm. I wonder if we should get rid of [these](https://github.com/NancyFx/Nancy/search?utf8=%E2%9C%93&q=PlatformServices) and remove the dependency as well?

khellang:
Oh, and all the tests broke 🤔

Sphiecoh:
`Compiling Nancy for .NETFramework,Version=v4.5.2
/home/sifiso/Projects/Nancy/src/Nancy/project.json(69,31): error NU1001: The dependency System.Xml could not be resolved.`

How do I fix this on Ubuntu 16.04 ? So I can run tests and fix the failing one's.

thecodejunkie:
For full-framework build you need to have Mono installed

Sphiecoh:
``` 
mono -V
Mono JIT compiler version 4.2.1 (Debian 4.2.1.102+dfsg2-7ubuntu4)
```

Sphiecoh:
I think we need to remove the tests , any objections ?

src/Nancy/DefaultRootPathProvider.cs
```diff
@@ -14,7 +14,7 @@ public class DefaultRootPathProvider : IRootPathProvider
         public string GetRootPath()
         {
 #if CORE
-            return Microsoft.Extensions.PlatformAbstractions.PlatformServices.Default.Application.ApplicationBasePath;
+            return System.IO.Directory.GetCurrentDirectory();
```
thecodejunkie: Not entirely convinced that this is the best approach - the current working directory can change during the lifetime of an application, by such a simple thing as a call to ` Environment.CurrentDirectory = <some path>`

thecodejunkie: Not to mention that the behavior would diverge between netcore and full-framework

thecodejunkie:
@Sphiecoh why would we remove the tests?

src/Nancy/DefaultRootPathProvider.cs
Sphiecoh: But we need a strategy because the status quo is deviating from happy path , having to implement a roothpath provider all the time.

thecodejunkie:
@Sphiecoh just to help me gain a bit of context

> When using Nancy with a view engine on netcoreapp , the default root path points to the debug folder which then requires views to be copied to output. This however has a limitation because updates to views need to be copied.

You are talking about a "self-hosted" scenario here right? Like running on top of Kestrel, i.e just like when you've used `Nancy.Hosting.Self`?

Sphiecoh:
Yes

thecodejunkie:
@Sphiecoh OK.. so it's always been the case that you need to copy views, to the output folder, when you run self-hosted (core or full-framework). Ie you'd set them to "copy to output" in VS or similar approach. This is the way it's been since we introduced self-host and it's the nature of how that kind of hosting works. We cannot have the root path be higher up than the actual application, that's a security problem and it would also mean that you could serve files from a path that is not the path of your actual application.

The `IRootPathProvider` is used by many things in Nancy, such as static file handling, views, file downloads and uploads and so on, so this change would cause a massive regression to a lot of existing applications (again, it would also promote a very bad practice for any new apps)

thecodejunkie:
Right, we've had a talk and we're not comfortable with putting the root path above the application path. The solution is to copy your views to your output folder.

Thanks

Issue #2719, opened 2017-03-20
Migrate to new project system


thecodejunkie:
The time has come to migrate away from the old `*.csproj` and `project.json` formats in favour of the new and improved `*.csproj` format that is supported by Visual Studio 2017 and Visual Studio Code 

## Todo
The following steps are required to be done on each repository

- Migrate to the new project system using `dotnet migrate` (including adding missing projects and verify all tests) (@thecodejunkie)
- Add `Directory.build.props` file and update all `* .csproj` files (see [MiniProfiler](https://github.com/MiniProfiler/dotnet/blob/master/src/Directory.build.props) for an example)
- Add `DisableImplicitFrameworkReferences` element to all `*.csproj` for full-framework builds (see [OwinHttpMessageHandler](https://github.com/damianh/OwinHttpMessageHandler/blob/netstandard/src/OwinHttpMessageHandler/OwinHttpMessageHandler.csproj#L18) for an example)
- Migrate back to `netstandard1.3` again (need to update `build.cake` to reflect this, i.e all references of `netstandard1.6`)
- Update `build.cake` to support running on Mono with the new project SDK (@akoeplinger confirmed that `xbuild` will no longer be able to build the new project format on mono`)
- Update all test projects to use `xunit 2.2.0` and `xunit.runners.visualstudio 2.2.0`

## Adopting usage of NETStandard.Library

1. Add the `DisableImplicitPackgeReference` element and set it to true
2. Remove all our explicit package references for netstandard1.x TFM
3. Add an explicit reference to `NETStandard.Library v1.6.1`
4. Add what ever additional packages that we need that will not be part of `NETStandard.Library`

## Demo projects
The only `Nancy.Demo.*` project that was ported to `project.json` was the new `Nancy.Demo.Hosting.Kestrel` project. Because of this, all other demo projects ends up with very bloated `*.csproj` files after `dotnet migrate` has been run on the projects. We need to clean these up and make them use the new, lighter, format

http://rehansaeed.com/cleaning-up-csproj/

## Extra
Investigate using a single MSBuild invocation to speed up the execution time. Look at example https://gist.github.com/dasMulli/69f5303aa79a8cd4060e44891c90fd2d that was suggested by @dasmulli

## Build script
[...]

PR #2435, opened 2016-04-30
Add support for .net 4.5 data annotations

xt0rted:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open
- [x] I have verified made sure that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)
- [x] I have provided test coverage for your change (where applicable)

### Description
This is an almost exact copy/paste of [xt0rted/Nancy.Validation.DataAnnotations.Extensions](https://github.com/xt0rted/Nancy.Validation.DataAnnotations.Extensions). I updated the coding style in some areas, but didn't carry over what was in the sample project.

I also removed a prior mono hack since we're targeting versions which fix the bug we were working around.

The tests I carried over used a base class so I didn't duplicate the setup code. If you'd like that removed say the word.

This fixes #1577

PR #2720, opened 2017-03-22
Migrate to the new project system

thecodejunkie:
This pull-request will handle the parts of the [#2719 Migrate to new project system](https://github.com/NancyFx/Nancy/issues/2719) epic, which are related to the main Nancy repository. 

## Todo
- [x] Perform a `dotnet migrate` and clean up things like solutions files and `*.MSBuild` folders
- [x] Update `NuGet.config` and remove all sources except `https://api.nuget.org/v3/index.json`
- [x] Make `Nancy.ViewEngines.Razor` and `Nancy.ViewEngines.Razor.BuildProviders` compile again
- [x] Make `Nancy.Validation.DateAnnotations` compile again
- [x] Make `Nancy.Validation.DateAnnotations.Tests` compile again
- [x] Update all test projects to use `xunit 2.2.0`, `xunit.runners.visualstudio 2.2.0` and `Microsoft.NET.Test.Sdk 15.0.0`
- [x] Add `Directory.build.props` file for projects in `test/`
- [x] Add `Directory.build.props` file for projects in `src/`
- [x] Add a `.csproj` for `Nancy.ViewEngines.Spark.Tests`
- [x] Add base package tags in `Directory.build.props` for projects in `src/`
- [x] ~~Add all projects in `samples/` to `Nancy.sln`~~ (will be done as a separate pull-request)
- [x] ~~Clean up `*.csproj` files for projects in `samples/`~~ (will be done as a separate pull-request)
- [x] Update `*.csproj` files
- [x] Make all tests pass again

## Steps for updating the *.csproj files
- [x] Update `*.csproj` files to take an implicit dependency on `NETStandard.Library` instead of explicitly cherry-pick packages
- [x] Add `DisableImplicitFrameworkReferences` element and set it to `true` for `net452` (see [gist](https://gist.github.com/thecodejunkie/f87c2b680f9235f33a0543b53ece1d6e) and https://github.com/dotnet/sdk/pull/633#issuecomment-272515440)
- [x] Remove all uses of the `RuntimeFrameworkVersion` element
- [x] Remove all explicit references of `NETStandard.Library` and `Microsoft.NETCore.App`
- [x] Update all references of `Microsoft.Extensions.PlatformAbstractions` to `1.1.0`
- [x] Update all references of `Microsoft.Extensions.DependencyModel`  to `1.1.1` (or `1.1.0` ?)
- [x] Investigate (and if possible remove) the use of `Microsoft.NETCore.Platforms` package in the test projects
- [x] Bump all uses of `netcoreapp1.0` to `netcoreapp1.1`
- [x] Investigate (and if possible remove) the use of `PackageTargetFallback`
- [x] Remove explicit use of `PackageId` (will fallback to `AssemblyName`)
- [x] General clean up of elements (test projects don't need a version, sort the elements, order target frameworks)
- [x] Remove use of `AssemblyName` (will fallback to the name of the `.csproj`)
- [x] Update all `System.*` and `Microsoft.*` package reference to latest for `netstandard`
- [x] ~~Migrate back to `netstandard1.3` now that [Microsoft.Extensions.DependencyModel](https://www.nuget.org/packages/Microsoft.Extensions.DependencyModel/) supports it again (will have an API impact on `Stream` for methods such as `Begin/EndRead` and `Begin/EndWrite` which are not available on `netstandard1.3`)~~ (will be done as a separate pull-request)

See https://github.com/dotnet/sdk/pull/633 for discussions on config options in `*.csproj`

See [Additions to the csproj format for .NET Core](https://docs.microsoft.com/en-us/dotnet/articles/core/tools/csproj) on MSDN for `*.csproj` recommendations

### build.sh
- [x] Update `CAKE_VERSION=0.19.1`
- [x] Update `DOTNET_VERSION=1.0.1`
- [x] Update `DOTNET_INSTRALL_URI=https://raw.githubusercontent.com/dotnet/cli/rel/1.0.1/scripts/obtain/dotnet-install.sh`
- [x] Update to support running on Mono with the new project SDK (#2719 for details)

[MSBuild wrapper for macOS](https://gist.github.com/dasMulli/29c7d68f321cc5ac4a68407a6fd76d19)

### build.ps1
- [x] Update `$CakeVersion = "0.19.1"`
- [x] Update `$DotNetVersion = "1.0.1";`
- [x] Update `$DotNetInstallerUri = "https://raw.githubusercontent.com/dotnet/cli/rel/1.0.1/scripts/obtain/dotnet-install.ps1";`

## build.cake
- [x] ~~Update all references of `netstandard1.6` to `netstandard1.3`~~ (will be done as a separate pull-request)
- [x] Update `Restore-NuGet-Packages` task to use `NuGet.config`
- [x] Investigate if we can move Mono specific build configurations (`MonoDebug` and `MonoRelease`)
- [x] Update so that we do not run `net452` builds on *nix
- [x] Change the `UpdateProjectJsonVersion` method so that it updates the `VersionPrefix` element of `*.csproj` files instead of `project.json` (should be done after adding `Directory.build.props`)

## Changes performed in this pull-request (will be updated as more commits are added)
- Removed all `*.MSBuild` folders in `src/` and `test/`
- Removed `Nancy.sln` and `Nancy.sln.DotSettings`
- Renamed `Nancy.Next.sln` and `Nancy.Next.sln.DotSettings` to `Nancy.sln` and `Nancy.sln.DotSettings`
- Ran `dotnet migrate` on the entire repository
- Updated references of `xunit` to `2.2.0`
- Updated references of `xunit.runners.visualstudio` to `2.2.0`
- Updated references of `Microsoft.NET.Test.Sdk` to `15.0.0`
- Removed unused sources in `NuGet.config`
- Updated `build.sh` (see above for list of changes)
- Updated `build.ps1` (see above for list of changes)
- Removed all uses of the RuntimeFrameworkVersion element
- Added `DisableImplicitFrameworkReferences` for all `net452` targets
- Updated `*.csproj` files to take an implicit dependency on `NETStandard.Library` and `Microsoft.NETCore.App`
- Updated all references of `Microsoft.Extensions.PlatformAbstractions` to `1.1.0`
- Updated all references of `Microsoft.Extensions.DependencyModel` to `1.1.1`
- Removed all references of the `Microsoft.NETCore.Platforms package` in the test projects
- Removed all uses of the RuntimeFrameworkVersion element
- Updated all uses of `netcoreapp1.0` to `netcoreapp1.1`
- Removed use of `PackageTargetFallback` in all projects
- Removed explicit use of `PackageId`
- Sorted elements in `*.csproj` files
- Sorted TFMs in `TargetFrameworks` elements
- Removed explicit use of the `AssemblyName` element
- Updated all `System.*` and `Microsoft.*` package reference to latest for `netstandard`
- Added `Directory.build.props` file for projects in `test/`
- Added `Directory.build.props` file for projects in `src/`
- Added base package tags in `Directory.build.props` for projects in `src/`
- Added `.csproj` for `Nancy.ViewEngines.Spark.Tests` and added it to `Nancy.sln`
- Changed `VersionPrefix` to `Version` in `src/Directory.build.props`
- Cleaned up Build.cake
- Fixed broken tests on all tfms
- Updated `Nancy.sln` build configurations


test/Nancy.Testing.Tests/Nancy.Testing.Tests.csproj
```diff
@@ -26,12 +26,12 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0-preview-20170106-08" />
```
khellang: Can't we drop `-preview-20170106-08` here?

thecodejunkie: Oh shit I thought I did that yesterday. I must've forgotten to push that commit.. If you check the `Changes performed in this pull-request` section above, I even added it after my last commit :D I'll push that out. Well spotted 👍

test/Directory.build.props
```diff
@@ -0,0 +1,42 @@
+<Project>
+  <PropertyGroup>
+    <DisableImplicitFrameworkReferences Condition=" '$(TargetFramework)' == 'net452' ">true</DisableImplicitFrameworkReferences>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.1' ">
```
khellang: Is it weird that these `TargetFramework` conditions are in here when the actual target frameworks themselves are defined in the project file?

thecodejunkie: ¯\_(ツ)_/¯

I reckoned that we define shared stuff in this file and it will be used by projects based on their supported target frameworks. Ideally all of our test projects can be changed to support both `net451` and `netstandard`

test/Directory.build.props
```diff
@@ -0,0 +1,42 @@
...
+    <ProjectUnderTest>$(MSBuildProjectName.Substring(0, $(MSBuildProjectName.IndexOf(".Tests"))))</ProjectUnderTest>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../../src/Nancy/Nancy.csproj" />
+    <ProjectReference Include="../../src/$(ProjectUnderTest)/$(ProjectUnderTest).csproj" />
+    <ProjectReference Include="../../src/Nancy.Testing/Nancy.Testing.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(MSBuildProjectName)' != 'Nancy.Tests' and '$(MSBuildProjectName)' != 'Nancy.ViewEngines.Razor.Tests.Models' ">
```
thecodejunkie: This also a bit of a hack that we hopefully can clean up down the road. Basically we cannot include these files if we`re building `Nancy.Tests` because the files themselves are in that project, so we end up getting warnings about the files being declared twice.

If we include them for `Nancy.ViewEngines.Razor.Tests.Models` then we end up with a similar error because it will end up including the files both for `Nancy.ViewEngines.Razor.Tests.Models` and `Nancy.ViewEngines.Razor.Tests`

test/Directory.build.props
```diff
@@ -0,0 +1,42 @@
...
+    <DefineConstants>$(DefineConstants);MONO</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <ProjectUnderTest>$(MSBuildProjectName.Substring(0, $(MSBuildProjectName.IndexOf(".Tests"))))</ProjectUnderTest>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../../src/Nancy/Nancy.csproj" />
+    <ProjectReference Include="../../src/$(ProjectUnderTest)/$(ProjectUnderTest).csproj" />
```
thecodejunkie: This turned out nice.. automatically include the project that is being tested, based on our project naming convention 😄

thecodejunkie:
# Mono for Nancy 2.x?

We have supported Mono, pretty much since the inception of Nancy, as a way to run Nancy cross-platform. With the move to the new project system we are finding ourselves at a cross road. The Mono tooling does not support the new project format. It will support it but it is "down the road" and will not be part of 4.8 at least.

The `dotnet` tooling does not use Mono, if installed, if you try and build a full-framework target framework build (like `net452`) on a non-Windows system. That said, there is a way to hack around it by setting the `FrameworkPathOverride` using

> export FrameworkPathOverride=$(dirname $(which mono))/../lib/mono/4.5/

as per https://github.com/dotnet/netcorecli-fsc/wiki/.NET-Core-SDK-rc4#using-net-framework-as-targets-framework-the-osxunix-build-fails there is also the fact that Microsoft has dropped Mono support for their own x-plat story.

When I run `dotnet build`, after setting `FrameworkPathOverride`, on macOS I think most projects build except the `Nancy.ViewEngines.Razor` and `Nancy.ViewEngines.Razor.Tests`. They're both failing because of our dependency on `Microsoft.NET.Compilers` (for compiling views) 

> /Users/andreas/.nuget/packages/microsoft.net.compilers/1.3.2/tools/Microsoft.CSharp.Core.targets(67,5): error MSB6006: "csc.exe" exited with code 8. [/Users/andreas/Nancy/src/Nancy.ViewEngines.Razor/Nancy.ViewEngines.Razor.csproj]

It simply does not work on Mono using the dotnet tooling. Not sure if it resolves the wrong version or not base don the Runtime Identifier (explicitly setting the RID, before a restore, does not resolve it either).

## Tooling and code impact

So currently you cannot simply run a `dotnet build` (or `dotnet test`since it depends on the build) on our solution, when not on Windows since you will get the failure on Razor. We also have some legacy stuff around like the `MonoDebug` and `MonoRelease` configurations + corresponding `IF#` compiler directives (mostly in our tests). 

## Decision time

Do we support Mono for `2.x` or do we leave `1.x` for .NET Framework + Mono and make `2.x` for .NET Framework and .NET Core (for x-plat)? 

- Please leave a 👍 to make `2.x` for .NET Framework and .NET Core (for x-plat)
- Please leave a 👎 if we should not drop the Mono support

Also leave a comment if you have anything to add to the discussion

Yantrio:
Is it worth noting that dropping mono support here doesn't mean that people developing their apps on mono wont be able to consume a nancy package. It just means that they can't build nancy from source, correct?

thecodejunkie:
@Yantrio yeah, that is a very good point. Essentially it means that _we_ will not pay any special attention (in code, build scripts and so on) to verify that our code works on Mono. It does not mean that we will stop accepting pull-requests with fixes to keep Nancy functioning. Also with Mono moving more and more towards compatibility with the .NET Framework, the gap should be narrowing down as they progress as well.

Basically we'll stop with

```
#if Mono
  Do special thing
#endif
```

special cases in our code + no build against a Mono target on our CI severs

xt0rted:
I don't use mono so I'm not casting a vote on this, but as someone who has hit some annoying mono incompatibilities in the data annotations validation code (#2435 & https://github.com/dotnet/corefx/issues/8201) I'd love to see cross platform use .net core instead since it'd get rid of them and the special `#if` cases.

ChrisMcKee:
:+1: LET IT BURN 
3pj2zdg

Petermarcu:
If the new .net sdk tooling were to support building for .net framework targets when on OSX or Linux would that help? Sounds like that is one of a few challenged you are trying to work through.

thecodejunkie:
@Petermarcu to some extent yes.. to other no. There are still API differences between latest Mono and .NET Framework, meaning we have to work around those as well.. but sure if the tooling was opaque to the underlying framework then it would help

Petermarcu:
@thecodejunkie so if everything you needed was in some set of reference assemblies that represented the subset of .NET Framework and Mono that you need, you would at least be able to build the things you need for .NET Framework or Mono. I know the idea of publishing ref assemblies to support this is something being investigated. I wasn't sure if thats the only blocker for you.

thecodejunkie:
@mykemaas It would be useful to us if you could explain the reason(s) for your 👎 vote. Thanks

crmtogether:
Okay so i down voted this but maybe its not so bad an issue for us (yes i am being selfish and only thinking about myself here)
We build on windows but deploy on mono. And we don't use #mono but Type.GetType("Mono.Runtime"). 
Maybe that's the wrong way to use it but it is quicker than on windows server.

monocoder:
Nancy advantage lies in compatibility.

thecodejunkie:
@monocoder please elaborate. How will that be an advantage vs running x-plat using .net core ?

thecodejunkie:
@Petermarcu we've also run into things where the APIs are similar but the behaviour is different like #2435

mykemaas:
I downvoted because I use Nancy extensively in Mono.  I build in Visual Studio, but deploy to Mono.  Our projects are microservices that are attached to endpoints using nginx in Linux.  Dot Net Core has been a hard sell for me because of its many quirks and limitations.  I'm sure all of that will work out eventually, so maybe it doesn't matter, but for now, the Nancy pattern combined with (almost) full .NET in Mono has been truly awesome.

mykemaas:
I'd like to qualify what I said earlier...  I'm actually sort of okay with 2.0 departing from Mono, as long as 1.4 remains supported and updated.

thecodejunkie:
@mykemaas 1.x is in maintenance mode and will not have any feature releases, only 1.4.x releases if critical bugs surface

Petermarcu:
@thecodejunkie. Yes, there will likely always be all differences between the implementations. The goal over time is for them all to behave the same for the same apis. Let me push a little on making the reference assemblies for .Net framework available to cross platform projects so they can build. My understanding is that those will be made available sometime soon

TheAngryByrd:
Is is possible to build with mono 5.0 since csc is shipping with it? http://www.mono-project.com/docs/about-mono/releases/5.0.0/#csc

damianh:
If mono 4.6 supports NetStandard 1.6 then do we, as library developers and maintainers, need to have a build & test matrix that covers all deployment platforms? Or are we just duplicating the efforts of Microsoft?

Personally I wouldn't bother.

khellang:
awww yeah

thecodejunkie:
image

PR #2771, opened 2017-07-17
Change VS 2015 mention to VS 2017.

RhinoDevel:
Opening Nancy solution no longer works correctly with VS 2015.


khellang:
Thanks, and welcome as a Nancy contributor, @RhinoDevel! ❤️

PR #2772, opened 2017-07-20
Added ModuleSearchDepth property and logic

jbzcooper:
Changes to AppDomainAssemblyCatalog:

1. Added public ModuleSearchDepth property to enable scanning deeper into folder structure for NancyModules.
2. Added getDirectories method to scan a path for sub directories
3. Modified LoadNancyReferencingAssemblies method to leverage getDirectories method to search for dlls


thecodejunkie:
I think we are going to reject this. A big part of `2.0` is moving away from static state - for a whole range of reasons. So adding a `public static` property would be a step in the wrong direction. Also, this seem to be some what app-specific, especially since we've never had this request before over the years. 

With the introduction of `IAssemblyCatalog` it is easy enough to use a custom implementation if/when you need it. 

Thanks anyway!

PR #2793, opened 2017-09-12
README.md: Update copyright year

perlun:
Happened to see this while skimming through the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ This project is supported by the [.NET Foundation](http://www.dotnetfoundation.o
 
 ## Copyright
 
-Copyright © 2010 Andreas Håkansson, Steven Robbins and contributors
+Copyright © 2010-2017 Andreas Håkansson, Steven Robbins and contributors
 
 ## License
 


thecodejunkie:
@perlun Hi. Though commonly believed, the copyright notice does not have to be a date range and, in fact, many times it should not be a date range.

1. First of all the copyright is implicit, so we'd not really need it even though we have. But now that we have it it is important to understand that
2. The copyright year reflects the first year of the original publication, i our case 2010

>The general rule is that the year to include in a copyright notice is the year of first publication of the work. First publication is when the work is made available to the public without restriction.  This includes selling a book or leaving free copies of it in a public place or posting content on your blog. Publication does not include, for example, circulating copies internally to co-workers. For compilations or derivative works which incorporate previously published content, use the year of first publication of the compilation or derivative work.
>
>New versions or editions of works should contain the publication date of the new version or edition. For constantly evolving websites and blogs that contain works published over several years, the notice may include a range of years (e.g., 2009-2013), starting from the date of the oldest published elements and ending with the date of the newest published elements.

Source http://www.copyrightlaws.com/copyright-notice-year/

The last sentence is of most interest here

> For constantly evolving websites and blogs that contain works published over several years, the notice **may include a range of years** (e.g., 2009-2013), starting from the date of the oldest published elements and ending with the date of the newest published elements.

If we do put a date range on it, we are going to have to keep updating it so we do not void our copyright (and we'd have to do with across _all_ our repositories) if we forget to update it at the start of a new year. The copyright for Nancy is quite limited as well, since the entire source code is licensed under MIT. There are exceptions though, for example our logo is not licences under MIT and is protected by the copyright.

For these reasons, we've decided not to accept this pull-request.

Thanks anyway!

perlun:
Sure, it's up to you. 😉 Thanks for a great project nonetheless. ❤️

PR #2794, opened 2017-09-14
Ignore If-Modified-Since when If-None-Match is set

slozier:
From RFC 7232:
>A recipient MUST ignore If-Modified-Since if the request contains an
>   If-None-Match header field; the condition in If-None-Match is
>   considered to be a more accurate replacement for the condition in
>   If-Modified-Since, and the two are only combined for the sake of
>   interoperating with older intermediaries that might not implement
>   If-None-Match.


thecodejunkie:
Thanks!

PR #2821, opened 2017-11-12
Added global.json and pinned SDK version

thecodejunkie:
### Prerequisites

### Description
Added back the `global.json` file and pinned our SDK to 1.0.4


jchannon:
Projects array is no longer supported as the solution file takes care of that. Only the Sdk version is needed

thecodejunkie:
@jchannon updated. I've just gone with `1.0.4` - maybe we should go with `1.1.4` now instead? No point on sticking on an older `1.x` release is there? @khellang

jchannon:
Can do but you might find you'll need to update some package versions too in Nancy etc

thecodejunkie:
Maybe we pull in this one and treat that as a separate PR? There are some tooling stuff that's fixed post 1.0.4 which would be nice to have access to

PR #2841, opened 2017-12-22
Dynamic partial template name

dedo1911:
Follow up of PR #2834 
Replaces Model bindings in Partial template name to allow dynamic templates to be rendered.

_Info: There might be a better way to pass model and host parameters to ReplaceCurrentMatch and PerformSingleSubstitutions methods than all in a single row expression, but as of now It works._


dedo1911:
Test failed 🥇  I'll get it fixed and PR again 😄  Strange my env didnt warned about the error before.

PR #2848, opened 2018-01-03
Update to netstandard2.0

khellang:
### Prerequisites

### Description

This includes the changes from #2847. I'll have to rebase once that's merged :smile:


khellang:
Rebased on top of #2847

thecodejunkie:
@khellang Perhaps we should pull #2849 first and rebase your work on top of that. One less thing to worry about

khellang:
> @khellang Perhaps we should pull #2849 first and rebase your work on top of that. One less thing to worry about

Yep 👍

thecodejunkie:
@khellang wonder why this is failing on Travis but not AppVeyor? First I thought we were running into issued with Cake not being able to run since it's not built against netstandard 2.. but then it should be failing on AppVeyor as well

Only thing I've come across, so far, is https://codeblog.jonskeet.uk/2017/08/15/using-net-core-2-0-sdk-on-travis/

thecodejunkie:
@devlead @patriksvensson @gep13 either of you think this is related to CAKE + netstandard2?

devlead:
Cake.CoreCLR currently runs under .NET Core 1.0.5 which is included in .NET Core SDK 1.1.5, we've started porting it to .NET Standard 2 but not released yet.

Another option is to do like the [LitJSON/litjson](https://github.com/LitJSON/litjson) project, it compiles for .NET 4.5, .NET Standard 1.5 and .NET Standard 2.0 using .NET Core CLI/SDK 2.1.3, but Cake.exe is used which is running under full framework / Mono.

`build.sh` / `build.ps1` boostraps .NET SDK 2.1.3 and downloads Cake 0.24.0.

thecodejunkie:
@devlead so the difference would be

- Use non-core Cake
- Run Cake on Mono with Travis
- Cake runs the stuff on CLR

?

devlead:
Basically yeah, or wait for us to finish the port.

Cake's just an orchstrator, so unless you need to load some specific .NET standard addin / assembly in your build scripts,  it'll just work the same as they use the same Roslyn compiler now. 

Not as clean, but should be fairly easy to switch once Cake is ported. 

And 3rd option would be to download both 1.x and 2.x SDK in the bootstrapper, use 1.x to run cake and register 2.x with tool resolution, so that's used for building stuff.

thecodejunkie:
@devlead well there's not really any current benefit with running `Cake.CoreCLR`, over plan `Cake`, I guess

thecodejunkie:
@khellang 👀 😄 

Do we have any idea what is going on with the failing Travis build? Do we need to poke someone to get us unstuck on this? Pretty much one of the few blockers I see for us running the build script of doom and putting out a 2.0. Anything I can help with?

khellang:
I have no idea. I can't reproduce it locally. Looks like the build.sh script doesn't find some file?

> $ ./build.sh --verbosity=minimal
> : No such file or directory

devlead:
@khellang are you using .NET Core 1.1.4?
https://github.com/khellang/Nancy/blob/15cd4009b152e38793111264ee3dc3c6476a39a8/.travis.yml#L12

This is a example of a travis file I'm using for a 2.x build (could prob be 2.1.4 now)
https://github.com/LitJSON/litjson/blob/develop/.travis.yml#L16

khellang:
We're using the 2.1.3 version of the SDK, but targeting .NET Core 2.0/.NET Standard 2.0 (not sure which you're asking for 😝 )

devlead:
@khellang then in the `.travis.yml`
```diff
- dotnet: 1.1.4
+ dotnet: 2.1.3
```

khellang:
Does that matter when we fetch the correct version anyway? We want to be able to build this locally as well, not rely on the environment to have the correct dotnet version on the PATH.

khellang:
See https://github.com/khellang/Nancy/blob/15cd4009b152e38793111264ee3dc3c6476a39a8/build.sh#L25

khellang:
That line was something @thecodejunkie added in https://github.com/NancyFx/Nancy/pull/2855. We shouldn't need a version of dotnet pre-installed since the build script fetches the correct version based on `global.json`. It's just another "source of truth".

thecodejunkie:
Because we're not running a netstandard 2 build of CAKE, we need to run cake using mono, which should run nancy againt a netstandard 2 target

So

- `build.sh` and `build.ps1` should grab the SDK version from `global.json`
- Our build environment should run `cake.exe` using mono on linux https://github.com/khellang/Nancy/blob/15cd4009b152e38793111264ee3dc3c6476a39a8/build.sh#L57

khellang:
> we need to run cake using mono

Yes, which means we need

https://github.com/khellang/Nancy/blob/15cd4009b152e38793111264ee3dc3c6476a39a8/.travis.yml#L10

not

https://github.com/khellang/Nancy/blob/15cd4009b152e38793111264ee3dc3c6476a39a8/.travis.yml#L12

We already take care of fetching the correct `dotnet` version in our build script. No need for Travis to do it for us 😄

devlead:
@khellang It can matter as it also fetches the dependencies needed for .NET Core, you could install those debian packages manually though.

devlead:
https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=netcore2x

khellang:
Anyway, it's just a matter of doing *more* work. That shouldn't be why the script is failing on Travis?

thecodejunkie:
@devlead pretty sure I based it on your travis setup for https://github.com/LitJSON/litjson/ so not sure it's a pre-requisite thing since your file works

@khellang can you try and delete that line please?

devlead:
@thecodejunkie my `.travis.yml` uses `dotnet: 2.1.3` though.

devlead:
Is the build.sh executable in git? If not you can do `git update-index --add --chmod=+x build.sh`

devlead:
Found the issue the `build.sh` is CRLF committed into repo which results bash barfing on the file.

Converted to LF, and it executes, downloads SDK and starts executing the build, it fails on the test Task for me, haven't had time to investigate more, but going from CRLF to LF should get your builds going on Travis.

khellang:
Thanks @devlead! Fixed the line-endings. Let's hope it gets us further 😄

khellang:
OK, looks like that got us a step further, but now it complains that it can't find `FakeItEasy` on Travis:

> System.IO.FileNotFoundException : Could not load file or assembly 'FakeItEasy, Version=4.0.0.0, Culture=neutral, PublicKeyToken=eff28e2146d5fd2c'. The system cannot find the file specified.

And that it can't find version 2.0.4 of the runtime on AppVeyor:

> It was not possible to find any compatible framework version
> The specified framework 'Microsoft.NETCore.App', version '2.0.4' was not found.
>  - Check application dependencies and target a framework version installed at:
>  - Alternatively, install the framework version '2.0.4'.

I'm certain that 2.0.4 is installed (locally in the repo), because it outputs the version number in the top of the script:

> Microsoft .NET Core Shared Framework Host
>  Version  : 2.0.4
>  Build    : 7f262f453d8c8479b9af91d34c013b3aa05bc1ff

But I think Cake might be using the globally installed (comes with the image?) `dotnet`.

thecodejunkie:
@khellang would it be `2.1.3` that would be globally installed because of the line in `.travis.yaml` ?

thecodejunkie:
@khellang does the test pass, locally on your machine, if you run them with `2.0.4`? Maybe @blairconrad has some ideas about FakeItEasy and netstandard 2?

thecodejunkie:
For whatever reason, these 3 tests started failing on netstandard 2

image

They work fine if I run them from master, but not on your branch with netstandard 2

blairconrad:
@thecodejunkie, I don't have great knowledge to bring at this point, but we do run our own tests under 2.0.2, but only on Windows via AppVeyor. Recently @adamralph tried to update us to using xUnit 2.3.1 in our own build (see FakeItEasy/FakeItEasy#1254), and got dodgy results, which @thomaslevesque investigated a little, but hit a wall. (You can see @adamralph tried with a number of other versions, including xUnit 2.3.0, which is being added in this PR). I'm not saying that's the cause, but I don't know enough to say it _isn't_.

I'll try to see if I can find anything else out.

global.json
```diff
@@ -1,5 +1,5 @@
 {
     "sdk": {
-        "version": "1.1.4"
+        "version": "2.1.3"
```
blairconrad: I'm probably off base, but should this not be 2.0.4?

blairconrad:
Trying to reproduce, but no luck so far. On a lark, I [rolled back the xUnit and pushed](https://travis-ci.org/blairconrad/Nancy/builds/340034934), but get a different failure, about missing 1.0.4. Probably I just haven't configured Travis properly.
At home, under the Nancy Vagrant Mono 4.x image, I have a compilation failure:

```
  Nancy.Testing -> /vagrant/Nancy/src/Nancy.Testing/bin/Release/netstandard2.0/Nancy.Testing.dll
/vagrant/Nancy/.dotnet/sdk/2.1.3/Roslyn/Microsoft.CSharp.Core.targets(84,5): error MSB6006: "csc.dll" exited with code 137. [/vagrant/Nancy/test/Nancy.Tests/Nancy.Tests.csproj]

Build FAILED.

/vagrant/Nancy/.dotnet/sdk/2.1.3/Roslyn/Microsoft.CSharp.Core.targets(84,5): error MSB6006: "csc.dll" exited with code 137. [/vagrant/Nancy/test/Nancy.Tests/Nancy.Tests.csproj]
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:35.14
An error occurred when executing task 'Compile'.
Error: One or more errors occurred.
        .NET Core CLI: Process returned an error (exit code 1).
```

which might be due to my changes to the source. I'll try again with the new xUnit, but in North American Morning.

blairconrad:
I've not had much time to play today, but when using xUnit 2.3.1, the [build still fails](https://travis-ci.org/blairconrad/Nancy/jobs/340124883), but in Nancy.Testing.Tests, earlier than when using 2.3.0. Interestingly, it doesn't obviously fail on a FakeItEasy-related problem, but in Nancy.Testing.Tests.TestingViewExtensions.GetModelExtententionsTests, which doesn't use FakeItEasy.
I'm not drawing any conclusions yet, other than "stuff be weird".

adamralph:
Out of interest, what is the reason for switching from `netstandard1.6` to `netstandard2.0`.

khellang:
> Out of interest, what is the reason for switching from `netstandard1.6` to `netstandard2.0`?

We get (more or less) the same platform support, but with much bigger API surface and tooling 😊

adamralph:
Fair enough. The biggest difference is dropping .NET Core 1.x but I guess that's no big deal now. I imagine anyone who was brave enough to use it can also move fast enough to have switched to .NET Core 2.0 by now.

blairconrad:
I keep going after this from an xUnit/FakeItEasy standpoint (but that was why I was invited), but I just created a fresh xunit project on top of dotnet 2.0.4/2.1.3, and it behaves how I'd hope both on Windows and the NancyMono4.x vagrant image (well, once I installed the proper dotnet).

khellang:
> @khellang would it be 2.1.3 that would be globally installed because of the line in .travis.yaml ?

Yes, 2.1.3 is the SDK that is globally installed, but that brings in version 2.0.4 of the runtime. It's a horrible horrible versioning scheme.

khellang:
> For whatever reason, these 3 tests started failing on netstandard 2

`netstandard2.0` is not a platform. It can't run tests. I think you're running the tests under .NET Core 2.0 and that it's an intentional behavior change between .NET Framework and .NET Core.

blairconrad:
For what it's worth, I [tried the new xUnit 2.4 beta](https://github.com/blairconrad/Nancy/compare/15f399549ab5...5ca10fc4e8ac), and [we have a successful Travis build](https://travis-ci.org/blairconrad/Nancy/builds/346246660).

khellang:
Yeah, looks like that fixed Travis, but it broke AppVeyor 😝 

> Detecting target frameworks in Nancy.Authentication.Basic.Tests.csproj...
Locating binaries for framework netcoreapp2.0...
Running .NET Core 2.0.4 tests for framework netcoreapp2.0...
Cake.exe : It was not possible to find any compatible framework version

blairconrad:
The "It was not possible to find any compatible framework version" error exists even before my commit. I sent 4365230a30481848c52f08cd624931c894fd31e5 up to AppVeyor and [see it](https://ci.appveyor.com/project/NancyFx/nancy/build/2.0.0-ci0001905-arfkmoon#L1120), and the it appears all through the test runs, which somehow still succeed. There's a puzzle here…

blairconrad:
Switching to 2.1.4/2.0.5 resolves that problem, and many tests run, but we do have a few [Should_not_throw_exception_when_path_contains_invalid_filename_character](https://ci.appveyor.com/project/NancyFx/nancy/build/2.0.0-ci0001905-pxrbmfqu#L1107) failures. I've not gone deeper than that yet.

PR #2849, opened 2018-01-03
Grab the SDK version from global.json

thecodejunkie:
This should prevent the build scripts from becoming out of sync with the SDK version (`build.ps1` looks completely changed because line-endings were changes to `CRLF`)


thecodejunkie:
@khellang 👀 😉

PR #2856, opened 2018-01-05
SelfHost: Change to callback based responses for better performance

jocull:
### Prerequisites

### Description

I noticed that the SelfHost was not performing well when using `async` methods. You can test a simple endpoint using `await Task.Delay(100)` and see them queue up as the engine handle blocks waiting for the task. This change converts it to complete/error callbacks so they no longer block. Hope it helps, and please let me know if I missed anything! I'm new here :)

### DIFF
--- a/src/Nancy.Hosting.Self/NancyHost.cs
+++ b/src/Nancy.Hosting.Self/NancyHost.cs
@@ -426,24 +426,31 @@
 
         private void Process(HttpListenerContext ctx)
         {
+            Action<Exception> onError = (e) =>
+            {
+                this.configuration.UnhandledExceptionCallback.Invoke(e);
+            };
+
+            Action<NancyContext> onComplete = (nancyContext) =>
+            {
+                try
+                {
+                    ConvertNancyResponseToResponse(nancyContext.Response, ctx.Response);
+                }
+                catch (Exception e)
+                {
+                    onError(e);
+                }
+            };
+
             try
             {
                 var nancyRequest = this.ConvertRequestToNancyRequest(ctx.Request);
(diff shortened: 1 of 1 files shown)


src/Nancy.Hosting.Self/NancyHost.cs
```diff
@@ -426,24 +426,31 @@ private void GotCallback(IAsyncResult ar)
...
+                catch (Exception e)
+                {
+                    onError(e);
+                }
+            };
+
+            try
+            {
+                var nancyRequest = this.ConvertRequestToNancyRequest(ctx.Request);
+                this.engine.HandleRequest(nancyRequest, onComplete, onError);
```
khellang: Where is this overload defined? I think I must be blind, cause I just can't find it 😕

jocull: I’m on my phone so I can’t be sure, but I believe this is it. Note that this is on the 1.x branch since I was working against stable.

https://github.com/NancyFx/Nancy/blob/b826aaf6c00b51804c0de9542c1f7c13a3ef2604/src/Nancy/NancyEngineExtensions.cs#L78

thecodejunkie:
@khellang this is a PR sent tot he `1.x-WorkingBranch`
@jocull `1.x` has been in maintenance mode for quite some time now and I'd rather see you sent this PR against `master`

khellang:
But that overload isn't there in 2.0. We've moved everything over to async/await and Tasks.

jocull:
I can try to check this out in 2.x - it may not be an issue in the new versions, but was an issue in the current stable.

thecodejunkie:
@jocull yeah it would be great if you could check if it happens on `master` / latest nuget release

PR #2861, opened 2018-02-05
The minimum version of nuget packages was increased. 

vik-borisov:
## Prerequisites
- [x] I have verified that there are no overlapping pull-requests open
- [x] I have verified that I am following the Nancy code style guidelines

## Description
Paket, package manager, think that the version "clinteastwood" is higher than "Pre*"


thecodejunkie:
Not quite sure what the intended change here is? `2.0-clinteastwood` is a pre-release of 2.0, following semver... the only packages we release as `-pre` are those that we host on MyGet, which are considered the "bleeding edge" packages

vik-borisov:
@thecodejunkie 
I incorrectly put it. 
For pre-releases, the minimum version is 2.0-clinteastwood, which is considered higher than the current version of the package.
For example:
default

Current version 2.0.0-Pre1902 is lower than dependency version 2.0-clinteastwood

thecodejunkie:
The real question is; why is a pre-release version published on our nuget feed 🤔 

any ideas @khellang @jchannon ? did we do a boo-boo when we've been poking around with the travis/appveyor setups?

jchannon:
https://www.nuget.org/packages/Nancy.Hosting.Aspnet/ I dont see that Pre1902 anywhere

I think I recall we had spotted this issue on MyGet when we started the netcore stuff and deleted the Pre packages and hence why we went with alpha, barneyrubble, clinteastwood to make them alphabetical fixing the semver resolving of numbers etc

khellang:
I don't think we've published anything to NuGet. To be honest, I have no clue what this PR is about.

jchannon:
Oh just noticed it was a PR 😄 

Yup sorry this PR makes no sense

khellang:
`2.0.0-clinteastwood` should never be considered higher than `2.0.0-pre123`. According to [SemVer](https://semver.org/), pre-release versions should be compared using ASCII sort, and AFAIK, 'P' comes after 'C' 😉 

> Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows: identifiers consisting of only digits are compared numerically and **identifiers with letters** or hyphens **are compared lexically in ASCII sort order**.

(emphasis mine)

If Paket says something else, I'd consider that a bug. Anyway, proper pre-releases (like `-clinteastwood`) and CI pre-releases (like `-pre123`) should never exist in the same NuGet feed. If that's the case, we might've screwed something up.

vik-borisov:
@thecodejunkie, it is Myget feed

vik-borisov:
@khellang 

> .0.0-clinteastwood should never be considered higher than 2.0.0-pre123. According to SemVer, pre-release versions should be compared using ASCII sort, and AFAIK, 'P' comes after 'C' 😉

but the version of -Pre *. P - is in the upper case.
default

khellang:
Oh lord 😱

vik-borisov:
as an option, rename releases to "pre- *" in lowercase

khellang:
Ideally, we should change it to `-pre.<ci-number>` (SemVer 2.0) so we wouldn't have to pad the number with leading zeroes, but that'll have to be an issue for another time.

khellang:
Thanks @vik-borisov! I've nuked all the existing packages on MyGet, so now we should have a single version; `2.0.0-pre1903`.

There's still an issue with the Nancy dependency, though:

image

Looks like packages that depend on the Nancy project end up referencing the wrong version.

khellang:
I'm honestly not sure where to look for this bug in the CAKE script 😢

vik-borisov:
@khellang 
I discarded these changes because this is the way to migrate from nuget to myget

vik-borisov:
@khellang This parameter is in Directory.Build.props file

PR #2868, opened 2018-02-24
Small perfomance improvments for Dictionary

DamirAinullin:
Use TryGetValue instead of ContainsKey and Getter for Dictionary to prevent double calculation of HashCode.


DamirAinullin:
@thecodejunkie Can you review and merge it please? It is simple improvement of work with Dictionary.

PR #2880, opened 2018-03-20
Fix memory leak when serving static content

jeffthiele-iq:
### Prerequisites

- [ ] I have written a descriptive pull-request title
- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/NancyFx/Nancy/pulls) open
- [ ] I have verified that I am following the Nancy [code style guidelines](https://github.com/NancyFx/Nancy/blob/45238076ad0b7f6ecabd6bae8469e30458d02efe/CONTRIBUTING.md#style-guidelines)

### Description

Added calls to dispose the `CancellationTokenSource` when the engine returns static content and on errors. Potentially fixes the memory leaks described in #2605 without reverting the fix in #2150

### DIFF
--- a/src/Nancy/NancyEngine.cs
+++ b/src/Nancy/NancyEngine.cs
@@ -115,9 +115,16 @@
             var staticContentResponse = this.staticContentProvider.GetContent(context);
             if (staticContentResponse != null)
             {
-                context.Response = staticContentResponse;
-                tcs.SetResult(context);
-                return tcs.Task;
+                try
+                {
+                    context.Response = staticContentResponse;
+                    tcs.SetResult(context);
+                    return tcs.Task;
+                }
+                finally
+                {
+                    cts.Dispose();
+                }
             }
 
             var pipelines = this.RequestPipelinesFactory.Invoke(context);
@@ -147,7 +154,14 @@
                 },
                 errorTask =>
                 {
-		            tcs.SetException(errorTask.Exception);
+                    try
+                    {
+                        tcs.SetException(errorTask.Exception);
+                    }
+                    finally
+                    {
+                        cts.Dispose();
+                    }
                 },
                 true);
 


src/Nancy/NancyEngine.cs
```diff
@@ -115,9 +115,16 @@ public Task<NancyContext> HandleRequest(Request request, Func<NancyContext, Nanc
...
             if (staticContentResponse != null)
             {
-                context.Response = staticContentResponse;
-                tcs.SetResult(context);
-                return tcs.Task;
+                try
+                {
+                    context.Response = staticContentResponse;
+                    tcs.SetResult(context);
+                    return tcs.Task;
```
cloudhunter89: Does the TaskCompletionSource use or reference the Cancellation Token?  I don't quite understand how the 
`try ... finally { cts.Dispose() }` is any different than

    cts.Dispose(); 
    return tsc.Task;

This isn't dependent on the asynchronous execution of anything, is it?

jeffthiele-iq: The only difference is that with `try{...}finally { cts.Dispose();}` will still dispose of the `CancellationTokenSource` whether an exception occurs on the intervening lines or not.

cloudhunter89: Ah, going for the equivalent behavior of declaring a using block inside this scope, and there is some possibility of an exception setting the result or the response.  That makes more sense to me.

cloudhunter89: In theory each of the individual `try...finally...` could be replaced with separate `using (cts) {...}` right?  (Just working out some personal edification here)

jeffthiele-iq: In these cases, yes, I think it would be functionally the same. Personally, I like being more explicit that we're only disposing of the item here since the typical use of `using` also instantiates the object and could potentially be misinterpreted by others looking at the code.

PR #2957, opened 2019-03-28
Self host project should also target netstandard2.0

Yitzchok:
### Prerequisites

### Description
Support for self hosting in a netstandard2.0 project.

### DIFF
--- a/src/Nancy.Hosting.Self/Nancy.Hosting.Self.csproj
+++ b/src/Nancy.Hosting.Self/Nancy.Hosting.Self.csproj
@@ -2,6 +2,6 @@
   <PropertyGroup>
     <Description>Enables hosting Nancy in any application.</Description>
     <PackageTags>$(PackageTags);Host</PackageTags>
-    <TargetFramework>net452</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net452</TargetFrameworks>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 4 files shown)


davidindra:
I am also interested in this PR, when will it be merged?

grumpydev:
Not sure about how the username is removed, would probably make more sense to remove the automatic reservation stuff if there's no way to get the username that's required to do it?

Yitzchok:
Let me see if it will work with https://www.nuget.org/packages/System.Security.Principal.Windows/

cocowalla:
AFAIK, URL reservations are only required if self hosting with http.sys (which you'd only use over Kestrel if you needed Windows authentication)?

And of course http.sys is Windows only, so perhaps the conditional checks here should be for the presence of Windows, rather than .NET Core?

grumpydev:
@cocowalla is there a simple way to do that at runtime?

khellang:
This package won't work on non-Windows OSes anyway, right? I'm surprised it even compiles for netstandard2.0...

grumpydev:
@khellang not checked the code, but I assumed it would just be a noop on non-windows platforms?

khellang:
So apparently it was ported to Core with 2.0 and subsequently brought into the standard. Still seems to be based on HTTP.SYS though; https://github.com/dotnet/platform-compat/issues/88

khellang:
Anyway, you can use `RuntimeInformation.IsOSPlatform` to check OS at runtime: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.isosplatform?view=netframework-4.8

grumpydev:
That would mean changing the .net target to be 471 though i guess? As long as it runs as is on linux/osx, and doesn't blow up because that package is there it should be fine, that auto registration code is only a fallback if it doesnt have permission to bind (which should only be the case on windows)

khellang:
> That would mean changing the .net target to be 471 though i guess?

I think it's fine as it is now, using the `System.Security.Principal.Windows` package. Not sure what it does to Mono support, though.

PR #3006, opened 2020-03-10
Add SameSite support for NancyCookies

### LINKED ISSUE #3002: NancyCookie doesn't have support for SameSite changes
Nancy Version: Tested 1.4.5 and 2.0.0

Currently with all the SameSite cookie changes being done in Chrome (and FF and Edge) there is no nice way to set the SameSite options for a `NancyCookie`. See https://docs.microsoft.com/en-us/aspnet/core/security/samesite?view=aspnetcore-3.1 for more details.

Particularly this becomes a higher priority now that Chrome 80 is stable and has breaking changes around SameSite properties.

The only currently workaround I can see for the moment is to add a OWIN middleware and manually change the cookies myself.

0x414c49:
### Prerequisites

### Description
Nancy Cookies doesn't support SameSite property. This PR adds SameSite property to cookies. The possible values for SameSite property are defined by an enum **SameSite.cs**:  `Lax`, `Strict` and `None`.

This change won't break anything, all cookies work as they suppose to work. All the tests are passed and two new tests added to cover the new functionality.

Fix #3002

### DIFF
--- a/src/Nancy/Cookies/INancyCookie.cs
+++ b/src/Nancy/Cookies/INancyCookie.cs
@@ -52,5 +52,10 @@ namespace Nancy.Cookies
         /// Whether the cookie is secure (i.e. HTTPS only)
         /// </summary>
         bool Secure { get; }
+
+        /// <summary>
+        /// Wheather the cookie is same site
+        /// </summary>
+        SameSite? SameSite { get; }
     }
 }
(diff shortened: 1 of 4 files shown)


drewfreyling:
Nice work on the pr. I think setting the secure value to true automatically isn't the right approach. That should be user set. Not all browsers are doing the none+secure thing. Even if they were it makes the code more obvious if the user is setting it themselves.

0x414c49:
> Nice work on the pr. I think setting the secure value to true automatically isn't the right approach. That should be user set. Not all browsers are doing the none+secure thing. Even if they were it makes the code more obvious if the user is setting it themselves.

Kinda agree with you. I will change to codes and push them.

